PR #12, opened 2019-05-05
Fix running under python 3

cybojenix:
This is by no means extensive. It just fixes the errors I got when running `build` and `up`.

- Check for `__iter__` after checking for `isinstance(..., str)`
  `__iter__` is present on `str` in python 3
- coerce `dict.values()` to a list before sorting it
  python 3 introduces `dict_values`, which is closer to being an iterable, and so does not have `.sort`

### DIFF
--- a/podman-compose.py
+++ b/podman-compose.py
@@ -63,8 +63,6 @@ def rec_subs(value, dicts):
     """
     if hasattr(value, "items"):
         value = dict([(k, rec_subs(v, dicts)) for k, v in value.items()])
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
would you rather add dependency on [six](https://pypi.org/project/six/).
would that make our life easier? [see this](https://pythonhosted.org/six/).

muayyad-alsadi:
I'll make python 3 officially supported ASAP

PR #19, opened 2019-06-27
Support security opt

abraverm:
* Fix RuntimeError: Set changed size during iteration


muayyad-alsadi:
your PR tries to solve #20 by using `copy()` 
but it won't work because the return value is not used, it should be in place.

I'll fix that

PR #22, opened 2019-07-08
Support healthcheck associative array

stefanb2:
Map keys to corresponding --healthcheck-XXX option

### DIFF
--- a/podman-compose.py
+++ b/podman-compose.py
@@ -474,6 +474,45 @@ def container_to_args(cnt, dirname, podman_path, shared_vols):
             args.extend(['--entrypoint', entrypoint])
         else:
             args.extend(['--entrypoint', json.dumps(entrypoint)])
+
+    healthcheck = cnt.get('healthcheck')
+    if healthcheck is not None:
+        if is_dict(healthcheck):
+            if 'test' in healthcheck:
+                command = healthcheck['test']
+                # test must be either a string or a list. If it’s a list,
+                # the first item must be either NONE, CMD or CMD-SHELL.
+                # If it’s a string, it’s equivalent to specifying CMD-SHELL
+                # followed by that string.
+                if is_str(command):
+                    if re.search(r'\s', command):
+                        # podman does not add shell to handle command with whitespace
+                        args.extend(['--healthcheck-command', '"/bin/bash -c \'{}\'"'.format(command)])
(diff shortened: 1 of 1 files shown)


stefanb2:
I needed to add `healthcheck` section to my docker-compose.yml because of containers/libpod#3507

muayyad-alsadi:
Thank you very much for this kind contribution. I'll look into it very soon

stefanb2:
Hmm, looks like `['CMD-SHELL', 'echo with whitespace']` does not work correctly, i.e. podman tries to find the executable `echo with whitespace` instead of executing a shell command line.

muayyad-alsadi:
there are two things here

1. build-time (in `Dockerfile`) specification of shell and healthchecks
2. run-time (in `docker-compose.yml`) specification of healthchecks (no shell)

build-time issues should be reported to buildah
run-time issues should be reported to podman

from the compose.yml reference

```
test: ["CMD-SHELL", "curl -f http://localhost || exit 1"]
```

which should be interpreted as  

```
test: ["CMD", "/bin/sh", "-c", "curl -f http://localhost || exit 1"]
```

assuming default shell `['/bin/sh', '-c']`

this means that podman should interpret `['CMD-SHELL', 'echo with whitespace']`
as ['CMD', '/bin/sh', '-c',  'echo with whitespace']


https://docs.docker.com/compose/compose-file/#healthcheck
https://docs.docker.com/engine/reference/builder/#healthcheck

muayyad-alsadi:
I've seen the diff, it's not `podman` fault, we should pass it the the arguments in wrong way,

no worries, I'll fixes it ASAP.

stefanb2:
Unfortunately things aren't that easy. It turns out that `--healthcheck-command` really only accepts a string, not a list of strings. Furthermore the string gets split on whitespace, i.e. 
```
--healthcheck-command "stat /etc/passwd || exit 1"
```
ends up in the container configuration as
```json
"Healthcheck": {
    "Test": [
        "stat",
        "/etc/passwd",
        "||",
        "exit",
        "1"
    ],
```

My updated patch tries to approximate as best as possible, but not all valid Docker Compose values will be able to be mapped correctly

muayyad-alsadi:
I've used 

```
from shlex import quote as cmd_quote
```

PR #30, opened 2019-08-21
Devel

barseghyanartur:
Add python packaging.

Let me know your PyPI username for ownership transfer.

In short, added proper python packaging. From now on it's possible to install the package from PyPI:

```bash
pip install podman-compose
```

or 

```bash
pip install https://github.com/muayyad-alsadi/podman-compose/archive/devel.tar.gz
```

All mentioned in docs. I marked it as status 3 alpha (cause it felt so).


barseghyanartur:
Sorry, wrong source -> target branch. See my other pull request.

PR #31, opened 2019-08-21
Python packaging

barseghyanartur:
Add python packaging.

Let me know your PyPI username for ownership transfer.

In short, added proper python packaging. From now on it's possible to install the package from PyPI:

```bash
pip install podman-compose
```

or 

```bash
pip install https://github.com/muayyad-alsadi/podman-compose/archive/devel.tar.gz
```

All mentioned in docs. I marked it as status 3 alpha (cause it felt so).


barseghyanartur:
Update: I've added you as project owner on PyPI.

https://pypi.org/project/podman-compose/

.gitignore
```diff
@@ -20,6 +20,7 @@ parts/
 sdist/
 var/
 wheels/
+.idea/
```
dockerest: this belongs in global .gitignore, not here

muayyad-alsadi:
Thank you, I have two notes

You said it might not work, but it's working fine as far as I know (the devel brach at least), please open an issue any composer file that does not work and you think it's podman-compose fault.

Why scripts/ is needed and find package in src/, you might want to make make that just `[ ]`

barseghyanartur:
@muayyad-alsadi:

- That part I copied. Initial text was "it's still underdevelopment and does not work yet."
- That `src` part was wrong. It's scripts part that make the ``podman-compose`` command available globally, without a need to manually registering it in your environment.

muayyad-alsadi:
It's a single file module not a package similar to this one

https://github.com/bottlepy/bottle/blob/master/setup.py

Please adjust setup.py in a way similar to it (no package and no scripts directory)

barseghyanartur:
@muayyad-alsadi:

Do you mean this:

- remove ``package_dir={'': '.'},`` line.
- remove ``scripts`` directory (where files clean_up.sh, make_release.sh and uninstall.sh are located).

Or something else?

barseghyanartur:
As for ``entry_points`` in setup.py, it's needed in order to make ``podman-compose`` command available. If we just use ``scripts`` directive (as in ``bottle``), the ``podman-compose.py`` command will be available instead.

muayyad-alsadi:
> As for entry_points in setup.py, it's needed in order to make podman-compose command available.

yes your entry-point is 100% correct

```
entry_points={
        'console_scripts': [
            'podman-compose = podman_compose:main'
        ]
    },
```

it's main function in podman_compose module

PR #32, opened 2019-08-22
Devel

barseghyanartur:
Tiny fix.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -875,7 +875,7 @@ def compose_up(compose, args):
         # `podman build` does not cache, so don't always build
         build_args = argparse.Namespace(
             if_not_exists=(not args.build),
-            **args.__dict__,
+            **args.__dict__
         )
         compose.commands['build'](compose, build_args)
     


muayyad-alsadi:
I like having extra comma for multi line args or lists or dicts

```
A=[
  1,
  2,
  3,
]
```

Sorry

barseghyanartur:
@muayyad-alsadi:

That comma is causing errors on Python 2.

muayyad-alsadi:
Python 2 is reaching EOL in 2020

barseghyanartur:
@muayyad-alsadi:

No problem at all. When I was proposing the fix, I was considering current state of the code at time and as far as I could see it was (and still is written) with Python 2 in mind.

This:

```python
from __future__ import print_function
```

Or this:

```python
PY3 = sys.version_info[0] == 3
if PY3:
    basestring = str
```

These are Python 2 artifacts.

Would you rather accept (and interested in) a Python 2 clean-up pull request?

PR #38, opened 2019-09-11
Add support for setting container ulimit

tylarb:
- supports ulimit: host or other single value
- As well as ulimit: [nofile=5000, nproc=5000] in array

Authored-by: Tyler Ramer <tramer@pivotal.io>

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -206,6 +206,17 @@ def norm_as_dict(src):
         raise ValueError("dictionary or iterable is expected")
     return dst
 
+def norm_ulimit(inner_value):
+    if is_dict(inner_value):
+        if not inner_value.keys() & {"soft", "hard"}:
+            raise ValueError("expected at least one soft or hard limit")
+        soft = inner_value.get("soft", inner_value.get("hard"))
+        hard = inner_value.get("hard", inner_value.get("soft"))
+        return "{}:{}".format(soft, hard)
+    elif is_list(inner_value): return norm_ulimit(norm_as_dict(inner_value))
+    # if int or string return as is
+    return inner_value
+
 
 # transformation helpers
 
@@ -477,6 +488,17 @@ def container_to_args(compose, cnt, detached=True, podman_command='run'):
         podman_args.append('-i')
     if cnt.get('tty'):
         podman_args.append('--tty')
+    ulimit = cnt.get('ulimit', [])
+    if ulimit is not None:
+        # ulimit can be a single value, i.e. ulimit: host
+        if is_str(ulimit):
+            podman_args.extend(['--ulimit', ulimit])
+        # or a dictionary or list:
+        else:
+            ulimit = norm_as_dict(ulimit)
+            ulimit = [ "{}={}".format(ulimit_key, norm_ulimit(inner_value)) for ulimit_key, inner_value in ulimit.items()]
+            for i in ulimit:
+                podman_args.extend(['--ulimit', i])
     # currently podman shipped by fedora does not package this
     # if cnt.get('init'):
     #    args.append('--init')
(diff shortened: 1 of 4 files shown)


muayyad-alsadi:
How about nested like

```
ulimits:
  nproc: 65535
  nofile:
    soft: 20000
    hard: 40000
```

And should we call normalize as dict helper

tylarb:
@muayyad-alsadi  sure I can do that - how do you think we should handle ulimit=host?

muayyad-alsadi:
The same way you do it, is_str ...

tylarb:
I used norm_as_list as the input to podman create should actually be a string, and it's cleaner to use the `"value=key"` returned by list than  `value + "=" + key` if we used a dict.

muayyad-alsadi:
No need for is_dict because norm as list would do that check for you.

The question is how do you want to handle/support nested dictionaries of soft / hard?
See the example above

tylarb:
@muayyad-alsadi  should we consider a mixed nested mix of list and dict?

i.e. these two options:
```
# unsure if this is even valid yaml
service1:
  ulimit:
   - nproc: {soft: 1000, hard: 2000}
   - nofile=5000
# this is valid yaml however
service2:
  ulimit:
    nproc:
    - soft=1000
    - hard=2000
    nofile: 5000
```

norm as dict only parses one list depth
```
def norm_as_dict(src):
    """
    given a list ["key1=value1", "key2"]
    return a dictionary {key1:value1, key2: None}
    """
    if src is None:
        dst = {}
    elif is_dict(src):
        dst = dict(src)
    elif is_list(src):
        dst = [i.split("=", 1) for i in src if i]
        dst = dict([(a if len(a) == 2 else (a[0], None)) for a in dst])
    else:
        raise ValueError("dictionary or iterable is expected")
    return dst
```

and docker-compose only supports dictionary - the the ability to use a list may be unecessary:

https://docs.docker.com/compose/compose-file/#ulimits

Let me know what you have in mind and I'll make it happen.

tylarb:
I'm thinking scope should be:

- just string:
```
ulimit: host
```
- list, and no soft/hard limit (or use colon the way podman does)
```
ulimit:
- nproc=1024:2048
- nfile=5000
```

- dictionary, with nested soft/hard limit:
```
ulimits:
  nproc: 65535
  nofile:
    soft: 20000
    hard: 40000
```

and No mixed list/dict.

muayyad-alsadi:
> I'm thinking scope should be:

I agree, we might just into "<LIMIT_KEY>=<SOFT>:<HARD>"

if string take it as is
norm as dict, and loop over `<LIMIT_KEY>`, `<INNER_VALUE>`

```python
def norm_ulimit(inner_value):
    if is_dict(inner_value):
        soft = inner_value["soft"]
        hard = inner_value.get("hard", soft)
        return "{}:{}".format(soft, hard)
    elif is_list(inner_value): return norm_ulimit(norm_as_dict(inner_value))
    # if integer or string just return it as is
    return inner_value

# ...
ulimit = cnt.get('ulimit', [])
if ulimit is not None:
    if is_str(ulimit): ...
    else:
       ulimit = norm_as_dict(ulimit)
       ulimit = [ "{}={}".format(ulimit_key, norm_ulimit(inner_value)) for ulimit_key, inner_value in ulimit.items() ]
            
```

what do you think?

we need a test docker-composer.yml with something like

```
podman run --rm -ti --ulimit nofile=512:1024 busybox sh -c 'ulimit -S -n; ulimit -H -n'
```

tylarb:
I've added your suggestions with a check for requiring at least one of soft or hard limit.

I'm not entirely sure of the test structure, and what I've added seems a little heavy for what you currently have in place, but it does demonstrate the behavior is correct.

Let me know if there anything else to cover

muayyad-alsadi:
Thank you

PR #42, opened 2019-09-14
Support extends

mariushoch:
https://docs.docker.com/compose/extends/

Doesn't yet support referring to another file.

Needs Python 3.5+ (for merging dicts using **)

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -753,6 +753,9 @@ class PodmanCompose:
             print(" ** merged:\n", json.dumps(compose, indent = 2))
         ver = compose.get('version')
         services = compose.get('services')
+
+        services = self._resolve_service_extends(services)
+
         # volumes: [...]
         shared_vols = compose.get('volumes', {})
         # shared_vols = list(shared_vols.keys())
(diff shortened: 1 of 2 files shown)


mariushoch:
This is somewhat WIP still :/

muayyad-alsadi:
please don't check for docker-compose file version. 
just implement the feature.

BTW: we already support two ways of extending

1. via passing multiple files (by default "docker-compose.override.yml"), for example with -f 1.yml 2.yml

2. via yaml magic like [tests/tamlmagic](https://github.com/containers/podman-compose/blob/devel/tests/yamlmagic/docker-compose.yml)

mariushoch:
> please don't check for docker-compose file version. just implement the feature.

Good idea, no need to be strict here. Removed.
> BTW: we already support two ways of extending

Hm, I don't think these fit the needed semantics close enough.

muayyad-alsadi:
I'm considering using:

```
def rec_merge(target, source):
```

https://github.com/containers/podman-compose/blob/dee813a4b6f0f06131f021147ae09cef3eb5c875/podman_compose.py#L644

mariushoch:
> I'm considering using:
> 
> ```
> def rec_merge(target, source):
> ```
> 
> https://github.com/containers/podman-compose/blob/dee813a4b6f0f06131f021147ae09cef3eb5c875/podman_compose.py#L644

How would that look? I'm fairly sure that's not going to be enough for my use cases :/

muayyad-alsadi:
I'll handle it.

muayyad-alsadi:
files are not supported

https://github.com/containers/podman-compose/commit/9e0dd2da9d70731cf2068d62b495f6e2883e8870

PR #43, opened 2019-09-15
Fix podman-compose run command parsing

mariushoch:
This got confused with the main command (up, down, …), thus leading to:

```
    cmd = self.commands[cmd_name]
TypeError: unhashable type: 'list'
```


muayyad-alsadi:
There is a similar `docker-compose exec`
Is there a common name that can be used for both run_command and exec_command?
Ex. `cnt_command`?

mariushoch:
That sounds sensible, will change.

PR #44, opened 2019-09-15
podman compose up/run: Assume root

mariushoch:
I needed this in order to make several docker-compose work, that required writing to mounted volumes (as these will be owned by root). I'm not sure this is entirely the right approach/ best way to go (it only worked for me).

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1071,6 +1071,8 @@ def compose_up(compose, args):
 
     create_pods(compose, args)
     for cnt in compose.containers:
+        # Default to root, as podman maps the current user to root in the container
+        cnt['user'] = 'root'
         podman_args = container_to_args(compose, cnt,
             detached=args.detach, podman_command=podman_command)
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
This seems very wrong! It should default to container user specified in dockerfile.

mariushoch:
> This seems very wrong! It should default to container user specified in dockerfile.

That makes sense… but what if there's none, in that case we default to the current user (but the volumes will still be owned by root). Should we maybe set `--userns=keep-id` (just in case no user is given explicitly?).

mariushoch:
I just tried `--userns=keep-id`, but ran into https://github.com/containers/libpod/issues/3993 when trying to start the pod (though it works with `-t identity`, but we can't use that as default… it will probably also work with `-t cntnet` if the infra container is also set up with `--userns=keep-id`). @muayyad-alsadi do you have a suggestion on how to continue here?

PR #57, opened 2019-10-04
Adding support for "PS"

aric49:
adding support for a `podman-compose ps` request to view the status of running podman-compose projects.  This is similar to `docker-compose ps` as outlined here:  https://docs.docker.com/compose/reference/ps/

Let me know if this is something you'd be interested in pulling into the podman-compose project.  Thank you for all the work you're doing.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1014,6 +1014,13 @@ def compose_down(compose, args):
     for pod in compose.pods:
         compose.podman.run(["pod", "rm", pod["name"]], sleep=0)
 
+@cmd_run(podman_compose, 'ps', 'show status of containers')
+def compose_ps(compose, args):
+    if args.quiet == True:
+        compose.podman.run(["pod", "ps", "--ctr-ids", "--filter", f"name={compose.pods[0]['name']}"])
+    else:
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
what do you think if we change it to be

```
        compose.podman.run(["ps", "--filter", f"label=io.podman.compose.project={proj_name}"])
```

instead of

```
        compose.podman.run(["pod", "ps", "--filter", f"name={compose.pods[0]['name']}"])
```

aric49:
that's actually a much cleaner way to do it. However, locally when I try run that locally I can't see the `proj_name` variable from the context of the function:

```
  File "./podman_compose.py", line 1196, in <module>
    main()
  File "./podman_compose.py", line 1193, in main
    podman_compose.run()
  File "./podman_compose.py", line 695, in run
    cmd(self, args)
  File "./podman_compose.py", line 871, in wrapped
    return func(*args, **kw)
  File "./podman_compose.py", line 1022, in compose_ps
    compose.podman.run(["pod", "ps", "--ctr-names", "--ctr-status", "--filter", f"label=io.podman.compose.project={proj_name}"])
NameError: name 'proj_name' is not defined
```

I think there is somewhere I'm missing that I need to pass `proj_name` as an argument to the ps function...

muayyad-alsadi:
```
proj_name = compose.proj_name
```

PR #59, opened 2019-10-05
Fix some code styles issues

williamjmorenor:
Hello, first than all thanks for this awesome project.

Since there is not a official code style from the README or docs I am fixing some pep8 issues with the help fo the [pycodestyle tool](https://pypi.org/project/pycodestyle/)

I will making a commit per error code and will try to not make innecesary changes.

Regards

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -635,7 +635,7 @@ class Podman:
         self.compose = compose
         self.podman_path = podman_path
         self.dry_run = dry_run
-    
+
     def output(self, podman_args):
         cmd = [self.podman_path]+podman_args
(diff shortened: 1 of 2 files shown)


muayyad-alsadi:
Thank you for your contribution,

I'm sorry for not merging it.

We I want to add a code style I'll run the commands myself.

Yes, it's important but not as having bugs in volume mount or compose up not running dependencies

williamjmorenor:
I do understand but at less pick the unused funtion in setup.py

muayyad-alsadi:
> I do understand but at less pick the unused funtion in setup.py

@williamjmorenor picked

PR #64, opened 2019-10-20
Restart containers

mellon85:
Properly manage podman container lifetime. Restart containers in case they unexpectedly die.
Doesn't support restarting a container if it was removed (i.e. podman rm -f ID)

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -10,6 +10,7 @@
 from __future__ import print_function
 
 import sys
+import logging
 import os
 import argparse
 import subprocess
@@ -35,6 +36,9 @@ import json
 import yaml
 
 __version__ = '0.1.6dev'
+
+logger = logging.getLogger(__name__)
+logging.basicConfig()
 
 PY3 = sys.version_info[0] == 3
 if PY3:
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
Thank you, this seems a lot of work

podman create supports --restart=policy

We might just pass it?

https://github.com/containers/libpod/blob/master/docs/podman-create.1.md

mellon85:
Cool, sure, I'll switch to that, but I would still keep the changes to avoid having threads and use sigchld if that's fine with you

muayyad-alsadi:
I'll review the code in the weekend

muayyad-alsadi:
I've reviewed the code and I liked the idea

http://man7.org/linux/man-pages/man3/atexit.3.html
https://docs.python.org/3/library/atexit.html

but I don't think we can do it.
because it only handles normal exit
for example if a service went out of memory limit (set by `podman run -m 100m` or `mem_limit: 100m` in `docker-compse.yml`) or someone send a signal ...etc.

if you don't like threads, we can use `select.epoll`

https://stackoverflow.com/questions/3194018/wait-the-end-of-subprocesses-with-multiple-parallel-jobs

I'll keep this PR open. and maybe revisit it (after fixing volume mounts)

thank you for your contributions.

Issue #68, opened 2019-10-31
unknown mount option delegated


chriswalg:
After updating to Fedora 31, my docker containers will stop working and I've worked with podman.

Then I built an if query in the script docker-compose, which checks whether docker-compose or podman-compose exists and uses the correct commands.
https://gitlab.com/foodsharing-dev/foodsharing/blob/podman/scripts/docker-compose

https://gitlab.com/foodsharing-dev/foodsharing/tree/podman/scripts
https://gitlab.com/foodsharing-dev/foodsharing/tree/podman/docker

After that I installed podman-compose dev.

Unfortunately, I get an error message and the other containers do not start.

`[christian@thinkpad-local foodsharing]$ ./scripts/start
podman pod create --name=foodsharing_dev --share net -p 18080:18080 -p 18084:80 -p 18090:8080 -p 11337:1337 -p 18086:8086 -p 8083:8083 -p 4000:3000 -p 11338:1338 -p 18089:8089/udp -p 16379:6379 -p 13306:3306 -p 18081:80
94e4b500ff56c871949c3dbc0c084f18cc50ddc06bb1309d162a6c2cea9a0b80
0
Namespace(T=False, cnt_command=['sh', '-c', 'chown -R 1000 /app/client/node_modules'], command='run', detach=False, dry_run=False, e=None, entrypoint=None, file=['/home/christian/git/foodsharing/scripts/../docker/docker-compose.dev.yml'], label=None, name=None, no_ansi=False, no_cleanup=False, no_deps=True, podman_path='podman', project_name='foodsharing_dev', publish=None, rm=True, service='client', service_ports=False, transform_policy='1podfw', user='root', volume=None, workdir=None)
Traceback (most recent call last):
  File "/usr/local/bin/podman-compose", line 11, in <module>
    load_entry_point('podman-compose==0.1.6.dev0', 'console_scripts', 'podman-compose')()
  File "/usr/local/lib/python3.7/site-packages/podman_compose.py", line 1264, in main
    podman_compose.run()
  File "/usr/local/lib/python3.7/site-packages/podman_compose.py", line 753, in run
    cmd(self, args)
  File "/usr/local/lib/python3.7/site-packages/podman_compose.py", line 936, in wrapped
    return func(*args, **kw)
  File "/usr/local/lib/python3.7/site-packages/podman_compose.py", line 1122, in compose_run
    podman_args = container_to_args(compose, cnt, args.detach)
[...]

PR #70, opened 2019-11-01
Added mount option delegated and cached

chriswalg:
Close Issue https://github.com/containers/podman-compose/issues/68


darkdragon-001:
Would be nice if this could be released soon!

PR #89, opened 2019-11-21
Update README with dnf install instructions.

mavjs:
According to the package review request on Bugzilla ([rhbz#1767240](https://bugzilla.redhat.com/show_bug.cgi?id=1767240)), it is now available in f31 since 2019-11-17.


kaclak:
Duplikat dari #
Duplikat dari #

PR #97, opened 2019-12-22
Fix variable escaping with $$

garotosopa:
Variable escaping did not seem to work as expected.

In the example below, the variable turns out as just `This is a $` instead of `This is a $literal`:

```yaml
version: "3.7"
services:
    variables:
        image: busybox
        command: ["/bin/busybox", "sh", "-c", "export | grep EXAMPLE"]
        environment:
            EXAMPLE_LITERAL: This is a $$literal
```

It seems that the regular expression did not match `$$`, but it did match `$literal` starting from the second dollar sign.

This pull request rearranges the matching mechanism into just one regular expression, similar to what [Python itself does with templates](https://github.com/python/cpython/blob/5f2c1345a79f205c680ed6e0a6ed44199546d79e/Lib/string.py#L104). This fixes this issue and so far didn't introduce any regressions for me. While at it I'm also submitting tests for all supported interpolation syntax.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -141,25 +141,19 @@ def fix_mount_dict(mount_dict, proj_name, srv_name):
 # ${VARIABLE?err} raise error if not set
 # $$ means $
 
-var_re = re.compile(r'\$(\{(?:[^\s\$:\-\}]+)\}|(?:[^\s\$\{\}]+))')
-var_def_re = re.compile(r'\$\{([^\s\$:\-\}]+)(:)?-([^\}]+)\}')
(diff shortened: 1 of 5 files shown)


muayyad-alsadi:
I'll review this soon and merge it. thank you for contributing.

garotosopa:
There seems to be a conflict for this pull request. Is there any interest in getting this merged? If so I can look at it, even though I don't plan to use podman-compose any more. Otherwise I suggest this pull request to be closed so that at least it doesn't count as pending.

yuvalk:
I think similar issue exist in podman run

PR #114, opened 2020-02-14
Add support for setting dns servers and search domains

ajayramaswamy:
I have a need to force my container to use only the internal DNS server
and search domains, this used to work with docker-compose but stopped
when I moved to podman-compose and Fedora 31


fignew:
Is there anyway this can be merged? @muayyad-alsadi If you need a left hand I'd be honored to assist you with issues/pull requests if you would like the help.

PR #117, opened 2020-02-26
Use mount type volume instead of converting to bind

### LINKED ISSUE #13: support for volumes
* shared volumes (defined in the root section of the docker-compose file)
  * and support for mounting these volumes in services
* "just" volumes, a path to mount a value at that place (is this even needed with 1. in place?)

<https://docs.docker.com/compose/compose-file/#volumes>

jerdna-regeiz:
Simply use the supported mount type volume instead of converting it to bind.

This fixes a lot of issues for me, but I am not enough involved to decide if this is the right way to do it. 

It works for me as far to use shared volumes and pre-populate them. If anyone sees further issues or requirements, please let me know.

AFAIK Fixes #13


xcffl:
It has been fixed at https://github.com/containers/podman-compose/blob/294f8ee37bb37ab0535558182cf41d99dfb3cb11/podman_compose.py#L95

PR #118, opened 2020-02-27
Pass ulimit parameters when building container

adamjonescdk:
When running podman-compose build, pass ulimit arguments from
docker-compose.yml through to podman build.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -442,6 +442,21 @@ def mount_desc_to_args(compose, mount_desc, srv_name, cnt_name):
 
 
 
+def container_to_ulimit_args(cnt, podman_args):
+    ulimit = cnt.get('ulimits', [])
+    if ulimit is not None:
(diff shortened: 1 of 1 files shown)


adamjonescdk:
This change helps to resolve issues with .Net Core containers where "dotnet restore" or "dotnet build" runs out of open file handles during the build phase with the default limit of 1024.

muayyad-alsadi:
we already have this when running containers

https://github.com/containers/podman-compose/blob/devel/podman_compose.py#L509

should we unify parsing of both.

adamjonescdk:
I've moved the ulimit parsing code into a separate function as part of this changeset.  I didn't see any other places which were duplicating it, so with this change I think we just have the build and run operations calling the same helper.  (I may well be missing something, though!  Apologies if I'm misunderstanding.)

adamjonescdk:
Thank you :)

PR #128, opened 2020-03-23
Pass "target" build parameter when building an image.

nepella:
As the title says, this passes the ["target" build parameter](<https://docs.docker.com/compose/compose-file/#target>) through to `podman build`.

If anyone wonders where this might be used: For a particular PHP application, I have a Dockerfile with a common [stage](<https://docs.docker.com/develop/develop-images/multistage-build/>) that builds assets and two child stages to serve the static assets via NGINX and the PHP code by PHP-FPM.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1008,6 +1008,8 @@ def build_one(compose, args, cnt):
         "build", "-t", cnt["image"],
         "-f", dockerfile
     ]
+    if "target" in build_desc:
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
I guess this is done here.

https://github.com/containers/podman-compose/commit/f9915c47b48eee4de2eff945d73cdae9fb28c303

PR #132, opened 2020-04-02
Add support for --build-arg option

jerdna-regeiz:
Pass --build-args to all build processes just like docker-compose


muayyad-alsadi:
how to pass multiple build-args? and does this code handle it?

Issue #116, opened 2020-02-21
SyntaxError Fresh install


ChapChap:
Hi,

I've just dowloaded the script using
```shell
curl -o /usr/local/bin/podman-compose https://raw.githubusercontent.com/containers/podman-compose/devel/podman_compose.py
chmod +x /usr/local/bin/podman-compose
```

And I immediately have this error
```shell
# podman-compose --help
  File "/usr/local/bin/podman-compose", line 1049
    **args.__dict__,
                   ^
SyntaxError: invalid syntax
```

muayyad-alsadi:
Are you sure you are using pyrhon 3? Which distro and python version?

ChapChap:
RHEL 7.6 and python 3.4

ChapChap:
```shell
root@host:chap# podman-compose
  File "/usr/local/bin/podman-compose", line 905
    **args.__dict__,
                   ^
SyntaxError: invalid syntax
root@host:chap# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.5 (Maipo)
root@host:chap# which python
/bin/python
root@host:chap# ll /bin/python*
-rwxr-xr-x 2 root root 11328 Jan  4  2018 /bin/python3.6m
-rwxr-xr-x 2 root root 11328 Jan  4  2018 /bin/python3.6
-rwxr-xr-x 2 root root 11312 Mar 23  2018 /bin/python3.4m
-rwxr-xr-x 2 root root 11312 Mar 23  2018 /bin/python3.4
-rwxr-xr-x 1 root root 11920 Sep 12  2018 /bin/python2.7
lrwxrwxrwx 1 root root     9 Apr  5  2019 /bin/python2 -> python2.7
lrwxrwxrwx 1 root root    18 Apr  8  2019 /bin/python36 -> /usr/bin/python3.6
lrwxrwxrwx 1 root root     9 Apr  8  2019 /bin/python3 -> python3.4
lrwxrwxrwx 1 root root     9 Feb 21 14:54 /bin/python -> python3.6
```

Enrico204:
By looking at the source, python 3.6 is the minimum supported version

muayyad-alsadi:
Please try devel branch

Enrico204:
@muayyad-alsadi the issue remains in the `devel` branch. It's actually caused by the line `1061`, context:

```
        # `podman build` does not cache, so don't always build
        build_args = argparse.Namespace(
            if_not_exists=(not args.build),
[...]

PR #137, opened 2020-04-04
Add Python 3.6 as minimum requirement in the README

Enrico204:
By looking at the source, `podman-compose` supports running only in Python 3.6 or later (see #116). This commit adds the requirement to the README file.


muayyad-alsadi:
I guess the reason was due to `,` I've fixed that. and I think it can run on 3.x

PR #147, opened 2020-04-14
mount volumes directly

joequant:
This patch modifies the mounts since podman now can mount volumes.

This patch causes the container contents to be copied into the volume and also mounts the volume
to allow for exec's, both of which are consistent with docker behavior.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -358,44 +358,14 @@ def tr_1podfw(project_name, services, given_containers):
     return pods, containers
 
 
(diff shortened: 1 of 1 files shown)


jerdna-regeiz:
Duplicate of https://github.com/containers/podman-compose/pull/117?
Or if different/better you could copy the test

muayyad-alsadi:
I believe this is already done in one of my previous commits.

thank you for contributing.

PR #158, opened 2020-04-22
add compose_exec

flobz:
add compose_exec function


dwt:
@muayyad-alsadi is this obsolete since #292 was merged? Is there stuff in here that should be salvaged?

muayyad-alsadi:
I'll look into it

PR #174, opened 2020-05-13
Allow empty default/error value in substitution

mohd-akram:
Docker compose allows both.


muayyad-alsadi:
Please give us a compose with default value to test the change

mohd-akram:
```yaml
version: '2.1'
services:
  web:
    environment:
      - USERNAME=${USERNAME:?}
      - PASSWORD=${PASSWORD:-}
```

```shell
# should fail
podman-compose --dry-run up
# should print nothing
USERNAME=user podman-compose --dry-run up | grep -F '${PASSWORD:-}'
```

PR #179, opened 2020-05-21
needed working devel branch, hindered by non-existant args.build_arg

jotelha:
Needed "devices" option only available within devel branch, however devel branch broken as of yesterday due to "orphaned"  `args.build_arg`, probably just WIP, removed here. 

Just close pull request if obsolete.

Best,

Johannes

PR #180, opened 2020-05-22
ENH: forward restart policy, compose build args appended up args

jotelha:
Two suggestions:

1 - Simple forwarding of restart policy to `podman run`, tested for policy 'always' as in the following compose file snippet on podman v 1.6.4:

```yaml
...
services:
  mongodb:
    image: mongod-on-smb
    init: true
    restart: always
...
```

2 - appending `build` command args to up command args, as `up` might build images if they don't exist already. That replaces https://github.com/containers/podman-compose/pull/179 .

Best,

Johannes

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -567,7 +567,7 @@ def container_to_args(compose, cnt, detached=True, podman_command='run'):
     if cnt.get('privileged', None):
         podman_args.append('--privileged')
     if cnt.get('restart', None) is not None:
-        podman_args.append(['--restart', cnt['restart']])
+        podman_args.extend(['--restart', cnt['restart']])
     container_to_ulimit_args(cnt, podman_args)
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
please pull the latest branch and use something like this

```
@cmd_parse(podman_compose, ['stop', 'restart'])
def compose_parse_timeout(parser):
...
```

to make build args work in up

for example

```
@cmd_parse(podman_compose, ['build', 'up'])
def compose_build_parse(parser):
...
```

please also look around line 1281

```
parser.add_argument("--build-arg", metavar="key=val", action="append", default=[],
        help="Set build-time variables for services.")
```

I guess this should be removed (revert last change)

jotelha:
Thanks and sorry for the mess :)

muayyad-alsadi:
welcome, hope it's working fine for you.

PR #183, opened 2020-05-24
20200524 down timeout

jotelha:
* Allow for `--timeout` option in `down` command (as available for docker-compose, https://docs.docker.com/compose/reference/down/).

* Force `--timeout` to be `int`, otherwise parsing error by `podman`.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1160,8 +1160,14 @@ def compose_up(compose, args):
 
 @cmd_run(podman_compose, 'down', 'tear down entire stack')
 def compose_down(compose, args):
+    podman_args=[]
+    timeout=getattr(args, 'timeout', None)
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
Which version of podman does not accept float?

jotelha:
Some version 1.6.4 that comes with CentOS 8.1:

```console
$ cat /etc/redhat-release 
CentOS Linux release 8.1.1911 (Core) 
$ podman --version
podman version 1.6.4
$ podman stop -t 360.0 anything
Error: invalid argument "360.0" for "-t, --timeout" flag: strconv.ParseUint: parsing "360.0": invalid syntax
$ podman stop -t 360 anything
Error: no container with name or ID anything found: no such container
```

PR #190, opened 2020-06-08
Always set hostname

trusch:
This ensures that the hostname is always properly set. Some software may require it.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -507,6 +507,7 @@ def container_to_args(compose, cnt, detached=True, podman_command='run'):
     podman_args = [
         podman_command,
         '--name={}'.format(cnt.get('name', None)),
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
> Some software may require it.

Which requires it?

muayyad-alsadi:
as far as I know docker does not do this

```
$ docker inspect myproj_myserv_1 | grep -i '"Hostname"'
            "Hostname": "f23ff7f17226",
```

and if we want to do it, we should respect hostname in like 559

https://github.com/containers/podman-compose/blob/devel/podman_compose.py#L559

maybe we might need to add `else random_hex(12)`

```python
if cnt.get('hostname', None):
        podman_args.extend(['--hostname', cnt['hostname']])
else:
        podman_args.extend(['--hostname', rnd_hex(12)])


```

PR #193, opened 2020-06-10
fixes for podman v1.6.4

wznoinsk:


wznoinsk:
NOTE: I'm using podman-compose with this simple fix to deploy cachethq-docker (https://github.com/CachetHQ/Docker/blob/master/docker-compose.yml) successfully

muayyad-alsadi:
You might need to use .get() instead of []

PR #194, opened 2020-06-12
Create non-existing binding volumes

### LINKED ISSUE #185: Create non-existing bind volumes
I am not sure whether `docker` or `docker-compose` does this and conversely I don't know whether the issue lies with `podman` or `podman-compose`:

For postgres containers we often have a setup like this:

```
services:

  db:
    image: postgres
    restart: always
    environment:
      POSTGRES_PASSWORD: example
    volumes:
    - "./pg-data:/var/lib/postgresql/data"
```

in docker the folder `pg-data` is created automatically, if it doesn't exist. Via podman-compose I get

```
Error: error checking path "/tmp/foo/./pg-data": stat /tmp/foo/./pg-data: no such file or directory
125
```

lcarva:
Ideally, this would be done by podman, but in the mean time do it in
podman-compose for a better user experience.

Fixes #185

Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -89,6 +89,12 @@ def parse_short_mount(mount_str, basedir):
         mount_type = "bind"
         # TODO: should we use os.path.realpath(basedir)?
         mount_src = os.path.join(basedir, os.path.expanduser(mount_src))
+        try:
+            os.makedirs(mount_src, exist_ok=True)
+        except FileExistsError:
+            # This error may occur if mount_src is a path to an existing file. If
+            # it's an existing directory, an exception is not raised.
+            pass
     else:
         # Named volume
         # - datavolume:/var/lib/mysql
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -60,7 +60,7 @@ def try_int(i, fallback=None):
 dir_re = re.compile("^[~/\.]")
 propagation_re = re.compile("^(?:z|Z|r?shared|r?slave|r?private)$")
 
-def parse_short_mount(mount_str, basedir):
+def parse_short_mount(compose, mount_str, basedir):
     mount_a = mount_str.split(':')
     mount_opt_dict = {}
     mount_opt = None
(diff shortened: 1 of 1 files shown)


podman_compose.py
```diff
@@ -89,6 +89,8 @@ def parse_short_mount(mount_str, basedir):
         mount_type = "bind"
         # TODO: should we use os.path.realpath(basedir)?
         mount_src = os.path.join(basedir, os.path.expanduser(mount_src))
+        if not os.path.exists(mount_src):
```
xflavioribeiro, line 92: use `try/catch` instead a conditional prevent a unexpected issue

xflavioribeiro, line 92: use `try/catch` instead a conditional prevent a unexpected issue

lcarva: Good idea! Fixed it.

lcarva:
> Them user could mount in a non permitted directory and cause a issue

If that's the case, do we want the exception to bubble up? Or silently fail? Bubbling up the error might give the user a better idea of what the problem actually is.

xflavioribeiro:
Errors like it should be catched and printed to user understand what is wrong.

lcarva:
Updated the PR to only perform this action if in compatibility mode as suggested in #185.

podman_compose.py
```diff
@@ -89,6 +89,13 @@ def parse_short_mount(mount_str, basedir):
         mount_type = "bind"
         # TODO: should we use os.path.realpath(basedir)?
         mount_src = os.path.join(basedir, os.path.expanduser(mount_src))
+        if compose.compatibility_mode:
+            try:
+                os.makedirs(mount_src, exist_ok=True)
```
martinetd, line 94: just a nitpick with this: if the bind source is an existing file and not a directory, os.makedirs will fail even with exist_ok, so a previous check is required

martinetd: nm I'm blind it's just below, deleting this comment

amn:
I see that the compatibility mode [is not in the principal (`devel`) branch](https://github.com/containers/podman-compose/blob/devel/podman_compose.py#L388) -- issue #185 was closed but the commit did not introduce the compatibility mode, was the idea scrapped? There's the [PR #194](https://github.com/containers/podman-compose/pull/194) with [a commit adding the mode](https://github.com/containers/podman-compose/pull/194/commits/98bd4cf1ce9f97907b8d608f9d49d95e6b84ec28) but it's not merged. Sorry for sharing my confusion, but I tripped over the "let's create empty directories for volumes for which source files are missing" behaviour today, as I haven't used Docker and am not used to its behaviour.

PR #198, opened 2020-06-17
little improvement

sycured:
Some simplification and refacto about import (optimization) which can help about speed and memory usage

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [devel]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [devel]
+  schedule:
+    - cron: '0 13 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Override automatic language detection by changing the below list
+        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
+        language: ['python']
+        # Learn more...
+        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+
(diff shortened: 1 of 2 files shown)


muayyad-alsadi:
as far as I know, python will evaluate the entire module even if you just import a single function

for example


```python
# mymodule.py

def myfunc(): pass

print("foo")
```

if you type `from mymodule import myfunc` then `foo` will be printed despite being outside the myfunc.

can you please cite some official documentation that support your claim.
otherwise keeping the module as namespace if cleaner.

diefans:
the lookup is indeed faster if you directly import module attributes, but at the cost of maintenance.
```
❯ ipython
Python 3.7.4 (default, Apr 16 2020, 14:04:07)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.8.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os

In [2]: from os import path

In [3]: def foo():
   ...:     x = os.path
   ...:

In [4]: def bar():
   ...:     x = path
   ...:

In [5]: import dis

In [6]: dis.dis(foo)
  2           0 LOAD_GLOBAL              0 (os)
              2 LOAD_ATTR                1 (path)
              4 STORE_FAST               0 (x)
              6 LOAD_CONST               0 (None)
              8 RETURN_VALUE

In [7]: dis.dis(bar)
  2           0 LOAD_GLOBAL              0 (path)
              2 STORE_FAST               0 (x)
              4 LOAD_CONST               0 (None)
              6 RETURN_VALUE
```

sycured:
@muayyad-alsadi The support of my claim is already did by @diefans … Thanks a lot @diefans to be more quick

@diefans, my ide knows that `path` is from `os` and I see it so, this depend of each developer…
It's possible to do (an example): `from os import path as os_path`

diefans:
@sycured IMHO the module import makes more sense:
- less noise
- clear namespaces when reading code
- easy use of additional attributes without a new import statement
- clear and consistent mocking

Code is read a lot more than it is written and not everybody has a fancy IDE at hand/terminal/browser

sycured:
Sorry to use an ide that understand optimization (pycharm/vscode/theia).

So I'll have the same return about the cognitive complexity of container_to_args… 58 which it's high and also SonarQube is telling me that thing, not just flake8 with my settings : https://sonarcloud.io/project/issues?id=sycured_podman-compose&open=AXLIE3KE1HlXjxKTbGm-&resolved=false&types=CODE_SMELL

diefans:
@sycured complexity is not caused by import names or attribute lookup, but by the amount of code paths. 
In the case of `container_to_args` I would use a data class e.g. https://www.attrs.org/en/stable/ for representation and use https://github.com/Tinche/cattrs to unstructure it into a command line - that would save you over 100 lines of code. This would for sure introduce the right shape of maintainable code.

sycured:
@diefans I know but for me it's the same fight: optimization

I won't start a war about coding style, I prefer to reserve it to a blog post…

If you prefer to close this PR, you can do it… Now, I know your philosophy about secure coding… Because letting access to all sys or os when you just need one thing it isn't a good practice about less privilege or access.

diefans:
@sycured in python there is no such thing like "secure coding" or "a private or protected something" and also no philosophy about those concepts.

Honestly I don't get your point about "privilege", "access" and "good practice".

But a benefit of using module imports is to just have it at hand without introducing another import. Also mocking is not so confusing, since you mock always the source module and not the imported name of your module in question.

IMO your PR is just introducing a lot more names without improving anything. But anyways - I am only interested in this project and how it evolves, that's why I commented.

PR #208, opened 2020-06-29
Do not specify a 1 second timeout when stopping

fignew:
Even though we're throwing away this container we need to respect the default timeout in order to properly save volumes and close connections.


jotelha:
This has already been suggested in an earlier pull request https://github.com/containers/podman-compose/pull/183.

fignew:
You're right, closing.

PR #210, opened 2020-06-30
Pass build volume params

dincamihai:
Pass volumes to build command.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1072,8 +1072,13 @@ def build_one(compose, args, cnt):
             raise OSError("Dockerfile not found in "+ctx)
     build_args = [
         "build", "-t", cnt["image"],
-        "-f", dockerfile
     ]
+
+    for it in cnt['build']['volumes']:
+        build_args += ['-v', it]
+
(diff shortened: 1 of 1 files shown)


brejoc:
@muayyad-alsadi Was the underlying issue fixed in another PR? I have to admit I'm a bit confused by this being closed without comment. Thanks!

muayyad-alsadi:
@brejoc  it was automatically closed by github when I switched devel branch from `0.1.x` to `1.x`

regarding the fix

I double checked the spec and v3 and they don't take volumes property

https://github.com/compose-spec/compose-spec/blob/master/build.md
https://docs.docker.com/compose/compose-file/compose-file-v3/#build

more over, the following like will give `KeyError` when it's missing (which is the case for everybody)

```
    for it in cnt['build']['volumes']:
```

moreover I'm afraid that introducing build volumes might introduce non-reproducible builds

brejoc:
@muayyad-alsadi Ah, I didn't know that Github was automatically closing things like this! Thanks for looking into this!

PR #212, opened 2020-07-06
Pass environment variables from .env file

akirabaruah:
According to the docker-compose docs[1], environment variables specified
without values should use corresponding values from shell variables or
defaults set in a `.env` file.

[1]https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers

Test: See tests/environment/README.md.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -524,14 +524,22 @@ def container_to_args(compose, cnt, detached=True, podman_command='run'):
     net = cnt.get("network_mode", None)
     if net:
         podman_args.extend(['--network', net])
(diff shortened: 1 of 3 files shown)


madarez:
Bumping this thread up as I also think this is a very great feature parity to add in podman-compose :+1:!

mluhmann:
Bump this feature. Would like to use it with podman-compose as well.

juandspy:
Hoping to see this implemented soon.

daufinsyd:
Hi, any news on this PR ?

PR #237, opened 2020-09-28
Fixes #236: Ensure project name works with podman

### LINKED ISSUE #236: Running podman-compose from inside a .(dot) named directory fails
This is is true for directories not matching `/[a-zA-Z0-9][a-zA-Z0-9_.-]*/`

Ie,
```
$ ~./zsh > podman-comose up
using podman version: podman version 2.0.6
podman pod create --name=.zsh --share net
Error: error running pod create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
125
```

This should be addressed directly [here](https://github.com/containers/podman-compose/blob/20a86ea5631aa1f80ad68259b00e910f546a2509/podman_compose.py#L867) I guess.

helge000:
In case we use dir_basename as podman project name, ensure it matches
podman project name requirement regex: [a-zA-Z0-9][a-zA-Z0-9_.-]*

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -864,7 +864,8 @@ class PodmanCompose:
         os.chdir(dirname)
 
         if not project_name:
-            project_name = dir_basename.lower()
+            # More strict then acually needed for simplicity: podman requires [a-zA-Z0-9][a-zA-Z0-9_.-]*
+            project_name = re.sub(r'[^a-zA-Z0-9]', '', dir_basename)
         self.project_name = project_name
 
 


muayyad-alsadi:
@helge000 do we replace it with empty string or with `-` or `_`

what is the docker-compose behavior?

I prefer if it's non-zero length because I don't want to end with empty string, and I want to see the placeholder -
ex. `n%html` I guess it's better to be `n_html` instead of `nhtml` and `%%%%` should become `____` instead of empty string

what do you think?

helge000:
@muayyad-alsadi  - it already emulates [docker-compose's behaviour](https://github.com/docker/compose/blob/81ce72c95244cebd3711d9b81899e600803d6816/compose/cli/command.py#L176) -  I could have been more clear about in the commit msg I guess...

I basically have no preference; but you certainly have a point if our working dir consists entirely of invalid characters. That said, I guess this is as edge as case can get - maybe we should simply error out here?

After all, I'd go with the current approach to keep the regex plain and simple (I really dislike negative matches).

PR #241, opened 2020-10-08
Allow environment variables to be unset

zqad:
Leaving keys with empty values in YAML will result in the value
ending up being None after parsing the configuration file. This
should result in the variable being imported from the external
environment according to the Compose file version 3 reference.
The resulting action for podman should be an added "-e VAR"
(without =), which is working correctly.

However, when overwriting an external variable by setting it to
e.g. "", the result in docker-compose is that the variable is
unset. For podman, this means adding "-e VAR=". This is not the
case, and this patch does a more strict check to make this case
behave correctly.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -191,7 +191,7 @@ def norm_as_list(src):
     if src is None:
         dst = []
     elif is_dict(src):
-        dst = [("{}={}".format(k, v) if v else k) for k, v in src.items()]
+        dst = [("{}={}".format(k, v) if v is not None else k) for k, v in src.items()]
     elif is_list(src):
         dst = list(src)
     else:


zqad:
Clarifying with an example. Say for example that we have a docker-compose.yml file like so:
```yaml
---
version: '2'
services:
  alpine:
    image: alpine
    command: printenv FOO BAR
    environment:
      FOO:
      BAR: ''
```
We then set the $FOO and $BAR in our local environment:
```
$ export FOO=foo BAR=bar
```
When running docker-compose, $BAR is unset:
```
$ docker-compose up
Recreating ctest_alpine_1 ... done
Attaching to ctest_alpine_1
alpine_1  | foo
alpine_1  | 
ctest_alpine_1 exited with code 0
```
..but with podman-compose, $BAR is still set:
```
$ podman-compose up
using podman version: podman version 2.1.1
podman pod create --name=ctest --share net
645eb9824e55678c1826d7c259b90587d7a2db2418fc40c5297bfd1df24a094f
0
podman create --name=ctest_alpine_1 --pod=ctest --label io.podman.compose.config-hash=123 --label io.podman.compose.project=ctest --label io.podman.compose.version=0.0.1 --label com.docker.compose.container-number=1 --label com.docker.compose.service=alpine -e FOO -e BAR --add-host alpine:127.0.0.1 --add-host ctest_alpine_1:127.0.0.1 alpine printenv FOO BAR
0404c3c92089503c3705a3115ba7f3af1a3ef07a0317b706892e0459c369ddaf
0
podman start -a ctest_alpine_1
foo
bar
0
```
After this patch, however:
```
$ podman-compose up
using podman version: podman version 2.1.1
podman pod create --name=ctest --share net
fc7b751e47ce53387b712e55aa5462eef2889be6ef0537ea2924b70351001b9d
0
podman create --name=ctest_alpine_1 --pod=ctest --label io.podman.compose.config-hash=123 --label io.podman.compose.project=ctest --label io.podman.compose.version=0.0.1 --label com.docker.compose.container-number=1 --label com.docker.compose.service=alpine -e FOO -e BAR= --add-host alpine:127.0.0.1 --add-host ctest_alpine_1:127.0.0.1 alpine printenv FOO BAR
765a089e487149785082b3174db752d092d31d8bc0e4482549236fc1db207103
0
podman start -a ctest_alpine_1
foo

0
```
...$BAR is undefined, yielding the same behaviour as docker-compose.

Also doing a rebase on top of devel.

Issue #249, opened 2020-11-04
support CLI variables COMPOSE_PROJECT_NAME and COMPOSE_FILE


lutostag:
Hey there,

First off wanted to say what a nice great project you all are working on here. I was able to run rootless containers without having to change my compose files at all and get up and working! :guitar: :notes: 

Anyways, I was hoping to be able to use the same environment variables as specified by https://docs.docker.com/compose/env-file/#compose-file-and-cli-variables

The list is here (strikethrough'd ones that probably do not make much sense for podman):
* `COMPOSE_API_VERSION`
* ~`COMPOSE_CONVERT_WINDOWS_PATHS`~
* `COMPOSE_FILE`
* ~`COMPOSE_HTTP_TIMEOUT`~
* ~`COMPOSE_TLS_VERSION`~
* `COMPOSE_PROJECT_NAME`
* ~`DOCKER_CERT_PATH`~
* ~`DOCKER_HOST`~
* ~`DOCKER_TLS_VERIFY`~

I have at least used `COMPOSE_FILE` and `COMPOSE_PROJECT_NAME` a bit in some makefiles in current projects.

I am able to add support for those two in a PR, if that would be helpful to anyone else.

(Note that I do not need to load from the `.env` file exactly in my case, just having the defaults set from those env variables would be sufficient for my usecase).

Open to any feedback on the idea before offering a PR :smile_cat: 

Thanks again.

muayyad-alsadi:
please confirm that it works now.

muayyad-alsadi:
sorry, I did not see that you offered a PR. but there is a TODO, you might want to work on it

```
        # TODO: should read and respect those env variables
        # see: https://docs.docker.com/compose/reference/envvars/
        # see: https://docs.docker.com/compose/env-file/
        dotenv_dict.update({
            "COMPOSE_FILE": filename,
            "COMPOSE_PROJECT_NAME": self.project_name,
            "COMPOSE_PATH_SEPARATOR": ":",
        })
```

this means that we need to replace

```
            args.file = list(filter(os.path.exists, [
                "docker-compose.yml",
                "docker-compose.yaml",
                "docker-compose.override.yml",
                "docker-compose.override.yaml",
                "container-compose.yml",
                "container-compose.yaml",
                "container-compose.override.yml",
                "container-compose.override.yaml"
            ]))
```

with something like

```
compose_file_ls = os.environ.get("COMPOSE_FILE", "docker-compose.yml:container-compose.yml").split(":")
[...]

PR #250, opened 2020-11-04
Support COMPOSE_* env vars as default cli flags

lutostag:
Support for env variables setting default values as explained in:
https://docs.docker.com/compose/reference/envvars/ currently only
two that seem to make sense are COMPOSE_FILE and COMPOSE_PROJECT_NAME.

Relates to #249


d3xter:
@muayyad-alsadi @lutostag Any updates on this PR?

Its very annoying to always run `podman-compose -p my-project start' every time

PR #260, opened 2020-12-21
Add support external volumes

igk1972:
Example docker-compose.yml

```yaml
version: '3'
services:
  test:
     volumes:
        - data:/data
volumes:
  data:
     name: some
     external: true
```

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -130,8 +130,12 @@ def fix_mount_dict(mount_dict, proj_name, srv_name):
                 hashlib.sha256(mount_dict["target"].encode("utf-8")).hexdigest(),
             ])
         else:
-            # prefix with proj_name
-            mount_dict["source"] = proj_name+"_"+source
+            name = mount_dict.get('name')
+            if name:
+                mount_dict["source"] = name
+            else:
+                # prefix with proj_name
+                mount_dict["source"] = proj_name+"_"+source
     return mount_dict
 
 # docker and docker-compose support subset of bash variable substitution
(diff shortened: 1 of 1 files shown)


podman_compose.py
```diff
@@ -376,7 +380,7 @@ def assert_volume(compose, mount_dict):
     inspect volume to get directory
     create volume if needed
     """
-    if mount_dict["type"] != "volume": return
+    if mount_dict["type"] != "volume" or mount_dict["external"]: return
```
jomority, line 383: Here I get a `KeyError: 'external'` for some volumes.
I changed it to `mount_dict.get("external")` and it seems to work, but maybe there is a better way to produce a boolean.

podman_compose.py
```diff
@@ -480,6 +484,19 @@ def get_mount_args(compose, cnt, volume):
     basedir = compose.dirname
     if is_str(volume): volume = parse_short_mount(volume, basedir)
     mount_type = volume["type"]
+
+    vol = None
+    source = volume.get('source')
+    if source:
+      vol = compose.shared_vols[source]
```
jomority, line 491: For `{'type': 'bind', ...}` this throws a KeyError.
I wrapped this whole block in a `if mount_type == 'volume':`. Not sure though if it has other side effects.

jomority (commented):
Thanks for this pull!
I struggled for hours trying to get volumes from one pod mounted into another. This patch is the solution.

However, I got some errors. I commented my solutions for them inline.

PR #275, opened 2021-02-12
Move Python installation and test deps to requirement files

radeksm:
Installation with setup.py fails on missing yaml package, which should
be installed by "install_requires" but this also does not work because
podman_compose.__version__ also needs yaml.
Put all the Python dependencies into separate files:
    - requirements.txt
    - test-requirements.txt

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,14 +40,4 @@ setup(
     },
     include_package_data=True,
     license='GPL-2.0-only',
-    install_requires=[
-        'pyyaml'
-    ],
-    # test_suite='tests',
-    # tests_require=[
-    #     'coverage',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -40,14 +40,4 @@
...
-    install_requires=[
-        'pyyaml'
-    ],
-    # test_suite='tests',
-    # tests_require=[
-    #     'coverage',
-    #     'pytest-cov',
-    #     'pytest',
-    #     'tox',
-    # ]
```
tonyseek, line 52: This breaks pipx and a various of tools which invoke setuptools.

muayyad-alsadi: ok, check now

PR #280, opened 2021-02-28
literals

yuvalk:
this is a replacement to the stale and now conflicting pr#97

- Test variable interpolation in the YAML
- Fix variable escaping with $$
- Force adding an .env file for the tests

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -146,25 +146,19 @@ def fix_mount_dict(mount_dict, proj_name, srv_name):
 # ${VARIABLE?err} raise error if not set
 # $$ means $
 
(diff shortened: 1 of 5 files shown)


yuvalk:
was this resolved in a different way?

muayyad-alsadi:
maybe it was closed because I promoted `1.x` branch to be the new devel instead of `0.1.x`

I'll look into it.

muayyad-alsadi:
yes, it seems it was fixed by 

```
commit 00840d0613f4eb689975bc214f9b1883fc15dfe4 (garotosopa-devel)
Date:   Thu May 6 00:49:42 2021 +0300
```

the .env part is not yet done because I was under-impression it's correct see #370.

PR #289, opened 2021-04-22
Implement --exit-code-from

tjikkun:
This implements `--exit-code-from`. It switches from `threading.Thread` to
`concurrent.futures.ThreadPoolExecutor` because `threading.Thread` does not
support return codes. A downside is that `ThreadPoolExecutor` waits for
threads to exit, with no nice way to avoid that, so we run `podman stop`
for every container before exiting.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -19,7 +19,7 @@ import hashlib
 import random
 import json
 
-from threading import Thread
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
 
 import shlex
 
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
please test this

https://github.com/containers/podman-compose/commit/045cef299bec77e3ae2d1980ad32c1fd2c2896bf

does this solves your problem?

muayyad-alsadi:
I've added a fix and a simple test

> Also, the other containers are not stopped.

currently we don't stop containers.

PR #292, opened 2021-04-26
Add `podman-compose exec` support

dwt:
As it was missing and I needed it. :-)

Argument generation is quite rough, but that seemed to be the quickest way to get it. This supports all arguments supported by podman 1.6.4 - as that is my use case.

What do you think?


dwt:
Thanks!

PR #295, opened 2021-05-05
Prefer 'compose.yaml' as per compose spec

helge000:
As per [compose spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#compose-file), we should at least support, or better yet prefer `compose.yaml` (as implemented here):

>The default path for a Compose file is compose.yaml (preferred) or compose.yml in working directory

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -837,6 +837,10 @@ class PodmanCompose:
         cmd = args.command
         if not args.file:
             args.file = list(filter(os.path.exists, [
+                "compose.yaml",
+                "compose.yml",
+                "compose.override.yaml",
+                "compose.override.yml",
                 "docker-compose.yml",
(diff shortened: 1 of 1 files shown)


podman_compose.py
```diff
@@ -846,9 +850,9 @@ def _parse_compose_file(self):
                 "container-compose.override.yml",
                 "container-compose.override.yaml"
             ]))
-        files = args.file
+        files = args.file[0]
```
helge000, line 853: I guess you would rather have me remove this; however I strongly believe this is the right thing to do as it could lead to unexpected results. That said, this discussion would be best for another PR, then :)

muayyad-alsadi: yes, please remove this.

PR #320, opened 2021-07-21
Add support for secrets

samarthj:
Adds support for -
- (1) Declared secrets with the file location.
- (2) Declared secrets with file location, mounted as a different named secret.
- (3) Declared secrets with file location, mounted at arbitrary location.
- (4) External secrets (type=mount), mounted as original secret name.
- (5) External secrets (type=mount), mounted as original secret name, with specified uid, gid and mode.

Additionally added compose files for testing the positive and negative test-cases for the new functionality.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -513,6 +513,74 @@ def get_mount_args(compose, cnt, volume):
         args = mount_desc_to_mount_args(compose, volume, srv_name, cnt['name'])
         return ['--mount', args]
 
+
+def get_secret_args(compose, cnt, secret):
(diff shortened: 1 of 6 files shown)


samarthj:
Please keep in mind that compared to docker (4) and (5) are the only differentiating factor. They are only supported in swarm mode. However there is no such mode available for podman which maps 1:1 with docker. And since it was fairly trivial to add support for them, I added it. If it does not conform to the guidelines and adds support for something that should not exist, I am happy to remove (4) and (5).

Issue #297, opened 2021-05-12
Podman run log-driver option


jd-daniels:
I can't seem to find a way to pass --log-driver to the podman-compose call to podman via podman up. Is this supported?

From podman-run man page:

   --log-driver="driver"
       Logging driver for the container. Currently available options are k8s-file, journald, and none, with json-file aliased to k8s-file for scripting compatibility.

Thanks!

muayyad-alsadi:
I guess this can be solved with this proposal


https://github.com/containers/podman-compose/issues/278

muayyad-alsadi:
please test

jd-daniels:
This won't quite work for log-driver because podman-compose seems to call podman pod and podman start, not just podman run, the other two don't have a log-driver option.

Example:
[]$ podman-compose --podman-args="--log-driver=journald" up -d
['podman', '--version', '']
using podman version: podman version 2.2.1
podman pod --log-driver=journald create --name=deploy --share net -p 8443:8443 -p 443:8443
Error: unknown flag: --log-driver
125
<podman run redact>
Error: error retrieving pod deploy: no pod with name or ID deploy found: no such pod
125
podman start --log-driver=journald deploy_idp_1
Error: unknown flag: --log-driver
125

g-braeunlich:
Any update on this?

muayyad-alsadi:
Use --podman-<COMMAND>-args

g-braeunlich:
What about respecting the `logging` config for services as referenced [here](https://docs.docker.com/compose/compose-file/compose-file-v3/#logging)?
There are not many options to choose from currently for the driver:
* `k8s-file`
* `journald`
* `none`
Same for the dirver options:
* `path`
* `max-size`
* `tag` (journald only)

If you want, I could create a PR.

muayyad-alsadi:
Yes please

PR #321, opened 2021-07-27
Log

g-braeunlich:
log support + some pylint fixes. Feel free to reject if this is not what you want (in this case, I would suggest to update `.pylintrc`)

Closes: #297


muayyad-alsadi:
Can we have log only? And then we deal with pylint in another PR

muayyad-alsadi:
Thank you

g-braeunlich:
You are welcome!

PR #330, opened 2021-09-02
Fix `up` arguments parsing

abitrolly:
These is already parsed here.

https://github.com/containers/podman-compose/blob/502d7cc206da1f20cc9aef480f93f6b32d576f80/podman_compose.py#L1644


ret2libc:
I was going to submit exactly this. Awesome! Unfortunately specifying services is not implemented anyway :(

abitrolly:
Yep. But at least now it is obvious.

StevenDias33 (approved):

muayyad-alsadi:
> Unfortunately specifying services is not implemented anyway :(

it's implemented now
please help me test

abitrolly:
@muayyad-alsadi `podman-compose up <service>` works for me. Thanks.

Issue #315, opened 2021-07-09
Doc:  --transform_policy/-t option is missing default


rugk:
There is the help output…

```
$ podman-compose --help
usage: podman-compose [-h] [-f file] [-p PROJECT_NAME] [--podman-path PODMAN_PATH] [--no-ansi] [--no-cleanup] [--dry-run] [-t {1pod,1podfw,hostnet,cntnet,publishall,identity}]
                      {help,version,pull,push,build,up,down,ps,run,start,stop,restart,logs} ...

optional arguments:
  -h, --help            show this help message and exit
  -f file, --file file  Specify an alternate compose file (default: docker-compose.yml)
  -p PROJECT_NAME, --project-name PROJECT_NAME
                        Specify an alternate project name (default: directory name)
  --podman-path PODMAN_PATH
                        Specify an alternate path to podman (default: use location in $PATH variable)
  --no-ansi             Do not print ANSI control characters
  --no-cleanup          Do not stop and remove existing pod & containers
  --dry-run             No action; perform a simulation of commands
  -t {1pod,1podfw,hostnet,cntnet,publishall,identity}, --transform_policy {1pod,1podfw,hostnet,cntnet,publishall,identity}
                        how to translate docker compose to podman [1pod|hostnet|accurate]

command:
  {help,version,pull,push,build,up,down,ps,run,start,stop,restart,logs}
    help                show help
    version             show version
    pull                pull stack images
    push                push stack images
    build               build stack images
    up                  Create and start the entire stack or some of its services
    down                tear down entire stack
    ps                  show status of containers
    run                 create a container similar to a service to run a one-off command
    start               start specific services
    stop                stop specific services
    restart             restart specific services
    logs                show logs from services
```

…and [there is this doc in the repo](https://github.com/containers/podman-compose/blob/devel/docs/Mappings.md) (BTW is that published on GItHub Pages or so somewhere, as it is hard to find in a repo?)

Especially the last document is nice as it explains all those cryptic transform policy names (:see_no_evil:), however, it fails to mention one thing: **What is actually the default mode?**

From my experience I guess it is `1pod`, but it says  "(does not work)" in the doc at least… so I'm somewhat confused. :sweat_smile: 

## System

```
$ podman-compose version
using podman version: podman version 3.2.0
podman-composer version  0.1.7dev
podman --version
podman version 3.2.0
0
$ podman version
Version:      3.2.0
API Version:  3.2.0
Go Version:   go1.16.3
[...]

PR #332, opened 2021-09-06
docs: added the transform_policy default and description

samarthj:
Added the mapping doc details in the help as well. Just so it is apparent what each choice does. Considered it overkill at first, but felt that it might serve to have the description directly in the help, since it isn't too large.

Also removed the "[1pod|hostnet|accurate]". Felt it might create confusion with the referenced options.


samarthj:
fixes #315

PR #369, opened 2021-12-03
fix AttributeError when running a one-off command

ender672:
Without this, I get errors when running "podman-compose -p podname run".


ender672:
I started getting the AttributeError when the official podman-compose package was updated in Fedora 35.

PR #372, opened 2021-12-13
Starting/stopping/restarting all services and viewing logs, add all log arguments

AviDuda:
Howdy! I missed being able to start/stop/restart all services just like in docker-compose, so this PR takes care of it.
I'm not honestly sure if the reversing the service list for stopping and restarting is 100% correct. I found something like this in `compose_down` so hopefully it's good enough.

I also noticed the logs command doesn't allow viewing all logs at once so the change is included too, along with some missing arguments for the logs command.
I split the commits in case you want to have them in a different PR.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1575,10 +1575,13 @@ def compose_restart(compose, args):
 @cmd_run(podman_compose, 'logs', 'show logs from services')
 def compose_logs(compose, args):
     container_names_by_service = compose.container_names_by_service
-    target = None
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
> I'm not honestly sure if the reversing the service list for stopping and restarting is 100% correct. I found something like this in compose_down so hopefully it's good enough.

yes indeed, it's correct. if B depends on A, A should be started first and stopped last.

things seems to be fine, just please test it very well, with up, run, exec, ...etc.

PR #389, opened 2021-12-29
Fix substitution for empty variables

mohd-akram:
See https://docs.docker.com/compose/environment-variables/


muayyad-alsadi:
would you please provide a comment with an example yaml and envs.

mohd-akram:
```yaml
version: '2.1'
services:
  web:
    environment:
      - USERNAME=${USERNAME:?user}
      - PASSWORD=${PASSWORD:-pass}
```
```shell
# should fail
USERNAME= podman-compose --dry-run up
# should print PASSWORD=pass
USERNAME=user PASSWORD= podman-compose --dry-run up 2>&1 | grep -o 'PASSWORD=[a-z]*'
```

PR #392, opened 2021-12-30
Allow empty default/error value in substitution

mohd-akram:
Looks like #174 was undone in https://github.com/containers/podman-compose/commit/00840d0613f4eb689975bc214f9b1883fc15dfe4.

Merge #389 before this.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -179,8 +179,8 @@ var_re = re.compile(r"""
         (?:{
             (?P<braced>[_a-zA-Z][_a-zA-Z0-9]*)
             (?:(?P<empty>:)?(?:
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
would you please update this to fix the conflict 
and provide an example test

mohd-akram:
```yaml
version: '2.1'
services:
  web:
    environment:
      - USERNAME=${USERNAME:?}
      - PASSWORD=${PASSWORD:-}
```

```shell
# should fail
podman-compose --dry-run up
# should print nothing
USERNAME=user podman-compose --dry-run up | grep -F '${PASSWORD:-}'
```

Issue #106, opened 2020-02-06
Missing `config` command


yajo:
To debug your `docker-compose.yml` file it's very useful the `docker-compose config` command, which outputs the final configuration that gets used by docker-compose, including all merges, variable expansions, etc.

podman-compose is missing that subcommand.

pleonex:
This is also required to make it work with VS Code Remote Container extensions. It tries to run this command (after symlink to podman-compose):

```
docker-compose -f /myrepo/.devcontainer/docker-compose.yml config --services
```

mabulgu:
Hi. Is someone working on this issue? If not, I would love to start working on this.

PR #394, opened 2021-12-30
Add necessary commands for VSCode Devcontainers

howlowck:
Changes:
* Added `short` and `format` args to `version` command
* Added a trivial implementation of the `config` command which echos out the content when given an absolute path of a file

Resolves:
* microsoft/vscode-remote-release#5553
* #106

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,5 +70,4 @@ def compose_up(compose, args):
   scale              Set number of containers for a service
   top                Display the running processes
   unpause            Unpause services
-  version            Show the Docker-Compose version information
 ```
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1280,13 +1280,13 @@ class cmd_parse:
 
 @cmd_run(podman_compose, 'version', 'show version')
 def compose_version(compose, args):
-    # log("podman-composer version", __version__)
     if getattr(args, 'short', False): 
         print(__version__)
         return
     if getattr(args, 'format', 'pretty') == 'json':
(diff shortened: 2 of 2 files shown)


muayyad-alsadi:
does vscode expects
version on stdout or stderr?

the following line should be replaced with `log(...)` (it just a helper that just print to stderr)

```
        sys.stderr("File path is required. use -f flag to set the input file")
...
    if (file_path == None):
        sys.stderr("File path is required. use -f flag to set the input file")
        return
```

even more, this should be part of parsing, which is handled by `argparse` not need at all to process it

```
parser.add_argument("-f", "--format", type=str, default='pretty', 
        help="Format the output. Values: [pretty | json].")
```

should have choices like this

```
parser.add_argument("-f", "--format", choices=['pretty', 'json'], default='pretty'
                            help="Format the output")
```

please don't construct json by substitution and concatenation and replace, use `json.dumps`

```
        print('{ "version": "{version}" }'.replace('{version}', __version__))
```

should be

```
        res = {"version": __version__}
        print(json.dumps(res))
```

howlowck:
Hey @muayyad-alsadi, thanks for the quick review. vscode expects stdout. I will make the suggested updates.

howlowck:
Updated. Thanks again!

muayyad-alsadi:
I'll do some changes and need your help testing them.

muayyad-alsadi:
I've made config prints the merged yaml of multiple files
please test

howlowck:
Yes. Multiple files are merging properly.

PR #419, opened 2022-02-03
bring colors into container logging from unmerged pr #192

chutchi2:
Signed-off-by: Cody Hutchins <codyhutchins@seed-innovations.com>

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -876,7 +876,7 @@ class Podman:
         log(cmd_ls)
         return subprocess.check_output(cmd_ls)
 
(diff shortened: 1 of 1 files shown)


chutchi2:
Created PR because #192 appears to have achieved this feature perfectly but was never merged. I believe it is a simple addition and of use to any users both familiar with docker-compose behavior and not.

bluflowr:
This would definitely add value. Looks good.

muayyad-alsadi:
I've tested it with
https://github.com/containers/podman-compose/blob/devel/tests/testlogs/docker-compose.yml

and it seems working fine

josejsarmento:
Any ETA for when this will be merged into stable branch?

Issue #130, opened 2020-03-25
podman-compose gets confused with symlinks and extends


sedrubal:
I have this repo structure:

#### Structure:

```
.
├── project1
│   ├── docker-compose.yml
│   └── Dockerfile
├── project2
│   ├── docker-compose.yml
│   └── Dockerfile
├── metarepo
│   └── docker-compose.yml
└── docker-compose.yml -> ./metarepo/docker-compose.yml
```

And this main `docker-compose.yml`: 

#### `docker-compose.yml`

```docker-compose.yml
---
version: '2.4'

services:
  project1:
    extends:
      file: project1/docker-compose.dev.yml
      service: project1

  project2:
    extends:
      file: project2/docker-compose.dev.yml
      service: project2
...
```

A project-`docker-compose.yml` looks like this;

```docker-compose.yml
---
version: "2.4"

services:
    project1:
        build: .
...
```

While this works with `docker-compose`, `podman-compose` gets confused. AFAIK there are multiple issues:

1. `podman-compose` uses `os.path.realpath` instead of `os.path.abspath` in the line below to resolve the directory of the main `docker-compose.yml`. Thus `podman-compose` operates in `metarepo/` directory instead of the main directory.
https://github.com/containers/podman-compose/blob/efcbc75f631e693791a7f71abc49d9d05559ae92/podman_compose.py#L791
2. `podman-compose` does not resolve the build path to `project1/` / `project2/`. I think the `extends` logic is broken...

LyricPants66133:
I am also having this issue when trying to move a project from docker to podman. I received error messages described in #268.

muayyad-alsadi:
> podman-compose uses os.path.realpath instead of os.path.abspath 

please send me PR and I'll merge it.

vansari:
Hey @muayyad-alsadi 

I have created a PR for this issue. Can you please have a look at it?

Thanks a lot.

muayyad-alsadi:
although I've some comments, I'll merge it now and comment later when I get time.

I'll keep this open so that you check again when I'm done.

filipef9:
When this fix will be released?

muayyad-alsadi:
@filipef9 help us test latest push

jcmsys-github-link-terraformcloud:
I'd like to bump this issue, as it's a substantial blocker for moving certain workflows to podman-compose.

[...]

PR #433, opened 2022-02-20
[FIX] extended compose file with build context not working

vansari:
Hi @muayyad-alsadi and all other contributers,

With this PR I want to fix the issue with the extended Build-Context which is not working at this time.

I created a sample folder with following structure:

``` 
extends_w_file_subdir
 |--> docker-compose.yml
 |--> sub
   |--> docker-compose.yml
   |--> docker
     |--> example
       |--> Dockerfile
```

If I tried the cmd `./podman_compose.py -f tests/extends_w_file_subdir/docker-compose.yml up -d` I received the message that the docker/example/Dockerfile can not be found.

The goal is to prefix the context path in the extended File with path of the extending File.
I have also wrote a little pytest to check that the container was build successful.

Please can you check if this can solve some reported issues?
Thanks a lot!

PS: Can solve the Issue [#130](https://github.com/containers/podman-compose/issues/130)

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -148,7 +148,7 @@ def fix_mount_dict(compose, mount_dict, proj_name, srv_name):
         vols = compose.vols
         source = mount_dict.get("source", None)
         vol = (vols.get(source, None) or {}) if source else {}
-        name = vol.get('name', None) 
+        name = vol.get('name', None)
         mount_dict["_vol"] = vol
         # handle anonymouse or implied volume
         if not source:
@@ -928,7 +928,15 @@ class Podman:
         volumes = output.splitlines()
         return volumes
 
-def normalize_service(service):
+def normalize_service(service, sub_dir=''):
+    if sub_dir and 'build' in service and 'context' in service['build']:
+        context = service['build']['context']
+        if re.match('^(((\./)?(\w+)+(/\w+)+?$)|(\.$))', context):
+            if context.startswith('.'):
+                context = context.replace('.', sub_dir)
+            else:
+                context = sub_dir + os.sep + context
+            service['build']['context'] = context
     for key in ("env_file", "security_opt", "volumes"):
         if key not in service: continue
         if is_str(service[key]): service[key]=[service[key]]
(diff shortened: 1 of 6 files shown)


filipef9:
When this fix will be released?

podman_compose.py
```diff
@@ -928,7 +928,15 @@ def volume_ls(self, proj=None):
         volumes = output.splitlines()
         return volumes
 
-def normalize_service(service):
+def normalize_service(service, sub_dir=''):
+    if sub_dir and 'build' in service and 'context' in service['build']:
+        context = service['build']['context']
+        if re.match('^(((\./)?(\w+)+(/\w+)+?$)|(\.$))', context):
```
muayyad-alsadi, line 934: I typically compile my regex once, not per-function call.

podman_compose.py
```diff
@@ -928,7 +928,15 @@ def volume_ls(self, proj=None):
...
+            if context.startswith('.'):
+                context = context.replace('.', sub_dir)
+            else:
+                context = sub_dir + os.sep + context
```
muayyad-alsadi, line 938: I typically use os.path.join, is their a reason for noting using it?

muayyad-alsadi, line 938: I typically use os.path.join, is their a reason for noting using it?

podman_compose.py
```diff
@@ -1008,9 +1016,12 @@ def resolve_extends(services, service_names, environ):
                 content = yaml.safe_load(f) or {}
             if "services" in content:
                 content = content["services"]
+            subdirectory = filename.rsplit(os.sep, 1)[0]
```
muayyad-alsadi, line 1019: should this be replace with os.path.basename or os.path.dirname?

podman_compose.py
```diff
@@ -928,7 +928,15 @@ def volume_ls(self, proj=None):
         volumes = output.splitlines()
         return volumes
 
-def normalize_service(service):
+def normalize_service(service, sub_dir=''):
+    if sub_dir and 'build' in service and 'context' in service['build']:
```
muayyad-alsadi, line 932: does not seem to handle the case were context is passed as string build like this

```yaml
services:
    web1:
      build: ./context
```

podman_compose.py
muayyad-alsadi: I believe the regex is not needed at all, it should be just `.startswith('./')`

vansari: you can hit the path to the context without a './'

podman_compose.py
vansari: I will check it

vansari:
@muayyad-alsadi I will create a new PR today with the "fixes" and a newer test for it

muayyad-alsadi:
Please wait. I'm pushing my changes.

muayyad-alsadi:
@vansari please test my last push and I really appreciate your contribution.

vansari:
@muayyad-alsadi Nice work, I have updated my tests in another PR :)

PR #438, opened 2022-02-25
Bugfix/fixing tests

vansari:
Fixing only Tests


muayyad-alsadi:
Is this included in the following commit?

https://github.com/containers/podman-compose/commit/91a579b81e199c44795c99760fdc6edbadaf0265

is so let's close this. if not please open a new one.

PR #470, opened 2022-04-06
Support platform property

shanesmith:
As per https://github.com/compose-spec/compose-spec/blob/master/spec.md#platform

Example:

```
services:
  mysql:
    image: mysql:5.7
    platform: linux/x86_64
```


podman_compose.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/python3
+#! /usr/bin/env python3
```
muayyad-alsadi, line 1: Please keep this python3 not env

shanesmith: oops, my bad, fixed!

shanesmith:
The pylint failures seem to come from a previous PR and are fixed in https://github.com/containers/podman-compose/pull/468

PR #472, opened 2022-04-07
Use env to find a suitable python3

ankon:
This helps when you have multiple of those installed, for example on macOS both the one in /usr/bin and the one from homebrew in /usr/local/bin. env will work out the one you prefer to use.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1,4 +1,4 @@
-#! /usr/bin/python3
+#! /usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 # https://docs.docker.com/compose/compose-file/#service-configuration-reference


muayyad-alsadi:
I have a question, when you use pip3 I believe it will handle this properly

```python
# setup.py
...
entry_points={"console_scripts": ["podman-compose = podman_compose:main"]},
```

my script is placed as a module in whatever place python can find it
and a script is created by pip that imports it
the pip-created script would have the proper python path (the same one used to run pip).

ankon:
So actually what I did was install podman-compose with pip3, which used /usr/local/bin/pip3 as I wanted. 

Then I figured that I need the devel version (because changes), and installed that they way it was suggested: Make it executable and run it. That did not work, because now it went to the wrong python3. I really don't want to use `/usr/bin/python3`.

Do you think the change would "hurt" other users? Or is the problem here that the docs need updating, and the expectation that you would install it such that it works?

muayyad-alsadi:
you can install the devel branch using pip like this

```
pip3 install https://github.com/containers/podman-compose/archive/devel.tar.gz
```

> Do you think the change would "hurt" other users? 

I can't remember the full reasoning behind not using env
but I remember it's bad.

this article might help

https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_interpreter_invocation

consider this scenario

if you setup virtual env to develop some old python2 django project
but have python3 system wide ..etc.
if you activate your django venv you will broke podman-compose

muayyad-alsadi:
> It doesn't say it's bad in general, just that a user installing an app from a OS package expects to get an app that uses the OS packaged python (i.e. user doesn't need to know that this is python!).

It's their use case but there other use cases
I don't have an article about general.

But I gave you a general use case that is not related to RPM

you have an active python let's say /usr/bin/python37 and you installed py yaml and podman compose
Then you activated a venv on python38 to work on django
On your 38 there is no py yaml
So podman compose is broken
It should continue to use py37 which was used to install it not the active one

I believe there are some security concerns that a system wide tool should not  be poisoned by user defined env ...etc

Again I can't remember full reasoning behind this.

But this is my choice.

I'll update the readme

muayyad-alsadi:
> That doesn't give me the local sources for changing though. :)

I don't get it. If you want to install it without root.

```
pip install --user path-or-url.tar.gz
```

ankon:
I want to do

```sh
git clone https://github.com/containers/podman-compose && podman-compose
vi podman_compose.py # fix bugs, improve code, add features
```

And I want to be able to use that modified `podman_compose.py` with my python, without having to keep a local change to the shebang. Ideally just by calling `podman-compose` on the command-line for all my compose needs.

muayyad-alsadi:
I got your point.

for this case you can do what pip do
you create a `podman-compose` file with two lines in it

```
#! /usr/bin/env python3
import podman_compose
podman_compose.main()
```

ankon:
Right, that helps, thanks!

PR #473, opened 2022-04-07
Added port subcommand.

eyzeedev:
Doesn't deal with port ranges, and needs some work, but really just opening up a dialogue to drive through the rest of the implementation.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -12,6 +12,7 @@ import sys
 import os
 import getpass
 import argparse
+import itertools
 import subprocess
 import time
 import re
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
It seems that you undo some other work. You would you update your branch from latest devel

podman_compose.py
```diff
@@ -904,9 +905,6 @@ def container_to_args(compose, cnt, detached=True):
         if is_str(entrypoint):
             entrypoint = shlex.split(entrypoint)
         podman_args.extend(["--entrypoint", json.dumps(entrypoint)])
-    platform = cnt.get("platform", None)
```
muayyad-alsadi, line 907: Why remove this?

eyzeedev:
Rebased onto upstream devel with no changes. Reinstated accidental deletion.

eyzeedev:
It remains to implement port-ranges and indexes.

Issue #307, opened 2021-06-17
add systemd service files


tinywrkb:
It would be great if podman-compose could be shipped with systems services files.

I'm attaching my systemd service files. They are pretty basic, probably need to be improved so I'm not adding these myself in a PR, but they are working for me.  

_/usr/lib/systemd/system/podman-compose@.service_

```
[Unit]
Description=%i rootful pod (podman-compose)
After=network-online.target
Wants=network-online.target

[Service]
Type=oneshot
RemainAfterExit=true
WorkingDirectory=/etc/containers/compose/%i
ExecStart=/usr/bin/podman-compose up -d --remove-orphans
ExecStop=/usr/bin/podman-compose down

[Install]
WantedBy=multi-user.target
```

_/usr/lib/systemd/user/podman-compose@.service_
```
[Unit]
Description=%i rootless pod (podman-compose)

[Service]
Type=oneshot
RemainAfterExit=true
WorkingDirectory=%h/.config/containers/compose/%i
ExecStart=/usr/bin/podman-compose up -d --remove-orphans
ExecStop=/usr/bin/podman-compose down

[Install]
WantedBy=default.target
```

Previously I was using environment file in `/etc/conf.d` to set the compose file, but now that podman-compose defaults prioritize `compose.yaml` and `compose.yml`, I don't see a reason why we need this configurable.

If it's not obvious, each pod should have a folder in `/etc/containers/compose/` ~~and~~ **(edit:) or** `~/.config/containers/compose/`.

mathstuf:
I feel like `podman pod` has this now. Maybe the ability to create a pod from a compose file would be useful?

muayyad-alsadi:
the latest push
80e852717d9598ed95dd54b9c7b0ba7c99b6b5f5

creates a pod and put containers in it and passes proper dependencies `--requires`
so `podman pod start pod_<project_name>` will work fine.

please test that commit

muayyad-alsadi:
I'm considering a flag like `--register` 

```
podman-compose --register up --no-start
```

which would write `~/.config/containers/compose/<PROJECT>.env`

so it can be used like this

```
# /usr/lib/systemd/user/podman-compose@.service

[Unit]
Description=%i rootless pod (podman-compose)

[Service]
Type=simple
EnvironmentFile=%h/.config/containers/compose/%i.env
ExecStartPre=/usr/bin/podman-compose up --no-start
ExecStartPre=/usr/bin/podman pod start pod_%i
[...]

PR #477, opened 2022-04-16
Respect XDG_CONFIG_HOME when generating systemd unit

Ram-Z:
Addresses https://github.com/containers/podman-compose/issues/307#issuecomment-1078984418.

Possibly a better approach would be to let package maintainers install this unit file in /usr/systemd/user, but this would break the "single file" design.


muayyad-alsadi:
I would love if we add `--global` option as in the referenced comment.
I would love if we keep the `sudo` in the help.

Ram-Z:
> I would love if we add --global option as in the referenced comment.

The `--global` option only makes sense with `-a create-unit`, I wasn't sure how to introduce it nicely. Happy to take advice on how you'd like this handled.

However, I'm not convinced it should exist at all. `podman-compose` wants to be rootless, providing an easy way to install the systemd unit to the users home directory only is perfectly fine.

> I would love if we keep the sudo in the help.

The point of this MR is to install the user systemd unit in the users home directory where `sudo` is not needed.

muayyad-alsadi:
> However, I'm not convinced it should exist at all. podman-compose wants to be rootless, providing an easy way to install the systemd unit to the users home directory only is perfectly fine.


a typical valid use case is to install `podman-compose` rootful and used rootless.

PR #481, opened 2022-04-23
Add manpage

athos-ribeiro:
This initial manpage was based on the output of help2man.

Signed-off-by: Athos Ribeiro <athoscribeiro@gmail.com>


muayyad-alsadi:
please ship it in your debian diff. I can't keep it updated because we are in early development and we keep change things.

athos-ribeiro:
> please ship it in your debian diff. I can't keep it updated because we are in early development and we keep change things.

Thanks for the quick reply here!

Fair enough. I will maintain it downstream then.

Issue #206, opened 2020-06-28
Podman-compose - Nextcloud - (98)Address already in use: AH00072: make_sock: could not bind to address 0.0.0.0:80


osirase:
So this only occurs when running in my fully populated docker-compose.yml file. 
It works when it is launched on its own just with podman:
```
podman run -d \
-p 1337:80 \
-v ~/containerVolumes/nextcloud/:/var/www/html:z \
-v ~/containerVolumes/nextcloud/apps:/var/www/html/custom_apps:z \
-v ~/containerVolumes/nextcloud/config:/var/www/html/config:z \
-v ~/containerVolumes/nextcloud/data:/var/www/html/data:z \
-v ~/containerVolumes/nextcloud/theme:/var/www/html/themes:z \
nextcloud
```
And in a docker-compose.yml file on its own.
```
❯ cat docker-compose.yml
---
version: "2.0"
services:
  nextcloud:
    image: nextcloud
    container_name: nextcloud
    ports:
      - 1000:80
    volumes:
      - ./nextcloud:/var/www/html:z
      - ./nextcloud/apps:/var/www/html/custom_apps:z
      - ./nextcloud/config:/var/www/html/config:z
      - ./nextcloud/data:/var/www/html/data:z
      - ./nextcloud/themes:/var/www/html/themes:z
    restart: unless-stopped

```
However, when run with my full docker-compose.yml:
```
---
version: "3.0"
services:
  nginx:
    image: nginx
    container_name: nginxReverseProxy
    ports:
      - 8080:80
      - 4443:443
    volumes:
      - ./nginx/conf:/etc/nginx:ro
      - ./nginx/html:/usr/share/nginx/html:ro
      - ./nginx/log:/var/log/nginx:z
#      - ./nginx/pki
    restart: unless_stopped
  jellyfin:
    image: linuxserver/jellyfin
    container_name: jellyfin
    environment:
      - PUID=1000
      - PGID=1000
#      - UMASK_SET=022 #optional
    volumes:
      - ./jellyfin/library:/config:z
      - ./jellyfin/tvseries:/data/tvshows:z
      - ./jellyfin/movies:/data/movies:z
#      - /opt/vc/lib:/opt/vc/lib #optional
    ports:
      - 8096:8096
      - 8920:8920 #optional
#    devices:
#      - /dev/dri:/dev/dri #optional
#      - /dev/vcsm:/dev/vcsm #optional
#      - /dev/vchiq:/dev/vchiq #optional
[...]

PR #483, opened 2022-04-24
Add argument --multiple-pods

SadPencil:
This argument will create a pod for each container. This is a useful feature, because, by definition, all containers in a Podman pod share the same network namespace. Sometimes containers in a compose file listen to the same port and then bind error happens, which is not a problem in docker-compose.

This PR can be a temporary workaround for issue #206 (but it is just a workaround, so please don't close the issue)


SadPencil:
Oops. I need to close this PR as more issues brought by multiple pods.

PR #493, opened 2022-05-05
packaging: Move version definition to setup.cfg

zmc:
An unfortunate side-effect of defining the version inside a project's
main module is that dependency issues can break installation. In this
project, even though pyyaml and python-dotenv are listed in
install_requires, installation would fail if either were not present,
because the build process requires importing that module.

Signed-off-by: Zack Cerza <zack@cerza.org>

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -7,6 +7,12 @@
 # https://docs.docker.com/compose/django/
 # https://docs.docker.com/compose/wordpress/
 
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata
+
+__version__ = importlib_metadata.version("podman-compose")
 
 import sys
 import os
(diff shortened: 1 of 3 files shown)


muayyad-alsadi:
I would love for the podman-compose.py to be a standalone script too.

I don't want to add unnecessary dependencies

podman_compose.py
```diff
@@ -35,8 +41,6 @@
 import yaml
 from dotenv import dotenv_values
 
-__version__ = "1.0.4"
```
p12tic, line 38: I think it was fine if both setup.cfg and podman_compose.py duplicated the version. 

The dependency problems is a valid justification to do so. The downside is that these versions can get out of sync. However this can be addressed by having CI run a simple regex based test that extracts versions from files and compares them.

p12tic (changes requested):
Best to duplicate the version and add a test to ensure that the versions are synchronized.

Issue #513, opened 2022-06-30
Typo podman-composer with an R?


johanneskastl:
https://github.com/containers/podman-compose/blob/4a73ae86bc0b4a84461f07a391c2eab8893cee33/podman_compose.py#L1737

I would guess this is a typo, as the project and all other mentions always use `podman-compose` without the R at the end?

Should also be fixed here:
https://github.com/containers/podman-compose/blob/d38aeaa7130cfd7f93341a09460fba9ac7e4c2a5/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L38

I can open up a PR, if desired.

mokibit:
Closing as this was fixed by https://github.com/containers/podman-compose/pull/514.

PR #514, opened 2022-06-30
 Add pause,unpause commands support

Tayeh:


Tayeh:
the second  commit to fix the typo mentioned in the issue #513 ( not sure if required)

podman_compose.py
```diff
@@ -2341,6 +2341,18 @@ def _published_target(port_string):
...
             return
 
+@cmd_run(podman_compose, 'pause', 'Pause all running containers')
+def compose_pause(compose, args):
+    containers = list(reversed(compose.containers))
+    for container in containers:
+        compose.podman.run(['pause'], container['name'])
+
+@cmd_run(podman_compose, 'unpause', 'Unpause all running containers')
+def compose_pause(compose, args):
```
vansari, line 2351: Sry but is it not a little bit confusing that you named a function `compose_pause` even though the function body unpaused all containers?

Tayeh: fixed

muayyad-alsadi:
can we make it take services just like logs

```python
    parser.add_argument(
        "services", metavar="services", nargs="*", default=None, help="service names"
    )
```

you read it like this

```python
    container_names_by_service = compose.container_names_by_service
    if not args.services and not args.latest:
        args.services = container_names_by_service.keys()
    for service in args.services:
        # ...
```

muayyad-alsadi:
apply the diff here

https://github.com/containers/podman-compose/runs/7317924725?check_suite_focus=true

Tayeh:
> can we make it take services just like logs
> 
> ```python
>     parser.add_argument(
>         "services", metavar="services", nargs="*", default=None, help="service names"
>     )
> ```
> 
> you read it like this
> 
> ```python
>     container_names_by_service = compose.container_names_by_service
>     if not args.services and not args.latest:
>         args.services = container_names_by_service.keys()
>     for service in args.services:
>         # ...
> ```


@muayyad-alsadi  Thanks for this improvement, I added the (services argument) and improve the base code with a better implementation

PR #517, opened 2022-07-04
Add Homebrew installation to README

whirlwin:


muayyad-alsadi:
I can't recommend homebrew because

1. I don't maintain it
2. I prefer using pip

PR #520, opened 2022-07-13
Add multi-platform build support

ebekebe:
We are currently evaluating podman-compose not only as a means for development, but also to build containers in CI. We need to build multi-platform images (linux/arm64 and linux/amd64), which is supported by podman with the `--platform` or `--arch` options. This option is not yet exposed by podman-compose.

I acknowledge that docker-compose does not support this either. However, in the docker world there is [docker buildx bake](https://docs.docker.com/engine/reference/commandline/buildx_bake/) which allows building multi-platform images from a compose file. I am not aware of tooling for podman/buildah, that builds multi-platform images for compose files. I also feel that it is the right place to add support here rather than providing some extra tool.

This PR adds a new option `--platform` which enables the use of compose files as templates for building multi-platform images. In addition, it adds a `--all-platforms` options to `podman-compose push`.

Since `--platform` is more specific and maps better to the [manifest spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md) than `--arch`, we only need `--platform`.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1863,7 +1863,10 @@ def compose_push(compose, args):
             continue
         if services and cnt["_service"] not in services:
             continue
-        compose.podman.run([], "push", [cnt["image"]], sleep=0)
+        if getattr(args, "all_platforms", None):
+            compose.podman.run([], "manifest", ["push", "--all", cnt["image"], "docker://" + cnt["image"]], sleep=0)
+        else:
+            compose.podman.run([], "push", [cnt["image"]], sleep=0)
 
 
 def build_one(compose, args, cnt):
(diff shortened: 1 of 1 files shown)


shaneHowearth:
Hi @ebekebe - I'm in need of `--platform` with `podman-compose` too and found your PR to be quite helpful. The main branch has shifted a bit since your PR but I was able to work out where to make the changes except one that isn't included - when I run I get "AttributeError: 'Namespace' object has no attribute 'platform'". My python is woefully out of date and I cannot figure out where to add that field to the `args` object (the specific line triggering the error is
```
if 1 < len(args.platform):
```

ebekebe:
Hi @shaneHowearth , I am not familiar with the project anymore and can't help. We moved away from podman-compose and are using skaffold now.

PR #521, opened 2022-07-13
Exit codes

ebekebe:
Failure exit codes for `build` and `push` are currently not forwarded as exit code for podman-compose. This makes it impossible to use podman-compose in a scripted environment like CI.

With this PR, podman-compose stops building/pushing when the underlying podman command fails and forwards its exit code.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1863,7 +1863,9 @@ def compose_push(compose, args):
             continue
         if services and cnt["_service"] not in services:
             continue
-        compose.podman.run([], "push", [cnt["image"]], sleep=0)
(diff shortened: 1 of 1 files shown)


mi-volodin:
@ebekebe looks like other commands (like `run`) also suffers from this. Or am I missing something? Why did you choose to fix just these?

emmanuel-apau:
sad to see an unmaintainable repo

p12tic:
Seems like the issue fixed by this PR is still present at least for some commands.

@ebekebe Would you be open to rebase the PR?

p12tic:
This PR is no longer needed after https://github.com/containers/podman-compose/pull/1259 landed.

PR #523, opened 2022-07-13
Added podman kill command

tayeh:
this PR adds the kill command 
you can use it like this to call all services 
`podman-compose kill --all`
`podman-compose kill --all -s SIGKILL`
or you can pass the service name to args 
`podman-compose kill web`
`podman-compose kill redis -s SIGKILL`


muayyad-alsadi:
your branch can't be merged, please pull latest code.

Tayeh:
i will open a new one

PR #525, opened 2022-07-14
Add healthcheck.disable support

inganault:
Support `disable` property in the `healthcheck` section
Ref: https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck
Ref2: https://github.com/docker/compose/blob/1.28.x/compose/config/config.py#L904

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -931,11 +931,15 @@ def container_to_args(compose, cnt, detached=True):
     platform = cnt.get("platform", None)
     if platform is not None:
         podman_args.extend(["--platform", platform])
+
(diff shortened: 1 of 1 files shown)


inganault:
It won't emit `--health-cmd`, but it will produce `--no-healthcheck` from the existing condition in this line
https://github.com/containers/podman-compose/blob/49aa47de3bbf5a7797fdf9c1159c8c729bd66711/podman_compose.py#L954

From the compose file ref
> To disable any default healthcheck set by the image, you can use disable: true. This is equivalent to specifying test: ["NONE"]

PR #538, opened 2022-08-16
Use top-level name property as project name

dotsimon:
The [compose-spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#name-top-level-element) states that the top-level name property should be used if user doesn't set one explicitly.

This PR implements that requirement such that the project name is set from the first valid of:

1. `-p` command line option
2. top-level `name` element
3. basename of the current directory

The following compose file snippet would set the project name to `proj_1234`

```yaml
name: proj_1234
services:
  foo:
    ...
```

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1424,17 +1424,6 @@ class PodmanCompose:
         # TODO: remove next line
         os.chdir(dirname)
 
-        if not project_name:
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
Please apply the style diff here

https://github.com/containers/podman-compose/runs/7873432910?check_suite_focus=true

PR #546, opened 2022-08-27
add support for network_mode: none, and arg change: --no-pod -> --pod

rtgiskard:


muayyad-alsadi:
I want pod to be true, false, auto
where auto is false for podman 3.x and true 4.x
and the default is false in podman-compose 1.0.x and the default is auto for 1.1.x

rtgiskard:
As pod is a feature of podman, traditional docker-compose do not have the concept, for consistent behavior,
 it should be specified explicitly if desired. But if it's a better practice, current behaviour may be better.

Or maybe a default configuration file can be load if exist, I'd like to have some default options in the configurarion
like which or even multiple compose file to load, some of the default podman args and etc.

PR #557, opened 2022-09-19
Output: use logging, remove garbage output by default

### LINKED ISSUE #489: Lower verbosity by default or add option to allow removing prefixed and suffixed debug output
podman-compose prints additional debug information when commands are executed, and there appears to be no way to silence it. Some tools process CLI output as-is and expect it to be in a certain format (such as JSON), but the additional output printed by podman-compose before the actual command output can break these tools.

Perhaps lower the default verbosity to not print debug output unless `-v` or `--verbose` is passed.

Leaving the current behaviour the same, and adding a flag `--silent`, `--verbosity=none` or similar, to achieve the same effect.

**Additional context**
The version I'm using is 1.0.3.

Below is an example of the problem. Suppose I have some tool that wants to execute `echo hi`, and I want to wrap this in a container using Podman, I override the command to be:

```
podman exec --interactive --tty some_container_1 echo hi
```

The output is:

```
hi
```

Great, that works. Now I want to use podman-compose to do the same, and execute the command from an existing container in my pod, so I turn this into:

```
podman-compose exec some_container echo "hi"
```

The output is now the expected output, prefixed and suffixed with debug information:

```
['podman', '--version', '']
using podman version: 4.0.3
podman exec --interactive --tty some_container_1 echo hi
hi
exit code: 0
```

This debug information is useful, but is usually redundant in scenarios such as `podman-compose up` (and **also seems to expose secrets** if you pass them through the environment, as it also logs the environment variables it passes), so the best approach seems to lower the verbosity by default.

jamesob:
Fixes #489.

Introduces a `--verbose` flag if you want to see all the noise that was previously printed by default.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -22,6 +22,7 @@ import hashlib
 import random
 import json
 import glob
+import logging
 
 from threading import Thread
 
@@ -70,10 +71,7 @@ def try_float(i, fallback=None):
     return fallback
 
 
-def log(*msgs, sep=" ", end="\n"):
-    line = (sep.join([str(msg) for msg in msgs])) + end
-    sys.stderr.write(line)
-    sys.stderr.flush()
+log = logging.getLogger(__name__)
 
 
 dir_re = re.compile(r"^[~/\.]")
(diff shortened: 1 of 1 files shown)


errodrigues:
Any updates on this?

alvarolop:
Hello, any updates on this?

maurerle:
@jamesob can please you rebase onto main to resolve the conflict?

Hopefully this helps to get this merged.

jamesob:
I've stopped using podman because the tooling is too bad, but I'll try to get back to this at some point soon.

maurerle:
So you don't mind if i rebase this and file a new PR?

jamesob:
@maurerle by all means!

maurerle:
This is superseded by #741

PR #564, opened 2022-09-28
Removing "-a" from the "start" args array. 

UnKulMunki:
Removing "-a" from the "podman start" args array. It messes with automated container build in scripts.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2058,7 +2058,7 @@ def compose_up(compose, args):
         obj = compose if exit_code_from == cnt["_service"] else None
         thread = Thread(
             target=compose.podman.run,
-            args=[[], "start", ["-a", cnt["name"]]],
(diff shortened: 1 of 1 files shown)


p12tic (approved):
Looks great, thanks!

The PR needs a rebase. Probably it would be easier to start off the `main` branch, delete `-a` and push to the branch of this PR.

bailsman:
Correct me if I'm wrong, but wouldn't removing "-a" change the semantics of `podman-compose up`? Currently, podman-compose attaches to containers by default, which matches the default behavior of docker-compose. https://docs.docker.com/reference/cli/docker/compose/up/#description. I think removing the "-a" causes `podman-compose up` to behave the same as `podman-compose up --detach` and there would no longer be any way to achieve the current behavior.

p12tic:
Good catch, I will thus close this PR.

Issue #573, opened 2022-10-19
Unable to use Secrets with SELinux enabled


henryreed:
**Describe the bug**
Using Fedora CoreOS 36 (ostree version 36.20220918.3.0), I am unable to use `podman` secrets in `podman-compose` because the secret is not mounted with SELinux relabelling.

**To Reproduce**
Example yaml:
```yaml
secrets:
  key:
    file: ./key
services:
  busybox:
    image: busybox:latest
    command: /bin/sh -c "id; ls -lash /run/secrets"
    secrets: 
      - key
```

Where `key` is a file containing some data:
```bash
[core@localhost secrets]$ ls -lh key 
-rw-r--r--. 1 core core 14 Oct 18 23:09 key
[core@localhost secrets]$ cat key 
VERYSECUREKEY
```
Make sure SELinux is set to enforcing in your operating system. Running `podman-compose up` we receive the following error from within the container:
```bash
podman-compose up

...

podman start -a secrets_busybox_1
uid=0(root) gid=0(root) groups=10(wheel)
ls: /run/secrets/key: Permission denied
total 0      
     0 drwxr-xr-x    2 root     root          60 Oct 18 23:13 .
     0 drwxr-xr-t    3 root     root          42 Oct 18 23:13 ..
exit code: 1
```

**Expected behavior**
`podman-compose` should properly relabel the secret to have SELinux grant read/write access to that specific container, similar to how mounts can be relabelled using the capital letter Z, e.g.:
```
volumes:
      - /var/home/core/config:/config:Z
```

**Actual behavior**
The secret is not relabelled and SELinux denies access.

**Output**

```bash
[core@localhost secrets]$ podman-compose version
['podman', '--version', '']
using podman version: 4.2.0
podman-composer version  1.0.3
podman --version 
podman version 4.2.0
exit code: 0
[core@localhost secrets]$ podman-compose up
['podman', '--version', '']
using podman version: 4.2.0
** excluding:  set()
podman stop -t 10 secrets_busybox_1
secrets_busybox_1
exit code: 0             
podman rm secrets_busybox_1                        
aa4ce341fc92a9eb82fd156d6315493cf0e63d55da4cfa2b937741a210dcf824
exit code: 0
['podman', '--version', '']          
using podman version: 4.2.0                
** excluding:  set()              
['podman', 'network', 'exists', 'secrets_default']
[...]

PR #574, opened 2022-10-19
Use SELinux mount flag for secrets

henryreed:
Fixes issue [573](https://github.com/containers/podman-compose/issues/573). 

I tested this in Fedora CoreOS 36, which fixed the issue I had. I also tested it in Ubuntu 22.04, which does not have SELinux. The Ubuntu test showed that secrets can be mounted just fine with the 'Z' SELinux option even on systems that don't have SELinux installed.

For reference, the 'Z' option makes it s.t. SELinux allows a specific container to access a file or folder; other containers will be denied access by SELinux. This is unlike the lowercase 'z' option, which would have SELinux permit a file or folder from being accessed by _any_ container.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -538,7 +538,7 @@ def get_secret_args(compose, cnt, secret):
         source_file = os.path.realpath(
             os.path.join(basedir, os.path.expanduser(source_file))
         )
-        volume_ref = ["--volume", f"{source_file}:{dest_file}:Z,ro,rprivate,rbind"]
+        volume_ref = ["--volume", f"{source_file}:{dest_file}:z,ro,rprivate,rbind"]
         if uid or gid or mode:
             sec = target if target else secret_name
             log(
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -538,7 +538,7 @@ def get_secret_args(compose, cnt, secret):
         source_file = os.path.realpath(
             os.path.join(basedir, os.path.expanduser(source_file))
         )
-        volume_ref = ["--volume", f"{source_file}:{dest_file}:ro,rprivate,rbind"]
+        volume_ref = ["--volume", f"{source_file}:{dest_file}:Z,ro,rprivate,rbind"]
         if uid or gid or mode:
             sec = target if target else secret_name
             log(


benyaminl:
have you test this on non SELINUX Env? I mean it should also check whether there are selinux or not, because a lot of ubuntu/debian user use podman-compose and podman on non selinux env... sadly... 

2nd, have you thought of sharing secret with other container? there are :z with small z, yes small.... 

Have you seen 
https://docs.podman.io/en/latest/markdown/podman-run.1.html#volume-v-source-volume-host-dir-container-dir-options

https://docs.podman.io/en/latest/markdown/podman-run.1.html#volumes-from-container-options


The <b>z</b> option tells Podman that <b>two containers share the volume content</b>. As a result, Podman labels the content with a shared content label. Shared volume labels allow all containers to read/write content. The <b>Z</b> option tells Podman to <b>label the content with a private unshared label</b>. Only the current container can use a private volume.

henryreed:
@benyaminl Sorry for the late reply.

I did test this in Ubuntu 22.04, which does not have SELinux installed by default, as stated in the OP. You are correct that the lowercase `z` would allow multiple containers to read the secret. I wish there was a way to compartmentalize a secret via SELinux s.t. a select few containers could view the secret, and others would not be able to, but as far as I can tell that is not possible.

For usability's sake I think the lowercase `z` flag should be used. I can make the change to this pull request sometime this week.

Akruidenberg:
Any progress?

benyaminl:
@muayyad-alsadi

denMaier:
this seems like an important feature

henryreed:
> Any progress?

Apologies for the "week" taking so long. Pushed the change to a small `z` now, just need to wait on a repo maintainer to approve.

benyaminl:
Ping Mr. @muayyad-alsadi

muayyad-alsadi:
I'm very sorry it took my almost a year, and after all this time, I merged it then reverted it.

Security is a priority over convenience specially with secrets.
Marking secrets as shareable does not feel right at all. I totally agree with @benyaminl 

So if we have a problem, we need to fix the problem, we should not `chmod 777` everything.
to test SELinux we can use any modern fedora VM to verify whatever you are trying to accomplish because they have functional SELinux, ubuntu is known for not having working selinux, they use apparmor

muayyad-alsadi:
> if you run a webapp container and a MariaDB container, you would need to supply the same MariaDB password to each container with two different secrets. 

you convinced me with this example

> isn't similar to chmod 777;

it's, think of it as multi-tenant env.
a container can steal the secret of others
but I did not think about the previous use case. so I'm sold now

----

but I have another approach, if you agree with me let's do another PR

in the code line ~590 

```
    # v3.5 and up added external flag, earlier the spec
    # only required a name to be specified.
    # docker-compose does not support external secrets outside of swarm mode.
    # However accessing these via podman is trivial
    # since these commands are directly translated to
    # podman-create commands, albiet we can only support a 1:1 mapping
    # at the moment
    if declared_secret.get("external", False) or declared_secret.get("name", None):
```

the current code does not create "internal named secrets" it just volume mounts them

```
if source_file:
   --volume
else:
   --secret
```

just like volumes, let's create internal named secrets (as opposed to external secret)
so we read top level secrets and create named secret for them like this `f"{proj}_{name}"`
in a way similar to `assert_volume()` 

```
---
secrets:
  key:
    file: ./key
services:
  busybox:
    image: busybox:latest
    command: /bin/sh -c "id; ls -lash /run/secrets"
    secrets: 
      - key
```

will translate to

```
podman secret create -l 'io.podman.compose.project=vol' myproj_key ./key
podman run --secret myproj_key,target=key,uid=0,gid=0,mode=0
```

the logic of prefixing project_name is if not external just like volumes.

and BTW in ~609 this is no longer the case

```
        if target and target != secret_name:
            raise ValueError(err_str.format(target, secret_name))
        if target:
            log(
                'WARNING: Service "{}" uses target: "{}" for secret: "{}".'.format(
                    cnt["_service"], target, secret_name
                )
                + " That is un-supported and a no-op and is ignored."
            )
```

it's supported as I showed above

first do you agree with my proposal?
if not why?
if yes then would you do it? should I wait a PR from you?

muayyad-alsadi:
also let's fix #671 

I'm waiting your opinion on the subject

henryreed:
@muayyad-alsadi 

My apologies for responding so late, I have been under the weather the last few days.

First, I agree with your proposal. Making secrets per container would negate every problem and I think overall it is a much more elegant solution. Let me try to work it, I will make a new PR--I will let you know if I'm not bright enough to figure it out :)

thatnerdjosh:
@henryreed - is this something you're working on currently?

thatnerdjosh:
> @thatnerdjosh @muayyad-alsadi 
> 
> Hi folks--unfortunately, I took on way too much this year and I won't be able to support this. Sorry for dipping out and the constant back and forth.

No worries - I had just stumbled on this issue when using podman-compose and wanted to see if it's being worked on - if I find some time in the near future I might take a stab at solving it.

It should just be a matter of adding the 'z' flag if I'm understanding correctly, right?

henryreed:
> It should just be a matter of adding the 'z' flag if I'm understanding correctly, right?

It is a little more complicated than that. In this case, you'll need to implement a call to use the native `podman secrets` to generate a secret. Then, when creating the container you will need to supply a secrets argument. See the message [here](https://github.com/containers/podman-compose/pull/574#issuecomment-1504897778) for details.

jmaibaum:
Chiming in here, trying to get https://git.sr.ht/~emersion/sr.ht-container-compose/ to work with podman and podman-compose on Fedora Silverblue, I got so far as to relabel all native volumes with :Z, but I'm stuck getting the `configs:` working, which seem to be closely related to secrets.

Bottom line so far seems to be that I end up with a lot of "No such file or directory" errors for all the configs specified and used by the individual services.

Progress on this issue, supporting secrets and regular configs would be highly appreciated, though I understand that while there is a `podman secrets` command to natively support secrets, it seems that the proposed solution in https://github.com/containers/podman-compose/pull/574#issuecomment-1504897778 does not work for `configs:`.

Would you prefer to open a separate issue for configs?

jmaibaum:
FWIW, I have just added #835 for `configs:` support as it seems to be not yet implemented at all after a quick glance over the code. Sorry for adding noise here...

PR #576, opened 2022-10-20
chore: Run tests folder as pytest cases

jaysonsantos:
Not sure if this is a wanted thing.
I have an issue with devcontainers and before sending the pr with the fix, i'd like to write a test case for it to avoid breaking other stuff.
So, I will send a stacked pr after this one with the actual fix

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1348,9 +1348,9 @@ class PodmanCompose:
             xargs.extend(shlex.split(args))
         return xargs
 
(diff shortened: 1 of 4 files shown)


p12tic:
Thanks for the PR. I'm sorry that it took 2 years to get back to this. Currently the integration tests are already ran automatically (via unittest). I'm thus closing the PR as it is no longer relevant.

PR #613, opened 2023-01-04
Add fake "podman network exists" command

paulmillar:
Motivation:

podman-compose uses the "podman network exists" command to avoid creating the same network twice.  This command was added with podman v3.1.0.

Debian stable has an older version of podman (v3.0.1) that doesn't support the "podman network exists" command.

A symptom of this problem is podman-compose failing with lines like:

    subprocess.CalledProcessError: Command '['podman', 'network', 'exists', 'scicatlive_default']' returned non-zero exit status 125.

    During handling of the above exception, another exception occurred:

    [...]
    subprocess.CalledProcessError: Command '['podman', 'network', 'create', '--labelect=scicatlive', 'scicatlive_default']' returned non-zero exit status 125.

Modification:

Abstract the two places where podman-compose checks if a network already exists.  This is now handled by a specific method.

Check the podman version.  If the podman version is earlier than v3.1.0 then simulate the "podman network exists" command by parsing the output from "podman network ls", otherwise simply call the "podman network exists" command directly.

Result:

podman-compose is now able to create a network with versions of podman before v3.1.0.


maurerle:
I think this is not needed anymore now as all recent systems have a more recent version (bookworm has 4.3.1 for example)

paulmillar:
I'm sure that Debian "bookworm" (v12) includes a newer version of podman, and so doesn't suffer from this problem.  Therefore "simply" upgrading to bookworm should solve the problem this patch addresses.

Debian "bullseye" is now "oldstable", so it is still supported.  Under Debian LTS, the lifetime of bullseye is currently undefined:

https://wiki.debian.org/LTS

However, the above page says "roughly July 2024 to June 2026", which would mean people may (somewhat reasonably) continue to use bullseye for another three years.

podman_compose.py
```diff
@@ -669,6 +670,22 @@ def norm_ports(ports_in):
         ports_out.append(port)
     return ports_out
 
+def assert_network_exists(compose, net_name):
+    podman_version = sv.Version(compose.podman_version)
+    if sv.SimpleSpec('<3.1.0').match(podman_version):
+          output = compose.podman.output([], "network", ["ls"])
```
p12tic, line 676: Must be indented by 4 spaces. Right now it's 6.

podman_compose.py
```diff
@@ -669,6 +670,22 @@ def norm_ports(ports_in):
...
+          offset = None
+          for raw_line in output.splitlines():
```
p12tic, line 678: Output parsing should be moved to separate function and a unit tests added for it.

p12tic:
@paulmillar Are you still using podman-compose on bullseye? I would be open to supporting bullseye version of podman if someone would be willing to setup testing in bullseye and fix other issues found.

Right now I'm aware that podman-compose requires python 3.11 due some usages of asyncio. Probably there are equivalents that can be used on older Python. We need to support 3.9 if bullseye is considered.

After this it's probably almost trivial to add some github actions matrix lines to enable support for bookworm too.

p12tic (changes requested):
Needs tests.

PR #634, opened 2023-02-09
add support for network priorities

benibr:
This PR adds support for the `priority` parameter in the network section of a service as described in https://github.com/compose-spec/compose-spec/blob/master/spec.md#priority
Till now this parameter is silently dropped but it should order the arguments of the `--net` parameter that is passed to `podman run`. 
This patch changes the `net_names` variable to a list so that it can keep the order. It the `networks` section is given as dict the networks are sorted first by priority and per priority alphabetically.

Signed-off-by: Benedikt Braunger <bb@emlix.com>

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -775,17 +775,28 @@ def get_net_args(compose, cnt):
     ip = None
     ip6 = None
     if cnt_nets and is_dict(cnt_nets):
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
I've adopted your solution but did some slight changes
I've move sort outside the loop
used lists instead of dict 
and added a test

PR #635, opened 2023-02-10
Support network_mode: none

### LINKED ISSUE #566: "network_mode: none" is currently not supported
According to https://github.com/compose-spec/compose-spec/blob/master/spec.md#network_mode, it should disable all container networking.

micheljung:
Fixes #566


muayyad-alsadi:
would you please add a `docker-compose.yaml` in tests with this case

PR #656, opened 2023-04-07
Fixes #636: env-file shall be resolved relative to the CWD

### LINKED ISSUE #636: --env-file flag only works with path relative to the container-compose filepath
**Describe the bug**

--env-file flag only works with path relative to the container-compose filepath.

please try to reproduce the bug in latest devel branch

**To Reproduce**
here's my project layout:
```
root/
    - proj1/
        - container-compose.yml
        - envs/
            - 1.env
```
1. if run from the root `podman-compose -f proj1/container-compose.yml --env-file proj1/envs/1.env`, the env vars will not be loaded.

**Expected behavior**
The env vars should be loaded when env-file points to the env file path.

**Actual behavior**
What is the behavior you actually got and that should not happen.

env vars not loaded.

**Environment:**
 - OS: WSL
 - podman version: 4.4.1
 - podman compose version: 1.0.4(built from latest)

**Additional context**

1. if run from the root `podman-compose -f proj1/container-compose.yml --env-file envs/1.env`, the env vars will be loaded correctly.

Please correct me if this is the expected behavior for podman/podman-compose.

bugfest:
- Removed TODO
- Implemented env-file relative to CWD
- Added tests scenarios


bugfest:
Fixes #636

muayyad-alsadi:
I wonder if removing chdir is ok
I guess we should just move it after loading env
keep the comment to consider removing it

I don't know the concequences

bugfest:
Ok @muayyad-alsadi will do. We should probably start creating some tests to cover these scenarios

bugfest:
> I wonder if removing chdir is ok I guess we should just move it after loading env keep the comment to consider removing it
> 
> I don't know the consequences or totally removing it

Done

PR #680, opened 2023-04-21
Add build subcommand  and --build arg to compose_run

Evedel:
The default behavior of `docker compose run` is to build a service if it does not exist, or force build if `--build` flag is provided. `podman-compose` currently does not build missing images before `run` and does not have `--build` flag on `run` at all. Which might be quite confusing for someone migrating between the two. That would be nice if the behavior is consistent.

docker compose:
image

podman-compose 1.0.6:
image

podman-compose with this change:
image


muayyad-alsadi:
Just use black to reformat your changes.

Issue #622, opened 2023-01-19
podman-compose config should split command string into array


loynoir:
**Is your feature request related to a problem? Please describe.**
See differnet between `podman-compose` and `docker-compose` using reproduce case in  https://github.com/containers/podman-compose/issues/620

input `command: sleep infinity`

docker-compose config output
```
    command:
    - sleep
    - infinity
```

podman-compose config output `command: sleep infinity`

**Describe the solution you'd like**
https://github.com/containers/podman-compose/issues/621

**Describe alternatives you've considered**
Implement in python.

**Additional context**

mokibit:
I can no longer reproduce this issue.
Environment:
```
Debian GNU/Linux 13 (trixie)
Python 3.13.5
podman-compose version 1.6.0
podman version 5.4.2
```
`compose.yml` file:
```
services:
  test:
    image: busybox
    command: sleep infinity
```
For comparison with `docker-compose`:
```
$ docker-compose --version
Docker Compose version 2.26.1-4
```
Run the `config` command:
```
$ podman-compose config
services:
  test:
    command:
    - sleep
    - infinity
    image: busybox
```
Compare with `docker-compose`:
```
$ docker-compose config
name: test-podman-compose
services:
  test:
    command:
      - sleep
      - infinity
    image: busybox
    networks:
      default: null
networks:
  default:
    name: test-podman-compose_default
```
Regarding `command` args,  `podman-compose` behaves the same as `docker-compose`.

PR #681, opened 2023-04-21
Allow config to merge strings and lists in "command" and "entrypoint"

Evedel:
There is an inconsistency in which types are allowed to be merged where. While the `docker compose` allows (only) strings and lists to be merged for both `entrypoint` and `command`, `podman compose` only allows that for `command`. Entrypoint is also not merged but extended when it is `list` in both configs.

Merge `str` & `str` without spaces:
Change: none
image

Merge `str` & `str` with spaces:
Change: result is split into a list of strings
image

Merge `list` and `list`:
Change: fixes incorrect extension
image

Merge `str` and `list`:
Change: allows merge
image

Not a `str` or a `list` of `str` in any position:
Change: more verbose output
image


Evedel:
going to resolve #622

muayyad-alsadi:
Should we just replace split with shlex.split?

Evedel:
Very good point! Huge thanks for pointing that out @muayyad-alsadi, I've added a bunch of test cases including one covering a particular scenario where `shlex` will be most critical.

Evedel:
redone this PR in the spirit of #685

Issue #612, opened 2023-01-03
Error "ValueError: can't merge value of build of type <class 'str'> and <class 'dict'>;" caused by build property defined as a string value


peterkracik:
**Describe the bug**
If the `build` property in the `docker-compose.yml` is defined as a string, the `podman compose up` throws `ValueError: can't merge value of build of type <class 'str'> and <class 'dict'>;`
After some digging, I found out, that it has to be defined as a class

to reproduce:

in `docker-compose.yml`

**Actual behavior**
```yaml
my_service:
  build:  ./my_service_context_path
...
```
-> error: ValueError: can't merge value of build of type <class 'str'> and <class 'dict'>;`

```yaml
my_service:
  build:  
    context: ./my_service_context_path
...
```
-> works correctly.

**Expected behavior**
to be able to define build as a string value, which is supported by docker-compose

**Output**

```
$ podman-compose version
['podman', '--version', '']
using podman version: 4.3.1
podman-composer version  1.0.3
podman --version
podman version 4.3.1

$ podman-compose up
['podman', '--version', '']
using podman version: 4.3.1
Traceback (most recent call last):
  File "/opt/homebrew/bin/podman-compose", line 33, in <module>
    sys.exit(load_entry_point('podman-compose==1.0.3', 'console_scripts', 'podman-compose')())
  File "/opt/homebrew/Cellar/podman-compose/1.0.3_1/libexec/lib/python3.10/site-packages/podman_compose.py", line 1775, in main
    podman_compose.run()
  File "/opt/homebrew/Cellar/podman-compose/1.0.3_1/libexec/lib/python3.10/site-packages/podman_compose.py", line 1022, in run
    self._parse_compose_file()
  File "/opt/homebrew/Cellar/podman-compose/1.0.3_1/libexec/lib/python3.10/site-packages/podman_compose.py", line 1091, in _parse_compose_file
    rec_merge(compose, content)
  File "/opt/homebrew/Cellar/podman-compose/1.0.3_1/libexec/lib/python3.10/site-packages/podman_compose.py", line 918, in rec_merge
    ret = rec_merge_one(target, source)
  File "/opt/homebrew/Cellar/podman-compose/1.0.3_1/libexec/lib/python3.10/site-packages/podman_compose.py", line 908, in rec_merge_one
    rec_merge_one(value, value2)
[...]

Issue #620, opened 2023-01-19
podman-compose config should expand relative path


loynoir:
**Describe the bug**
vscode devcontainer build failed when using `podman-compose` instead of `docker-compose`

**Debug**
- `docker-compose` expand relative path './xxx' '../xxx' to abosolute path, while `podman-compose` doesn't. And vscode devcontainer builder cwd is not parent dir of compose file. So, BOOM!
- Above inffect `.services.XXX.build.context`, `.services.XXX.volumes[XXX]`

**To Reproduce**

vscode settings
~/.config/Code/User/settings.json
```
    "dev.containers.dockerPath": "/tmp/podman-devcontainer-wrapper",
    "dev.containers.dockerComposePath": "/tmp/podman-compose-devcontainer-wrapper"
```

```
$ cat /tmp/podman-devcontainer-wrapper                                             
#!/bin/bash
BUILDAH_FORMAT=docker \
PODMAN_USERNS=keep-id \
podman "$@"
$ cat /tmp/podman-compose-devcontainer-wrapper 
#!/bin/bash
BUILDAH_FORMAT=docker \
PODMAN_USERNS=keep-id \
~/.local/bin/podman-compose "$@"
```

---
reproduce
```
$ pip3 install https://github.com/containers/podman-compose/archive/devel.tar.gz 
$ cd /tmp/reproduce
$ cat .devcontainer/docker-compose.yml
version: '3'

services:
  devcontainer:
    build:
      context: .
      dockerfile: Dockerfile
    volumes:
      - ..:/repo:cached
      - ../../dotfiles:/dotfiles:cached
    command: sleep infinity
```

**Expected behavior**
```
$  docker-compose -f ./.devcontainer/docker-compose.yml config
name: devcontainer
services:
  devcontainer:
    build:
      context: /tmp/reproduce/.devcontainer
      dockerfile: Dockerfile
    command:
    - sleep
    - infinity
    networks:
      default: null
    volumes:
    - type: bind
      source: /tmp/reproduce
      target: /repo
      bind:
        create_host_path: true
    - type: bind
      source: /tmp/dotfiles
      target: /dotfiles
      bind:
        create_host_path: true
networks:
  default:
    name: devcontainer_default
```

**Actual behavior**
```
$ BUILDAH_FORMAT=docker PODMAN_USERNS=keep-id ~/.local/bin/podman-compose -f ./.devcontainer/docker-compose.yml config
podman-compose version: 1.0.4
['podman', '--version', '']
[...]

PR #685, opened 2023-04-29
Allow config to merge strings and dicts in "build"

Evedel:
Context:
Trying to make `podman-compose` to work with vscode devcontainers.

vscode will take provided compose files, extend them, create it own extentions, merge them through `podman-compose config` and run. vscode will also create a dockerfile with a new name and place it in the temporary directory. However the merge of `build` section is not currently handled correctly.

The simplified problem is:
compose-1:
```
services:
  dev:
    build: ./dir-1
```

compose-2:
```
services:
  dev:
    build:
      context: ./dir-2
```

expected output
```
services:
  dev:
    build:
      context: ./dir-2
```

actual output
```
    raise ValueError(
ValueError: can't merge value of build of type <class 'str'> and <class 'dict'>
```


Evedel:
Going to resolve #612, and build a base for resolving #620 and #300

muayyad-alsadi:
`normalize` (which calls `normalize_service`) is called just before `rec_merge`

the current code looks like

```
def normalize_service(service, sub_dir=""):
    # make `build.context` relative to sub_dir
    # TODO: should we make volume and secret relative too?
    if sub_dir and "build" in service:
        build = service["build"]
        context = build if is_str(build) else build.get("context", None)
# ..........
            if is_str(build):
                service["build"] = context
            else:
                service["build"]["context"] = context
```

I believe we can just move normalize build as part of this by moving the last part at top
like this

```
def normalize_service(service, sub_dir=""):
    if "build" in service:
        build = service["build"]
        if is_str(build):
            service["build"] = {"context": build}
    # below this line build will always be dict, code should be simpler
    # make `build.context` relative to sub_dir
    # TODO: should we make volume and secret relative too?
    if sub_dir and "build" in service:
        build = service["build"]
        context = build if is_str(build) else build.get("context", None)
# ..........
```

Evedel:
Good call, thanks, that is done.

muayyad-alsadi:
Please resolve the conflict

Evedel:
resolved

muayyad-alsadi:
I still see conflict

Evedel:
Could you please doublecheck? Not sure if that was/is a glitch in GH GUI, cannot see any merge warnings...

muayyad-alsadi:
sorry for being late, it merged fine

> Not sure if that was/is a glitch in GH GUI

indeed that could be the case

PR #686, opened 2023-04-30
Convert build context path to absolute during final normalisation

Evedel:
Context:
Trying to make `podman-compose` work with `vscode` devcontainers.

`vscode` will take provided compose files, extend them, create its own extensions, merge them through `podman-compose config`, run, and attach to them. `vscode` will also create a dockerfiles and composes and place them in the temporary directories.

`docker-compose` is converting `context` to an absolute path, leaving `dockerfile` as a relative path. Meaning that the context directory is always fixed while the place of the compose file may vary **relative to that context directory**. Third-party tools developed for `docker-compose` will also rely on that behavior.

`podman-compose` is currently only normalizing a shape if `build` rather than its content, leaving the `context` directory relative. It results in the context path is **relative to the calling process** rather than to the location of compose file.

Minimal example:
```
services:
  dev:
    build: './'
```

Expected output of `podman-compose config`
```
services:
  dev:
    build:
      context: /absolute/path/to/compose-file
      dockerfile: Dockerfile
```

Actual output:
```
services:
  dev:
    build: ./
```

Please see [test cases](https://github.com/containers/podman-compose/compare/devel...Evedel:podman-compose:convert-build-context-path-to-absolute-during-normalisation?expand=1#diff-64fefc67a9001dc9beb4b243e4804de453235169ec614ed000b00faaf5619d45R17) for more examples of expected normalisations.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1601,7 +1601,7 @@ class PodmanCompose:
             compose.get("services", {}), set(args.profile)
         )
         compose["services"] = resolved_services
-        if not args.no_normalize:
+        if not getattr(args, "no_normalize", None):
             compose = normalize_final(compose, self.dirname)
         self.merged_yaml = yaml.safe_dump(compose)
         merged_json_b = json.dumps(compose, separators=(",", ":")).encode("utf-8")
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -24,7 +24,6 @@ import json
 import glob
 
 from threading import Thread
-from pathlib import Path
 
 import shlex
 
@@ -1292,10 +1291,12 @@ def normalize_service_final(service: dict, project_dir: str) -> dict:
     if "build" in service:
         build = service["build"]
         context = build if is_str(build) else build.get("context", ".")
-        context = str((Path(project_dir) / context).resolve())
-        dockerfile = "Dockerfile"
-        if "dockerfile" in service["build"]:
-            dockerfile = service["build"]["dockerfile"]
+        context = os.path.normpath(os.path.join(project_dir, context))
+        dockerfile = (
+            "Dockerfile"
+            if is_str(build)
+            else service["build"].get("dockerfile", "Dockerfile")
+        )
         if not is_dict(service["build"]):
             service["build"] = {}
         service["build"]["dockerfile"] = dockerfile
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1601,7 +1601,8 @@ class PodmanCompose:
             compose.get("services", {}), set(args.profile)
         )
         compose["services"] = resolved_services
-        compose = normalize_final(compose, self.dirname)
+        if not args.no_normalize:
+            compose = normalize_final(compose, self.dirname)
         self.merged_yaml = yaml.safe_dump(compose)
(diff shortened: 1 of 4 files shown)


Evedel:
```
=========================== short test summary info ============================
FAILED pytests/test_normalize_final_build.py::test__parse_compose_file_when_multiple_composes
ValueError: can't merge value of build of type <class 'str'> and <class 'dict'>
```
tests are currently failing, which is expected and will be resolved by #685

Evedel:
Also, this is the last one in a "series", with all those three merged (can check out [poc-vscode branch](https://github.com/Evedel/podman-compose/tree/poc-vscode)), it will be possible to run devcontainers natively with `podman-compose` 🎉 
image
image

muayyad-alsadi:
please don't use `pathlib` just like the rest of the code use `os.path.join()` as it will absolute path but if relative it would 

```
>>> os.path.normpath(os.path.join('/path/to/proj', './my_context'))
'/path/to/proj/my_context'
>>> os.path.normpath(os.path.join('/path/to/proj', '/abs/my_context'))
'/abs/my_context'
```

replace code like this

```
dockerfile = "Dockerfile"
if "dockerfile" in service["build"]:
            dockerfile = service["build"]["dockerfile"]
```

with

```
dockerfile = service["build"].get("dockerfile", "Dockerfile")
```


please rename `normalize_final` and `normalize_service_final`
into `normalize_compose` (or `normalize_services_all`) and `normalize_service` (or `normalize_service_one`)

please fix the unit test 

https://github.com/containers/podman-compose/actions/runs/4842977047/jobs/8630264268?pr=686

```
FAILED pytests/test_normalize_final_build.py::test__parse_compose_file_when_multiple_composes - ValueError: can't merge value of build of type <class 'str'> and <class 'dict'>
========================= 1 failed, 7 passed in 0.40s ==========================
Error: Process completed with exit code 1.
```

Evedel:
> please don't use `pathlib`

done

> replace code like this

done

> please rename `normalize_final` and `normalize_service_final` into `normalize_compose` (or `normalize_services_all`) and `normalize_service` (or `normalize_service_one`)

There are already `normalize` and `normalize_service` functions (they run pre-merge and per individual compose file), so that might be confusing if there will be a second set of `normalize` functions named similarly without indication in their names pointing to their functional difference, hence the "final" suffix (maybe `normalise_pre_merge`, `normalise_service_pre_merge` and `normalise_post_merge`, `normalise_service_post_merge` instead?).
Let me know which way you want to proceed with.

> please fix the unit test

Those unit tests are broken because of an unrelated issue, please have a look at #685, which is going to fix that. I'll rebase this pr once that one is merged.

Evedel:
fyi @muayyad-alsadi , I rebased pr and added `--no-normalize` flag (consistent with docker config) to fix tests, let me know what your thoughts re: `normalize` function names.

podman_compose.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/python3
+#!/usr/bin/env python3
```
Evedel, line 1: Can go back to the hardcoded interpreter path, but I think it more solid to discover interpreter with `env` 
```suggestion
#! /usr/bin/python3
```

Issue #679, opened 2023-04-21
--abort-on-container-exit does not work


DusanJanik:
**Describe the bug**
podman-compose up --abort-on-container-exit is not aborting other containers when one exits. This creates issues when running this command in a Jenkins job and it results in a failure of the job.

**To Reproduce**
Run this simple docker-compose.yml file with the --abort-on-container-exit option.
```version: "3"
services:
    too_long:
      image: busybox
      command: ["/bin/busybox", "sh", "-c", "sleep 3600; exit 0"]
      tmpfs:
        - /run
        - /tmp
    sh1:
      image: busybox
      command: ["/bin/busybox", "sh", "-c", "sleep 5; exit 1"]
      tmpfs:
        - /run
        - /tmp
    sh2:
      image: busybox
      command: ["/bin/busybox", "sh", "-c", "sleep 10; exit 2"]
      tmpfs:
        - /run
        - /tmp
```

**Expected behavior**
The sh2 and too_long containers should be aborted when the sh1 returns exit code 1.

**Actual behavior**
The sh2 and too_long containers keep running in the background after the sh1 exits. 

**Environment:**
 - OS: Linux
 - podman version: 4.2.0
 - podman compose version: 1.0.6

DusanJanik:
The actual behavior is the same on the devel version 1.0.7

hcd3:
I've run into a similar issue when migrating from Docker to Podman. Podman is unable to run my docker-compose.yml file with the --abort-on-container-exit option, which leads to a failed job in the Gitlab pipeline.

PR #687, opened 2023-05-02
Stop other containers when --abort-on-container-exit is used

jaikiran:
`podman-compose up`, like `docker-compose` provides a `--abort-on-container-exit` option to stop the containers that were launched by the `up` command when any of the other containers exit. In its current form in the `devel` branch, this isn't functional and the other containers keep running. This is also reported as an issue here https://github.com/containers/podman-compose/issues/679.

The commit in this PR adds this missing behaviour. I've tested this against one of the projects that I am using `podman-compose` in. Plus, I've also verified the trivial example that the reporter in https://github.com/containers/podman-compose/issues/679 provided and it now stops the rest of the containers.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1887,6 +1887,24 @@ class cmd_parse:  # pylint: disable=invalid-name,too-few-public-methods
         return wrapped
 
 
+# Thread used to start containers in a compose project. This thread keeps track of the corresponding container name
+# and the service name
+class _ContainerStartThread(Thread):
+    def __init__(self, service_name, container_name, target=None, name=None,
+                 args=(), kwargs=None, daemon=None):
+        super().__init__(target=target, name=name, args=args, kwargs=kwargs, daemon=daemon)
+        self._service_name = service_name
+        self._container_name = container_name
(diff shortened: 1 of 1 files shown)


jaikiran:
Any thoughts about this change?

muayyad-alsadi:
I'm considering moving to python3's builtin asyncio and drop all threads if possible.

I need some time to look into this change.

jaikiran:
Thank you @muayyad-alsadi. Please take your time. Looking forward to your changes and will test them out with our project when they are ready.

hcd3:
Hi @jaikiran and @muayyad-alsadi. Thank you for working on this feature!

Is there any update on this issue? I've run into the error which pertains to this MR and Podman support for the `--abort-on-container-exit` option would be greatly beneficial.

Falmarri:
@muayyad-alsadi I have an PR for moving to asyncio that fixes this issue as well https://github.com/containers/podman-compose/pull/810

jaikiran:
Closing, now that the asyncio approach has been merged.

Issue #695, opened 2023-05-17
podman-compose@.service Should be modified because of pod creation disabled by default


yanghl12138:
**Describe the bug**
Because pod are not created by default[#673](https://github.com/containers/podman-compose/issues/673), podman-compose@.service Should be modified

This is the file generated by `podman-compose systemd -a create-unit`
```
[Unit]
Description=%i rootless pod (podman-compose)

[Service]
Type=simple
EnvironmentFile=%h/.config/containers/compose/projects/%i.env
ExecStartPre=-/usr/bin/podman-compose up --no-start
ExecStartPre=/usr/bin/podman pod start pod_%i
ExecStart=/usr/bin/podman-compose wait
ExecStop=/usr/bin/podman pod stop pod_%i

[Install]
WantedBy=default.target
```

The line 
```
ExecStartPre=-/usr/bin/podman-compose up --no-start
```
should be
```
ExecStartPre=-/usr/bin/podman-compose --in-pod=1 up --no-start
```

**To Reproduce**
```yaml
version: "3.1"

services:
  test:
    image: busybox
    command: sh -c "while [[ 1 ]]; do uname -msr; sleep 5; done"
```
```bash
sudo podman-compose systemd -a create-unit
systemctl --user daemon-reload
podman-compose systemd -a register
podman-compose down
systemctl --user start 'podman-compose@busybox'
podman pod stats 'pod_busybox'
```

**Expected behavior**
Create pod_busybox and start the service using systemd command.

**Actual behavior**
No pod created.

**Output**

```bash
> systemctl --user status 'podman-compose@busybox'
× podman-compose@busybox.service - busybox rootless pod (podman-compose)
     Loaded: loaded (/etc/xdg/systemd/user/podman-compose@.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/user/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Wed 2023-05-17 11:18:51 CST; 4min 15s ago
    Process: 120985 ExecStartPre=/usr/bin/podman-compose up --no-start (code=exited, status=0/SUCCESS)
    Process: 121014 ExecStartPre=/usr/bin/podman pod start pod_busybox (code=exited, status=125)
        CPU: 801ms

May 17 11:18:49 server.fedora podman-compose[120985]: ['podman', 'network', 'exists', 'busybox_default']
[...]

PR #712, opened 2023-06-16
podman_compose: fix in-pod argparse type

StaticRocket:
If "bool" is provided as an argparse type it will only evaluate to false if an empty string is provided. This lambda attempts to resolve this by filtering for "1" or "true".

I used a lambda here for backwards compatibility as argparse didn't add the `BooleanOptionalAction` action until 3.9 and you still advertise 3.7 compatibility in the README and setup.

I was looking at a previous attempt to resolve this ( #546 ) and noticed it was NAK'd for not allowing auto selection but considering the systemd template that's currently present, auto selection is not possible and we seem to be making the assumption that the default value is true.

Maybe the arg should be dropped entirely? The stable release included that template with in-pod defaulting to False leading to fun behavior described in #695 . This is uncorrectable from the user side as the script calls itself with minimal args to create the container before registering it.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1927,7 +1927,7 @@ class PodmanCompose:
             "--in-pod",
             help="pod creation",
             metavar="in_pod",
-            type=bool,
+            type=lambda x: x.lower() in ["1", "true"],
             default=True,
         )
         parser.add_argument(


p12tic:
Thanks for the PR. I rewrote this idea to be more complete and actually check that the passed value can be converted to bool (empty, `1`, `true`, `0`, `false`). Otherwise nonsense values would still be accepted.

The improved PR is this one https://github.com/containers/podman-compose/pull/918. Thanks a lot for the suggested fix.

PR #729, opened 2023-07-16
Use defined environment variables in the image build process

hedayat:
Build images with service environment variables defined so that they can be used in the build process

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2083,6 +2083,10 @@ def build_one(compose, args, cnt):
         build_args.append("--pull-always")
     elif getattr(args, "pull", None):
         build_args.append("--pull")
+    env = dict(cnt.get("environment", {}))
+    for name, value in env.items():
+        build_args += ["--env", f"{name}" if value is None else f"{name}={value}"]
+
     args_list = norm_as_list(build_desc.get("args", {}))
     for build_arg in args_list + args.build_arg:
         build_args.extend(


emersion:
I don't believe this is correct. The service environment is not supposed to be set during build, I believe (at least Docker doesn't do that). This breaks a build for me where `JAVA_TOOL_OPTIONS` is set to `-javaagent:/app/opentelemetry-javaagent.jar` in the compose file (that .jar doesn't exist yet at the build stage).

hedayat:
> I don't believe this is correct. The service environment is not supposed to be set during build, I believe (at least Docker doesn't do that). This breaks a build for me where `JAVA_TOOL_OPTIONS` is set to `-javaagent:/app/opentelemetry-javaagent.jar` in the compose file (that .jar doesn't exist yet at the build stage).

I don't have access to a sample right now, but all my PRs were for compatibility with `docker compose` or `docker-compose`. Unless some behavior is changed there in updates (or in move from older `docker-compose` to newer `docker compose`), this change should be docker compose compatible (I'm not sure about docker itself. There ARE already differences in behavior between docker itself and `docker compose`, like how each one handles multi-line environment files).

emersion:
Weird. I'm working on a project where everyone but me uses a pure Docker setup, and I'm the only one hitting this issue. Commenting the env line when running `podman-compose build` and then uncommenting it when running `podman-compose up` works around the bug.

For reference, the project in question:

https://github.com/OpenRailAssociation/osrd/blob/ade99a6780db963ac26502a9de9d399633e7e26c/docker-compose.yml#L50

hedayat:
Hmm... interesting. I'll try to find the exact use case I had and see how it behaves under latest `docker compose`. Maybe there is a catch somewhere.

emersion:
Do you want me to open an issue to track this?

hedayat:
> Do you want me to open an issue to track this?

Sure, it makes sense! This PR is merged long ago anyway.

emersion:
https://github.com/containers/podman-compose/issues/985

shuckc:
This PR is completely incorrect and should be reverted, as suggested above. Compose runtime environments are absolutely not supposed to be available during the build stage, and in our case result in a failing build.

PR #737, opened 2023-07-29
Support podman's external rootfs management

ftyghome:
The podman-run command supports external rootfs management by using the --rootfs flag ([documentation](https://docs.podman.io/en/latest/markdown/podman-run.1.html#rootfs)).
By specifying the --rootfs flag, the first parameter of podman-run will be the path of container's rootfs, instead of the specification of image. This patch makes the compose file parser able to recognize the "rootfs" field, and produce the correct args for podman-run command.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1145,13 +1145,23 @@ async def container_to_args(compose, cnt, detached=True):
 
     # handle podman extension
     x_podman = cnt.get("x-podman", None)
+    rootfs_mode = False
     if x_podman is not None:
         for uidmap in x_podman.get("uidmaps", []):
             podman_args.extend(["--uidmap", uidmap])
         for gidmap in x_podman.get("gidmaps", []):
             podman_args.extend(["--gidmap", gidmap])
+        rootfs = x_podman.get("rootfs", None)
+        if rootfs is not None:
+            rootfs_mode = True
+            podman_args.extend(["--rootfs", rootfs])
+            log.warning(
+                "WARNING: x-podman.rootfs and image both specified, \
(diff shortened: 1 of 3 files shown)


podman_compose.py
```diff
@@ -1084,7 +1084,11 @@ def container_to_args(compose, cnt, detached=True):
         for gidmap in x_podman.get("gidmaps", []):
             podman_args.extend(["--gidmap", gidmap])
 
-    podman_args.append(cnt["image"])  # command, ..etc.
+    rootfs = cnt.get("rootfs", None)
```
p12tic, line 1087: Is this part of the compose spec? If not, podman-specific keys should be prefixed (e.g. podman.io-rootfs). Also we should have a documentation for such additions somewhere.

p12tic (changes requested):
Thanks for the PR. It looks good in principle. 

Unit tests are needed. `pytests/test_container_to_args.py` already tests some of the ways to call `container_to_args`.

ftyghome:
Hi! 
I have moved podman-specific keys under the "x-podman" field, as the "uidmap" and "gidmap" did.
Unit tests are added, and documents about the extensions are mentioned in the docs/Extensions.md.
Thanks for reviewing:)

p12tic (approved):
Looks great!

I've now realized that there are two ways to supply podman-specific config keys. This will be addressed in the future in a separate PR.

p12tic:
Rebased.

p12tic:
@ftyghome FYI I updated location of the rootfs config in https://github.com/containers/podman-compose/pull/917.

PR #741, opened 2023-08-07
Output: use logging, remove verbose output by default

### LINKED ISSUE #489: Lower verbosity by default or add option to allow removing prefixed and suffixed debug output

maurerle:
Fixes #489.

Introduces a --verbose flag if you want to see all the noise that was previously printed by default.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -379,7 +379,7 @@ async def assert_volume(compose, mount_dict):
     proj_name = compose.project_name
     vol_name = vol["name"]
     is_ext = vol.get("external", None)
-    log.debug(f"podman volume inspect {vol_name} || podman volume create {vol_name}")
+    log.debug("podman volume inspect %s || podman volume create %s", vol_name, vol_name)
     # TODO: might move to using "volume list"
     # podman volume list --format '{{.Name}}\t{{.MountPoint}}' \
     #     -f 'label=io.podman.compose.project=HERE'
@@ -555,9 +555,11 @@ def get_secret_args(compose, cnt, secret):
         volume_ref = ["--volume", f"{source_file}:{dest_file}:ro,rprivate,rbind"]
         if uid or gid or mode:
             sec = target if target else secret_name
-            log.warn(
-                f'WARNING: Service {cnt["_service"]} uses secret "{sec}" with uid, gid, or mode.'
-                + " These fields are not supported by this implementation of the Compose file"
+            log.warning(
+                "WARNING: Service %s uses secret %s with uid, gid, or mode."
+                + " These fields are not supported by this implementation of the Compose file",
+                cnt["_service"],
+                sec,
             )
         return volume_ref
     # v3.5 and up added external flag, earlier the spec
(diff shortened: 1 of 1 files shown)


maurerle:
I made the linter happy, so this can be merged

maurerle:
@muayyad-alsadi can I do anything to help getting this merged?

crdoconnor:
This would be extremely useful.

errodrigues:
+1
Any plans to include this in the next tag/release? Would be a game changer for QoL!

nassir90:
I too hope this can be merged soon. It's a pretty minor code change for a massive QoL improvement.

maurerle:
Hi @ParetoOptimalDev,
i think there is also the possibility to use docker-compose with podman - instead of the unmaintained podman-compose version.
See here:
https://docs.podman.io/en/latest/markdown/podman-compose.1.html

Other than that you can try my fork of podman-compose which has merged a lot of the missing features which are available in open pull requests here:
https://github.com/maurerle/podman-compose

thiagolsfortunato:
any news? 👀

crdoconnor:
I think this project is abandoned.

podman_compose.py
```diff
@@ -394,7 +392,7 @@ def assert_volume(compose, mount_dict):
     proj_name = compose.project_name
     vol_name = vol["name"]
     is_ext = vol.get("external", None)
-    log(f"podman volume inspect {vol_name} || podman volume create {vol_name}")
+    log.debug("podman volume inspect %s || podman volume create %s", vol_name, vol_name)
```
p12tic, line 395: Is there a way to keep f-string formatting here?

p12tic (approved):
Looks great, thank you!

The PR will need a rebase to latest main branch.

p12tic:
cc @maurerle or anyone else who's interested. This PR could easily land if someone rebased it.

podman_compose.py
maurerle: Thank you for your review!
There is no better approach to this as I know.
It is best practice to lazy eval in logging (which is currently only supported with % placeholders) but it is efficient to use f-strings everywhere else.

maurerle:
Thanks @p12tic ! hopefully this will get merged finally :)

p12tic:
We may need to re-add part of logging that prepended task name to log output. However this is useful only to podman-compose developers. There are many more users who will find their output much cleaner now. So it makes sense to postpone further improvements.

Issue #127, opened 2020-03-17
git URLs in context in docker-compose.yml


sanjayankur31:
Hello,

Not sure if this is meant to work yet, or not, but one can apparently include git urls in `context:` in docker compose files:

```
Either a path to a directory containing a Dockerfile, or a url to a git repository.
```

https://docs.docker.com/compose/compose-file/compose-file-v2/#build

This isn't working for me. I get:

```
Traceback (most recent call last):                                                                                                                                                            
  File "/usr/bin/podman-compose", line 11, in <module>                                                                                                                                        
    load_entry_point('podman-compose==0.1.6.dev0', 'console_scripts', 'podman-compose')()                                                                                                     
  File "/usr/lib/python3.8/site-packages/podman_compose.py", line 1267, in main                                                                                                               
    podman_compose.run()                                                                                                                                                                      
  File "/usr/lib/python3.8/site-packages/podman_compose.py", line 755, in run                                                                                                                 
    cmd(self, args)                                                                                                                                                                           
  File "/usr/lib/python3.8/site-packages/podman_compose.py", line 939, in wrapped                                                                                                             
    return func(*args, **kw)                                                                                                                                                                  
  File "/usr/lib/python3.8/site-packages/podman_compose.py", line 1012, in compose_build                                                                                                      
    build_one(compose, args, cnt)                                                                                                                                                             
  File "/usr/lib/python3.8/site-packages/podman_compose.py", line 996, in build_one                                                                                                           
    raise OSError("Dockerfile not found in "+ctx)                                                                                                                                             
OSError: Dockerfile not found in git://github.com/SilverLabUCL/docker-redmine-osb.git             
```

Even if I use `https://`, it still gives the same error.

muayyad-alsadi:
my podman version says

```
       The build context directory can be specified as the http(s) URL of an archive, git repository or Containerfile.
```

would you please try using `podman build ...` and tell me if the problem is in `podman` or `podman-compose`

sanjayankur31:
Sure, I'll have a look and report back. Cheers.

prk3:
I am getting a similar error message.

```
Error: error getting image "sundae_server": unable to find a name and tag match for sundae_server in repotags: no such image
Traceback (most recent call last):
  File "/usr/bin/podman-compose", line 11, in <module>
    load_entry_point('podman-compose==0.1.6.dev0', 'console_scripts', 'podman-compose')()
[...]

PR #745, opened 2023-08-11
Fixed git URL access in build context (#127)

rad10:
I noticed that podman-compose does not work for using URLs and repos, so I made a fix for it


theyoyojo:
👍 I'm using this commit in my fork

p12tic (changes requested):
Looks good in principle, but needs rebase on top of latest main. Also the PR lacks unit and integration tests.

p12tic:
Merged via https://github.com/containers/podman-compose/pull/1171.

PR #749, opened 2023-08-23
Add support for enable_ipv6

### LINKED ISSUE #751: Support enable_ipv6 for networks
I cannot automatically create networks with dual-stack support with `podman-compose up` so I have to edit the network json manually. This is also in the compose spec [1] and supported by `docker compose`
When `enable_ipv6: true` is given for a network, the `--ipv6` flag is passed to `podman network create`
This can currently be done by creating the network manually or editing the network's json file.
**Additional context**
I've opened #749 which implements this.

[1] https://github.com/compose-spec/compose-spec/blob/master/06-networks.md#enable_ipv6

maxi0604:
Fixes #751

This pull request adds support for the `enable_ipv6` network option when creating the network. This is supported by `docker compose` and is contained in the compose-spec [1]. The change simply exposes the `--ipv6` option of `podman network create`

PS: One could argue that this should be on by default and only disabled when `enable_ipv6: false` is set, but doing something like that should - if we even choose to do it at all -  be delayed at the very least until https://github.com/containers/podman/issues/15850 is fixed.

[1] https://github.com/compose-spec/compose-spec/blob/master/06-networks.md#enable_ipv6

### DIFF
--- a/pytests/test_network_create_args.py
+++ b/pytests/test_network_create_args.py
@@ -1,0 +1,143 @@
+import unittest
+
+from podman_compose import get_network_create_args
+
+
+class TestGetNetworkCreateArgs(unittest.TestCase):
+    def test_minimal(self):
+        net_desc = {
+            "labels": [],
+            "internal": False,
+            "driver": None,
+            "driver_opts": {},
+            "ipam": {"config": []},
+            "enable_ipv6": False,
+        }
+        proj_name = "test_project"
+        net_name = "test_network"
+        expected_args = [
+            "create",
+            "--label",
+            f"io.podman.compose.project={proj_name}",
+            "--label",
+            f"com.docker.compose.project={proj_name}",
+            net_name,
+        ]
+        args = get_network_create_args(net_desc, proj_name, net_name)
+        self.assertEqual(args, expected_args)
+
+    def test_bridge(self):
+        net_desc = {
+            "labels": [],
+            "internal": False,
+            "driver": "bridge",
+            "driver_opts": {"opt1": "value1", "opt2": "value2"},
+            "ipam": {"config": []},
+            "enable_ipv6": False,
+        }
+        proj_name = "test_project"
+        net_name = "test_network"
+        expected_args = [
+            "create",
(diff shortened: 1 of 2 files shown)


sgryphon:
This is important. Please merge.

pini-gh:
Anything preventing this to be merged?

p12tic (changes requested):
The PR looks great, thank you!

We need unit tests for it.

This part https://github.com/containers/podman-compose/blob/main/podman_compose.py#L713C1-L749C34 could be extracted to a separate function. Then a unit tests could be written just like for container_to_args in pytests/test_container_to_args.py

p12tic:
@maxi0604 Just a note that podman-compose has switched to `main` as the primary branch. You probably rebased against `devel`.

p12tic:
Also probably a nitpick, but it would be great if extraction of `get_network_create_args` would be the first commit? This way the amount of changes to code not under unit tests would be minimized.

pytests/test_network_create_args.py
```diff
@@ -0,0 +1,57 @@
...
+            "--label", f"io.podman.compose.project={proj_name}",
+            "--label", f"com.docker.compose.project={proj_name}",
+            "--label", "label1",
+            "--label", "label2",
+            "--internal",
+            "--driver", "bridge",
+            "--opt", "opt1=value1",
+            "--opt", "opt2=value2",
+            "--ipam-driver", "default",
+            "--ipv6",
```
p12tic, line 50: Would be great to have a separate test just for ipv6. It's hard to understand unit test when it's large.

In case you have time it makes sense to split other options into separate tests too, but I'm happy that you added any kind of tests for untested code in the first place :) For the PR only test for `--ipv6` is necessary in a strict sense.

p12tic:
Rebased and reordered commits.

pytests/test_network_create_args.py
p12tic: Thanks.

p12tic (approved):
Looks great!

PR #755, opened 2023-08-28
Add codespell config (to ignore loved "poped") and github workflow to prevent future typos

yarikoptic:
There were prior runs of codespell committed (e.g. 78f8cad7c457b200c31a05471fc2b89b1a7be84c ).  But typos would keep appearing unless caught at their inception, which is what this PR proposes to accomplish.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,20 @@
+---
+name: Codespell
+
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,4 @@
+[codespell]
+skip = .git,*.pdf,*.svg
+#
+# ignore-words-list =
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,4 +1,4 @@
 [codespell]
 skip = .git,*.pdf,*.svg
-#
-# ignore-words-list =
+# poped - loved variable name
+ignore-words-list = poped
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,3 +30,7 @@ repos:
             "-sn", # Don't display the score
             "--rcfile=.pylintrc", # Link to your config file
           ]
+  - repo: https://github.com/codespell-project/codespell
+    rev: v2.2.5
+    hooks:
+    - id: codespell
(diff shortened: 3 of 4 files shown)


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [devel]
```
p12tic, line 6: Let's run on all branches.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    branches: [devel]
```
p12tic, line 8: Let's run on all branches.

p12tic (changes requested):
Thanks for the PR. It looks good except in two places and even there the PR was perfect at the time it was created.

p12tic:
Seems like `[[DATALAD RUNCMD] run codespell throughout fixing 1 new typo](https://github.com/containers/podman-compose/pull/755/commits/89f33a04a9decc207a7a66edbddd7f57cd047aa8)` could have a cleaner commit message. Most people who will dig in git history don't care what exact tool was used when the change is a spelling fix :)

.github/workflows/codespell.yml
yarikoptic:
```suggestion
```

.github/workflows/codespell.yml
yarikoptic:
```suggestion
```

yarikoptic:
Ok, I can beg to differ that it is often great to discover what command actually did the change, but that's ok - I will later rewrite that commit and ping here

p12tic:
> great to discover what command actually did the change

I agree, but perhaps a better place would be in CONTRIBUTING.md and/or in error message that would be shown on Github actions logs in case spell check failed.

.github/workflows/codespell.yml
yarikoptic: done locally

p12tic: I can't commit this because I would still need to do a rebase to include the change into the commit that introduced `.github/workflows/codespell.yml`

yarikoptic:
> > great to discover what command actually did the change
> 
> I agree, but perhaps a better place would be in CONTRIBUTING.md and/or in error message that would be shown on Github actions logs in case spell check failed.

Sure thing it is good to have clear instructions/documentation. But it is IMHO quite orthogonal to commit messages being descriptive, especially in case of use of some one-off tool, but also if it is some well known tool -- I guess I just like provenance ;-) Of cause in the case here of a single typo  -- doesn't really matter much.  But anyways -- I simplified commit message. force-pushed

.github/workflows/codespell.yml
yarikoptic: no need - already done. I just thought first to do while on the phone, hence did suggestions, but then realized that need to get to laptop anyways.

p12tic:
Thanks a lot!

PR #763, opened 2023-09-08
Support additional_contexts

### LINKED ISSUE #762: Support `additional_contexts`
I have a project that is split to multiple sub-projects that share a common set of data.
Each sub-project includes a `docker-compose.yml` for setting things up.
To include the shared data, I would like to do this in `docker-compose.yml`:

```
services:
  myservice:
    build:
      context: .
      additional_contexts:
        data: ../data
```

And then in `Dockerfile` for `myservice`:

```
COPY --from=data somedata.json /data/somedata.json
```

But, when I run `podman-compose`, I see this:

```
[4/4] STEP 4/8: COPY --from=data somedata.json /data/somedata.json
? Please select an image: 
  ▸ registry.fedoraproject.org/data:latest
    registry.access.redhat.com/data:latest
    docker.io/library/data:latest
    quay.io/data:latest
```

Implement [additional_contexts] from the Compose Specification.

[additional_contexts]: https://github.com/compose-spec/compose-spec/blob/master/build.md#additional_contexts

1. Use Docker Compose instead, which [implements that feature](https://github.com/docker/compose/pull/10369).
1. Restructure my project so that each subproject contains a copy of the data directory,
or copy the data to sub-projects folder in a preparatory build step.
This would work, but requires restructuring the project even though `additional_contexts` is part of the Compose Specification.

**Additional context**
Yes, please! 😁

otto-liljalaakso-nt:
Fixes #762

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1464,6 +1464,12 @@ def normalize_service(service, sub_dir=""):
             if not context:
                 context = "."
             service["build"]["context"] = context
+    if "build" in service and "additional_contexts" in service["build"]:
+        if is_dict(build["additional_contexts"]):
+            new_additional_contexts = []
+            for k, v in build["additional_contexts"].items():
+                new_additional_contexts.append(f"{k}={v}")
+            build["additional_contexts"] = new_additional_contexts
     for key in ("command", "entrypoint"):
         if key in service:
             if is_str(service[key]):
@@ -2330,6 +2336,8 @@ async def build_one(compose, args, cnt):
         build_args.extend(get_secret_args(compose, cnt, secret, podman_is_building=True))
     for tag in build_desc.get("tags", []):
         build_args.extend(["-t", tag])
+    for additional_ctx in build_desc.get("additional_contexts", {}):
+        build_args.extend([f"--build-context={additional_ctx}"])
     if "target" in build_desc:
         build_args.extend(["--target", build_desc["target"]])
     container_to_ulimit_build_args(cnt, build_args)
--- a/pytests/test_normalize_service.py
+++ b/pytests/test_normalize_service.py
@@ -20,6 +20,14 @@ class TestNormalizeService(unittest.TestCase):
             {"build": {"context": "./dir-1", "dockerfile": "dockerfile-1"}},
             {"build": {"context": "./dir-1", "dockerfile": "dockerfile-1"}},
         ),
+        (
+            {"build": {"additional_contexts": ["ctx=../ctx", "ctx2=../ctx2"]}},
+            {"build": {"additional_contexts": ["ctx=../ctx", "ctx2=../ctx2"]}},
+        ),
+        (
+            {"build": {"additional_contexts": {"ctx": "../ctx", "ctx2": "../ctx2"}}},
+            {"build": {"additional_contexts": ["ctx=../ctx", "ctx2=../ctx2"]}},
+        ),
     ])
     def test_simple(self, input, expected):
         self.assertEqual(normalize_service(input), expected)
--- a/tests/additional_contexts/README.md
+++ b/tests/additional_contexts/README.md
@@ -1,0 +1,14 @@
+# Test podman-compose with build.additional_contexts
+
+```
+podman-compose build
+podman-compose up
+podman-compose down
+```
+
+expected output would be
+
+```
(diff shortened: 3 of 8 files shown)


theyoyojo:
👍 I'm using this commit in my fork

p12tic (changes requested):
Thanks for the PR, looks great!

Could you add unit tests (pytests/ directory) and also an automatic runner for the docker-compose.yml that you've added. There are examples of such tests in tests directory now.

otto-liljalaakso-nt:
I have added the automatic runner test in `tests/`. I also tried looking at unit tests, but I would need some guidance here.

This PR touches two functions: `normalize_service()` and `build_one()`.

`normalize_service()` is called by some unit tests, but not in systematic fashion: The test files are called `test_can_merge_build.py`, which is about `build` key specifically, and `test_can_merge_cmd_ent.py`, which I do not really understand, but apparently is about command line arguments.

My question 1: Should I create a new file `test_can_merge_additional_contexts.py` and write my unit tests there, or what is the expectation here?

`build_one()` is not called by any unit test, and is a bit problematic for unit testing because it does filesystem access.

My question 2: Is the expectation that I should develop enough unit testing framework, with filesystem mocking etc., for `build_one()`, so I can unit test the changes in this pull request? I can create something, I am just not sure if assigning such task to first time, drive-by contributor is a good idea.

tchernobog:
Hi @p12tic , we were also looking for this feature to be implemented, and it seems like @otto-liljalaakso-nt addressed partially your review comments.

Could you answer his question and move this forward?

p12tic:
I'm sorry, I missed the comment by @otto-liljalaakso-nt.

> normalize_service() is called by some unit tests, but not in systematic fashion: The test files are called test_can_merge_build.py, which is about build key specifically, and test_can_merge_cmd_ent.py, which I do not really understand, but apparently is about command line arguments.

I cleaned up `normalize_service()` tests. They are now in pytests/test_normalize_service.py. Indeed the tests were hard to understand, hopefully no more.

> My question 1: Should I create a new file test_can_merge_additional_contexts.py and write my unit tests there, or what is the expectation here?

As of now the best place for normalize_service tests is in pytests/test_normalize_service.py.

> build_one() is not called by any unit test, and is a bit problematic for unit testing because it does filesystem access.

Agreed. But I think the change is simple enough that we can leave it untested. I feel that the amount of effort needed to test build_one() is too high for the benefits.

> My question 2: Is the expectation that I should develop enough unit testing framework, with filesystem mocking etc., for build_one(), so I can unit test the changes in this pull request? I can create something, I am just not sure if assigning such task to first time, drive-by contributor is a good idea.

Completely agreed. Lack of testing framework is problem of the project. Until there's one, the parts that are hard to test will unfortunately be untested.

otto-liljalaakso-nt:
@p12tic I have updated `tests/normalize_service.py` to test this change. CI is showing red, as did `pre-commit` which instructed by `CONTRIBUTING.md`. According to my analysis, none of the reported problems are caused by changes in this PR:

1. Black wants to reformat some files. Some of those are not changed in this PR. `tests/normalize_service.py` is changed in this PR, but the changed Black wants to apply are much wider than that. So I think those problems should not be fixed in this PR.
2. Test case `test_podman_compose_include` fails, but it also fails on `master` branch, so I do not think the problem should be fixed in this PR.

otto-liljalaakso-nt:
If you prefer, I can add a commit that runs Black and applies the changes it wants to make. For the failing test case, I don't know what it is about.

p12tic (approved):
Thanks, everything looks good.

p12tic:
> If you prefer, I can add a commit that runs Black and applies the changes it wants to make. 

I ran `ruff format .` after rebasing, no need to do anything. Thanks!

otto-liljalaakso-nt:
> > If you prefer, I can add a commit that runs Black and applies the changes it wants to make.
> 
> I ran `ruff format .` after rebasing, no need to do anything. Thanks!

Great to hear that.

I would like to suggest reviewing [pre-commit configuration](https://github.com/containers/podman-compose/blob/main/.pre-commit-config.yaml) and the suggestion to run that in CONTRIBUTING.md. It looks like pre-commit and CI are using different tools for the same purpose, leading to confusion like this.

PR #765, opened 2023-09-09
Added handling `pid` option

### LINKED ISSUE #554: ignores `pid: host`
**Describe the bug**

`podman-compose` (silently) ignores the PID namespace-related container deployment option, such as `pid: host`.

Copy and pasting the `podman create ...` command line and manually adding `--pid=host`, then starting the affected container starts this container correctly with the container attached to the initial PID namespace. So this ain't a `podman` issue, but on `podman-compose`'s turf.

```
podman-compose version
['podman', '--version', '']
using podman version: 4.2.0
podman --version 
podman version 4.2.0
exit code: 0
```

**To Reproduce**

any `docker-compose.yaml` with `pid: host`, such as:

```docker-compose
version: '2.4'
services: 
    lxkns:
        image: localhost/lxkns:latest
        read_only: true
        build: 
            context: ../..
            dockerfile: deployments/podman/Dockerfile
        command:
            - "/lxkns"
            - "--initialcgroup"
            - "--debug"
            - "--docker=unix:///proc/1/root/run/docker.sock"
            - "--containerd=/proc/1/root/run/containerd/containerd.sock"
            - "--podman=unix:///proc/1/root/run/podman/podman.sock"
        ports:
            - "5010:5010"
        user: "65534"
        security_opt:
            # apparmor only accepts a profile name, which must have been loaded
            # by the system admin. You might want to use the lxkns-apparmor
            # profile to be found in this directory.
            - apparmor:unconfined #- apparmor:lxkns
            # Please note that Docker's default container seccomp profile is
            # adaptive: it allows all those syscalls for which the process has
            # matching capabilities. In consequence, this Docker default seccomp
            # profile thus does not interfere with lxkns. However, we still want
            # to slightly curb the unneeded parts of the powers bestowed upon us
            # by especially CRAP_SYS_ADMIN and CAP_SYS_PTRACE...
            # - seccomp:lxkns-seccomp.json
        # Clean all default capabilities Docker might otherwise grant us and
        # several of them we don't need; add only those capabilities we really
        # need, such as CAP_WORLD_DOMINATION.
        cap_drop:
            - ALL
        cap_add:
            - SYS_ADMIN       # change namespaces
            - SYS_CHROOT      # change mount namespaces
            - SYS_PTRACE      # access nsfs namespace information
            - DAC_READ_SEARCH # for fully discovering bind-mounted namespaces
            - DAC_OVERRIDE    # for accessing the containerd API socket
        # We need the full PID view for correct scanning.
        pid: host
```

`podman-compose -p lxkns -f deployments/podman/docker-compose.yaml up` -- this is the composer file listed above.

**Expected behavior**

Container to be attached to the initial PID namespace; `podman-compose` to emit the `--pid=host` CLI argument when creating the service container.

**Actual behavior**

`podman-compose` silently ignores any `pid:` deployment settings.

**Environment:**
 - OS: Linux Ubuntu 22.04.1LTS
 - podman version: 4.2.0
 - podman compose version: 1.0.3

Nota bene: where do I get the "git hex" for `podman-compose` from when installing via `pip3`?

timocov:
Fixes #554


p12tic (changes requested):
Looks good, thanks for the PR.

Could unit tests be added? There's an example in `pytests/test_container_to_args.py`.

timocov:
@p12tic added tests and rebased onto main.

p12tic (approved):
Thanks a lot!

PR #771, opened 2023-09-13
Fix environment variable precedents

wgnathanael:
Allow environment variables set in the host os to override the value of a variable defined in a .env file.
This is consistent with docker's environment variable precedence. Per https://docs.docker.com/compose/environment-variables/envvars-precedence/#advanced-example

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1804,8 +1804,8 @@ class PodmanCompose:
         os.environ.update({
             key: value for key, value in dotenv_dict.items() if key.startswith("PODMAN_")
         })
-        self.environ = dict(os.environ)
-        self.environ.update(dotenv_dict)
+        self.environ = dotenv_dict
+        self.environ.update(dict(os.environ))
         # see: https://docs.docker.com/compose/reference/envvars/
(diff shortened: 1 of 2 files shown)


p12tic (changes requested):
Need proper test for this.

wgnathanael:
I'd love to submit some tests but I can't even find existing tests that test the env/env-file features as is. Is there one I can use a template someplace?

charliemirabile:
@p12tic bump on this issue. I (and presumably @wgnathanael) would love to get this merged as I am currently relying on it in my own fork to be able to use podman-compose in production, but it needs tests. Happy to contribute some, but as @wgnathanael says, there doesn't seem to be any infrastructure in place for testing the existing environment variable code. Do you have any sense of how this could be tested in a reasonably straightforward way?

p12tic:
Thanks for the bump, I somehow missed @wgnathanael comment. If there is no testing infrastructure for a feature, it's more of a responsibility of the maintainer to define one instead of infrequent contributors. I will merge this PR as is.

p12tic:
Rebased and fixed merge conflict.

p12tic (approved):

PR #772, opened 2023-09-16
Add --in-pod parameter for systemd commands

ttsak:
systemd scripts depend an a pod to exist but this is not created unless the in-pod parameter is added

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2000,7 +2000,7 @@ def compose_systemd(compose, args):
                     f.write(f"{k}={v}\n")
         print(f"writing [{fn}]: done.")
         print("\n\ncreating the pod without starting it: ...\n\n")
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

RalfSchwiete:
Hi,

really nice work the podman-composer. IMHO it shouldn't be
`ExecStartPre=-{script} --in-pod %i up --no-start`
but
`ExecStartPre=-{script} --in-pod pod_%i up --no-start`
in the unit file.

Also it might be good to have a "ExecStopPost" line to remove the pod:
`ExecStopPost=/usr/bin/podman pod rm pod_%i`

Thank you for your great work,
Ralf

p12tic:
`--in-pod` is now default, so this PR is no longer needed. Thanks a lot for caring about this feature.

PR #777, opened 2023-09-20
Pass file secrets to "podman build" via parameter "--secret"

wiehe:
Dear maintainers!

At the moment podman-compose always passes file secrets to podman as a `--volume` option, both when starting containers and when building images.
I propose that file secrets for building are instead passed as a `--secret` option. This would make them available for
`RUN --mount=type=secret…` instructions and would improve compatibility with docker-compose.

### Example

```
# Dockerfile
FROM busybox
RUN --mount=type=secret,required=true,id=my_secret,target=/root/my_secret \
		echo "The secret is: "; cat /root/my_secret
```
To build directly from this Dockerfile one has to supply the secret like this: 
`podman build --secret=id=my_secret,src=./my_secret .`

```
# docker-compose.yaml
services:
  my-service:
    build:
      context: .
      secrets:
        - my_secret

secrets:
  my_secret:
    file: ./my_secret
```

Building from the docker-compose.yaml with `podman-compose build my-service`  should produce the same results as the `podman build …` command above.

Yours sincerely
wiehe

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -527,7 +527,11 @@ async def get_mount_args(compose, cnt, volume):
     return ["--mount", args]
 
 
-def get_secret_args(compose, cnt, secret):
+def get_secret_args(compose, cnt, secret, podman_is_building=False):
+    """
+    podman_is_building: True if we are preparing arguments for an invocation of "podman build"
+                        False if we are preparing for something else like "podman run"
+    """
     secret_name = secret if is_str(secret) else secret.get("source", None)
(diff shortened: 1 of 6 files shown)


p12tic (changes requested):
The PR looks good.

Please rebase on top of latest `main` branch and add unit tests (see pytest/ directory for examples).

wiehe:
Hi, thanks for considering my pull request! I have rebased it onto `main` and ported the test script from `pytest` to `unittest`.

p12tic (approved):

PR #780, opened 2023-09-24
Support all healthcheck options in Compose

recursiveribbons:
- Added start_interval to the healthcheck block parser
- Changed the healthcheck flags to the normalized podman flags

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1045,13 +1045,15 @@ async def container_to_args(compose, cnt, detached=True):
         else:
             raise ValueError("'healthcheck.test' either a string or a list")
 
-    # interval, timeout and start_period are specified as durations.
+    # interval, timeout, start_period, and start_interval are specified as durations.
     if "interval" in healthcheck:
         podman_args.extend(["--health-interval", healthcheck["interval"]])
     if "timeout" in healthcheck:
         podman_args.extend(["--health-timeout", healthcheck["timeout"]])
     if "start_period" in healthcheck:
         podman_args.extend(["--health-start-period", healthcheck["start_period"]])
+    if "start_interval" in healthcheck:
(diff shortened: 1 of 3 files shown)


p12tic (changes requested):
Thanks for the PR. It looks good.

Please add unit tests. See e.g. pytests/test_container_to_args.py for an example.

recursiveribbons:
(rebased branch onto main and force pushed)

recursiveribbons:
I'm not sure how to fix the unit test, as `--health-startup-interval` does not exist in podman 4.3.1, which is the version shipped with Debian Bookworm (latest version is 4.9.3)

recursiveribbons:
Do you actually want me to commit the entire formatted podman_compose.py file or just let the test fail?

also the pre-commit checks wants me to format 10 files so I'm not sure if I should

podman_compose.py
```diff
@@ -1032,12 +1029,12 @@ async def container_to_args(compose, cnt, detached=True):
                 podman_args.append("--no-healthcheck")
             elif healthcheck_type == "CMD":
                 cmd_q = "' '".join([cmd_quote(i) for i in healthcheck_test])
-                podman_args.extend(["--healthcheck-command", "/bin/sh -c " + cmd_q])
+                podman_args.extend(["--health-cmd", "/bin/sh -c " + cmd_q])
```
d0b3rm4n, line 1032: I don't think in `CMD` case this should be executed with ` /bin/sh` prevents cases from working where there is no  `/bin/sh` in the image like minimal images for Rust applications. I don't have docker to verify, but from the documentation, my understanding is that this is ran without shell.

p12tic:
Merged via https://github.com/containers/podman-compose/pull/1271

PR #783, opened 2023-09-24
Fix #782: add support for `http-proxy` option of podman

### LINKED ISSUE #782: add support for `http-proxy` option
By default proxy environment variables are passed into the container if set for the Podman process. This can be disabled by setting the `http-proxy` value to **false**(defaults to **true**) with `--http-proxy=false`. Currently podman-compose does not support this feature.

If the yaml file contains `http-proxy: false`, then add  `--http-proxy=false` to args.

ChuJiani:
#782

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -889,6 +889,8 @@ async def container_to_args(compose, cnt, detached=True):
         podman_args.extend(["--annotation", a])
     if cnt.get("read_only", None):
         podman_args.append("--read-only")
+    if cnt.get("http_proxy", None) is False:
+        podman_args.append("--http-proxy=false")
     for i in cnt.get("labels", []):
         podman_args.extend(["--label", i])
     for c in cnt.get("cap_add", []):
--- a/pytests/test_container_to_args.py
+++ b/pytests/test_container_to_args.py
@@ -152,3 +152,24 @@ class TestContainerToArgs(unittest.IsolatedAsyncioTestCase):
                 "busybox",
             ],
         )
+
+    async def test_http_proxy(self):
+        c = create_compose_mock()
+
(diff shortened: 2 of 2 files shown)


JAORMX (approved):

NerdToMars (approved):

p12tic (changes requested):
Thanks for the PR.

Please add unit tests. See e.g. pytests/test_container_to_args.py for an example.

ChuJiani:
Unit test added in `pytests/test_container_to_args.py`.

Are there any need to place this option into or behind [net_args](https://github.com/containers/podman-compose/blob/121b5f6ea1b6bc1f8ea4ddf4ebc7ff809d5b876e/podman_compose.py#L926)?

p12tic:
> Also: Do I need to place this option into or behind net args

There's stuff like `--dns` ten lines below your change, so I think it doesn't matter for now.

The PR looks good. Could you rebase on top of `main` instead of merging? This whole PR can be a single commit.

podman_compose.py
```diff
@@ -889,6 +889,8 @@ async def container_to_args(compose, cnt, detached=True):
         podman_args.extend(["--annotation", a])
     if cnt.get("read_only", None):
         podman_args.append("--read-only")
+    if cnt.get("http_proxy", None) == False:
```
p12tic, line 892: This should be `is False`.

ChuJiani:
Sure! I will figure it out.

ChuJiani:
Sorry I messed things up. It took me some time to rebase/reset. But It should work fine now.

By the way, it seems that the commit just before mine, which is about `pid`, does not pass the unit test.

p12tic:
Thanks, looks great!

PR #796, opened 2023-11-02
Print full dockerfile path instead of context on error

siteshwar:
... if the dockerfile is not found.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2120,7 +2120,7 @@ def build_one(compose, args, cnt):
             if os.path.exists(dockerfile):
                 break
     if not os.path.exists(dockerfile):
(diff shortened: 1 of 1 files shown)


siteshwar:
Current error message is confusing and does not provide the exact path where the dockerfile is being searched. This change should help in debugging such cases.

p12tic (changes requested):
Thanks for the PR.

Could you please add examples of previous error message and current error message to the commit message? Also, the first line of the commit message needs to be 72 characters or less, otherwise it's hard wrapped by tools and people see broken sentence.

Also, please rebase on top of latest main branch.

p12tic:
Merged via https://github.com/containers/podman-compose/pull/1222.

PR #804, opened 2023-11-18
Support multiple file includes and include cycles in include section

jasyip:
Previously, include sections had to be lists of size 1 and never properly addressed other sizes. Include cycles were possible to cause infinite loops. This commit supports including multiple compose files and can detect cyclic include behavior. This commit also parses each included file only once as an optimization.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -14,6 +14,7 @@ import sys
 import os
 import getpass
 import argparse
+import copy
 import itertools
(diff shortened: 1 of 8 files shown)


p12tic (changes requested):
Thanks for the PR.

In order for it to be accepted, unit tests and an automatic integration test will be needed.

PR #805, opened 2023-11-21
 Fix COMPOSE_PROJECT_NAME variable parsing order regression

nkay08:
In a previous change (9d5b255) the support to read `COMPOSE_PROJECT_NAME` from the `name` attribute in the `compose.yml` file was added. Since the attribute is only available after parsing the compose file, the resolution of the variable was done afterwards.

The variable is therefore usable inside the container, HOWEVER it can no longer be used for substitution while loading the compose file. Values that depend on this variable are therefore empty at the point of parsing the compose file.

`COMPOSE_PROJECT_NAME` is either loaded from the environment or set to `dir_basename` by default.

This commit changes the order of parsing `COMPOSE_PROJECT_NAME ` variable. First the compose file is loaded into a dict, then the single `name` attribute is evaluated and replaced, then if it does not exist, the default value is used.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1611,6 +1611,22 @@ class PodmanCompose:
                     sys.exit(1)
                 content = normalize(content)
                 # log(filename, json.dumps(content, indent = 2))
+
+                if not project_name:
+                    project_name = rec_subs(content.get("name", None), self.environ)
+                    if project_name is None:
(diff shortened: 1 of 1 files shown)


p12tic (changes requested):
Looks good, but we will need a unit test for this.

Cabanon:
I would really like to see this PR merged, this regression has broken many of my deployments

p12tic:
This fix has been merged via https://github.com/containers/podman-compose/pull/1165.

PR #830, opened 2024-02-01
Implement override and reset analog to docker-compose

SebTM:
https://docs.docker.com/compose/compose-file/13-merge/

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1149,6 +1149,47 @@ def flat_deps(services, with_extends=False):
     for name, srv in services.items():
         rec_deps(services, name)
 
+###################
+# Override and reset tags
+###################
+
+class OverrideTag(yaml.YAMLObject):
+    yaml_dumper = yaml.Dumper
+    yaml_loader = yaml.SafeLoader
+    yaml_tag = u'!override'
+
+    def __init__(self, value):
+        values = list()
+
+        for item in value:
+            values.append(item.value)
+
+        self.value = values
+
+    @classmethod
+    def from_yaml(cls, loader, node):
+        return OverrideTag(node.value)
+
+    @classmethod
+    def to_yaml(cls, dumper, data):
+        return dumper.represent_scalar(cls.yaml_tag, data.value)
+
+class ResetTag(yaml.YAMLObject):
+    yaml_dumper = yaml.Dumper
+    yaml_loader = yaml.SafeLoader
+    yaml_tag = u'!reset'
+
+    @classmethod
+    def toJSON(self):
+        return self.yaml_tag
+
+    @classmethod
+    def from_yaml(cls, loader, node):
+        return ResetTag()
+
+    @classmethod
+    def to_yaml(cls, dumper, data):
+        return dumper.represent_scalar(cls.yaml_tag, '')
 
 ###################
 # podman and compose classes
@@ -1245,6 +1286,14 @@ class Podman:
 
 
 def normalize_service(service, sub_dir=""):
+    if isinstance(service, ResetTag):
+        return service
+
+    if isinstance(service, OverrideTag):
+        tagObj = service
+        service = tagObj.value
+        print(subdir)
+
     if "build" in service:
         build = service["build"]
         if is_str(build):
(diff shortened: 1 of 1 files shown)


podman_compose.py
```diff
@@ -1284,7 +1333,12 @@ def normalize_service(service, sub_dir=""):
         if is_str(extends):
             extends = {"service": extends}
             service["extends"] = extends
-    return service
+
+    try:
+        tagObj.value = serivce
```
Vininhos, line 1338: Is this a typo?

SebTM: No you need to still normalize the values of the Override-tag to be valid in the end (as they are the ones used) but pass the override-tag through to the recursive merge to do the actual override. I used try to determinate if there is a tagObj defiend and if so get the normalized values to the tagObj and return it otherwise return the normalized service-definition.

SebTM: Also this isn't my most recent version anymore as I'm locally working an a version based on the latest https://github.com/maurerle/podman-compose but for now using docker-compose again.

p12tic: `serivce` seems to be non-existing variable though.

SebTM: It is, I'm not coming out of the python-space so I searched how to return it if set and another if not

SebTM: Edit: You are all right - missed the typo but that's not an issue/the latest state and needs a rebase.

How likely are my chances to get this approach merged, if I resubmit it?

p12tic: Good question, haven't looked into viability of the PR so far. Could you point me to the part of the compose spec that this implements? I didn't find with couple minutes of googling and better I don't make an error here :)

Overall I think podman-compose should implement all things possible in non-swarm compose that can be implemented on podman. We may argue about the approach of implementation though.

What I will definitely ask that's not part of the PR right now is unit and integration tests.

podman_compose.py
```diff
@@ -1149,6 +1149,47 @@ def flat_deps(services, with_extends=False):
...
         rec_deps(services, name)
 
+###################
+# Override and reset tags
+###################
+
+class OverrideTag(yaml.YAMLObject):
+    yaml_dumper = yaml.Dumper
+    yaml_loader = yaml.SafeLoader
+    yaml_tag = u'!override'
```
p12tic, line 1159: If I remember correctly, `u` prefix is from python 2 times. Am I correct that it's not needed here?

podman_compose.py
```diff
@@ -1629,7 +1709,7 @@ def _parse_compose_file(self):
         compose["services"] = resolved_services
         if not getattr(args, "no_normalize", None):
             compose = normalize_final(compose, self.dirname)
-        self.merged_yaml = yaml.safe_dump(compose)
+        self.merged_yaml = yaml.dump(compose)
```
p12tic, line 1712: Is this strictly needed?

podman_compose.py
```diff
@@ -1245,6 +1286,14 @@ def volume_ls(self, proj=None):
 
 
 def normalize_service(service, sub_dir=""):
+    if isinstance(service, ResetTag):
+        return service
+
+    if isinstance(service, OverrideTag):
+        tagObj = service
```
p12tic, line 1293: Use tag_obj spelling.

podman_compose.py
```diff
@@ -1245,6 +1286,14 @@ def volume_ls(self, proj=None):
...
+        service = tagObj.value
+        print(subdir)
```
p12tic, line 1295: Unneeded print.

p12tic (changes requested):
Looks good in principle.

Please add unit and integration tests.

SebTM:
Just to let you all know - I'm still interested in this feature but I have to put the PR from my side on-hold so if someone needs it / want's to take over - feel free 🙏🏻

FrankyBoy:
Just ran into these tags not working, which is a real issue for me.

mokibit:
Working on completing this PR.

p12tic:
Merged via https://github.com/containers/podman-compose/pull/1189.

Issue #831, opened 2024-02-05
Unable to run podman-compose`coroutine 'main' was never awaited` 


plessbd:
**Describe the bug**
Using the latest devel branch
```
[root ~]# pip3 install https://github.com/containers/podman-compose/archive/devel.tar.gz --no-cache-dir -U
Collecting https://github.com/containers/podman-compose/archive/devel.tar.gz
  Downloading https://github.com/containers/podman-compose/archive/devel.tar.gz
     \ 66 kB 209 kB/s
Requirement already satisfied: pyyaml in /usr/lib64/python3.9/site-packages (from podman-compose==1.0.7) (5.4.1)
Requirement already satisfied: python-dotenv in /usr/lib/python3.9/site-packages (from podman-compose==1.0.7) (0.19.2)
Building wheels for collected packages: podman-compose
  Building wheel for podman-compose (setup.py) ... done
  Created wheel for podman-compose: filename=podman_compose-1.0.7-py2.py3-none-any.whl size=36590 sha256=7a54cac3d47be895c4df0059b720cd0f10fc2cefed09a810bf9025cd1ed2985b
  Stored in directory: /tmp/pip-ephem-wheel-cache-7whykevl/wheels/40/e0/ea/7d6cae5c1132c86dbe88a6dbd9588186a74f39feee1e4ef96f
Successfully built podman-compose
Installing collected packages: podman-compose
Successfully installed podman-compose-1.0.7

[root ~]# podman-compose -v
<coroutine object main at 0x7fd33f8f7640>
sys:1: RuntimeWarning: coroutine 'main' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

**To Reproduce**
1. run `pip3 install https://github.com/containers/podman-compose/archive/devel.tar.gz`
2. run `podman-compose -v`

**Expected behavior**
podman-compose should work

**Actual behavior**
What is the behavior you actually got and that should not happen.

**Output**

**Environment:**
 - OS: RHEL 9.3

 - podman compose version: c5be5ba

**Additional context**

I belive this has to do with 
[root ~]# cat `which podman-compose`
```
#!/usr/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from podman_compose import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
as doing
```
[...]

PR #832, opened 2024-02-05
Fix issue #831

plessbd:
Fixes issue with asyncio

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -3239,9 +3239,11 @@ def compose_format_parse(parser):
         help="Pretty-print container statistics to JSON or using a Go template",
     )
 
-async def main():
+async def async_main():
     await podman_compose.run()
 
+def main():
+    asyncio.run(main())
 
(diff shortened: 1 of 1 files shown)


plessbd:
I do not belive these failures are "mine" but if there is something I should fix please let me know

podman_compose.py
```diff
@@ -3239,9 +3239,11 @@ def compose_format_parse(parser):
         help="Pretty-print container statistics to JSON or using a Go template",
     )
 
-async def main():
+async def async_main():
     await podman_compose.run()
 
+def main():
+    asyncio.run(main())
```
muayyad-alsadi, line 3246: I guess this is a typo, you made it recursive

it should be

```
    asyncio.run(async_main())
```

but I don't think this is the problem

plessbd: yeap, that was a typo, to many main and async and main ;)

Thank you for the fix

PR #851, opened 2024-02-20
fix(ipam_driver): do not pass --ipam-driver option when value set to …

fccagou:
…default

podman returns _unsupported ipam driver "default"_ when `--imap-driver default` parameter is passed.

So, when ipam driver is set to "default" in docker-compose.yml, --imap-driver must not be used when podman is called.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -734,7 +734,7 @@ def get_network_create_args(net_desc, proj_name, net_name):
         args.extend(("--opt", f"{key}={value}"))
     ipam = net_desc.get("ipam", None) or {}
     ipam_driver = ipam.get("driver", None)
-    if ipam_driver:
+    if ipam_driver and ipam_driver != "default":
         args.extend(("--ipam-driver", ipam_driver))
     ipam_config_ls = ipam.get("config", None) or []
     if net_desc.get("enable_ipv6", None):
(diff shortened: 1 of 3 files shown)


fccagou:
Sorry for the push force. 
Nothing changed in patch except commit message to refer to the new issue #852 according to CONTRIBUTING.md.

podman_compose.py
```diff
@@ -762,7 +762,7 @@ async def assert_cnt_nets(compose, cnt):
                 args.extend(("--opt", f"{key}={value}"))
             ipam = net_desc.get("ipam", None) or {}
             ipam_driver = ipam.get("driver", None)
-            if ipam_driver:
+            if ipam_driver and ipam_driver not in ("default"):
```
p12tic, line 765: I think the code would be clearer if we compared using `!=`. Also I think there's a bug here, because `("default")` is not a tuple. `("default",)` would be a tuple.

p12tic (changes requested):
Looks good in principle but will need unit tests.

podman_compose.py
fccagou: You are right, the missing ',' is strange because my tests worked ... 
Yes, we can compare using `!=`.
I do a patch with it.

p12tic:
Rebased and adjusted commit message.

p12tic:
@fccagou Sorry for the delay. I haven't noticed that you've force pushed with the fixes.

p12tic (approved):

Issue #798, opened 2023-11-08
podman-compose run -e doesn't like values with equals signs


aripollak:
**Describe the bug**
Passing an environment variable value that contains equals signs to `podman-compose run -e`, like base64 results used in e.g. AWS_SESSION_TOKEN, crashes podman-compose.

**To Reproduce**
With a simple docker-compose.yml:
```
services:
  bash:
    image: bash
```

Run:
`podman-compose run --rm -e FOO=bar= bash -c 'echo $FOO'`

**Expected behavior**
I expected this to print out `bar=`, like what happens if you run `podman run --rm -e FOO=bar= bash -c 'echo $FOO'`.

**Actual behavior**
```
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.3.1
Traceback (most recent call last):
  File "/home/ari/bin/podman-compose", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/ari/.local/pipx/venvs/podman-compose/lib/python3.11/site-packages/podman_compose.py", line 2941, in main
    podman_compose.run()
  File "/home/ari/.local/pipx/venvs/podman-compose/lib/python3.11/site-packages/podman_compose.py", line 1423, in run
    cmd(self, args)
  File "/home/ari/.local/pipx/venvs/podman-compose/lib/python3.11/site-packages/podman_compose.py", line 1754, in wrapped
    return func(*args, **kw)
           ^^^^^^^^^^^^^^^^^
  File "/home/ari/.local/pipx/venvs/podman-compose/lib/python3.11/site-packages/podman_compose.py", line 2269, in compose_run
    additional_env_vars = dict(map(lambda each: each.split("="), args.env))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: dictionary update sequence element #0 has length 4; 2 is required
```

**Output**

```
$ podman-compose version
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.3.1
podman-compose version 1.0.6
podman --version 
podman version 4.3.1
exit code: 0


$ podman-compose up
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.3.1
** excluding:  set()
['podman', 'ps', '--filter', 'label=io.podman.compose.project=ari', '-a', '--format', '{{ index .Labels "io.podman.compose.config-hash"}}']
[...]

PR #860, opened 2024-03-01
Allow run/exec -e with equals sign in value

aripollak:
Fixes: #798

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2502,6 +2502,12 @@ async def compose_exec(compose, args):
     container_names = compose.container_names_by_service[args.service]
     container_name = container_names[args.index - 1]
     cnt = compose.container_by_name[container_name]
+    podman_args = compose_exec_args(cnt, container_name, args)
+    p = await compose.podman.run([], "exec", podman_args)
+    sys.exit(p)
+
+
+def compose_exec_args(cnt, container_name, args):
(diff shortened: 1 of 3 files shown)


p12tic (changes requested):
Needs unit tests. Right now there's no framework to write tests for root compose_run easily, so it may not make sense to write them just right now.

aripollak:
Happy to add unit tests where appropriate, but I'm a little confused; is there a place that would make sense to add them now?

p12tic:
On a second look a different approach is possible in this particular case.

`compose_run()` contains a lot of argument processing that could be moved to a separate function (basically everything here https://github.com/containers/podman-compose/blob/main/podman_compose.py#L2419C39-L2455C42). Then we could add unit tests for that function.

aripollak:
@p12tic how does it look now? Specfying all the expected args in the passed `argparse.Namespaces` in the tests seemed a little duplicative, but that felt more appropriate for a unit test than calling into `PodmanCompose._parse_args()`. I think the interface of `compose_run_update_container_from_args()` would be more ideal if it wasn't changing the container dict, but I'm not familiar enough with the code to go and change it too much.

p12tic:
@aripollak I agree with the entirety of your comment.

The PR looks good, thanks!

p12tic (approved):

PR #885, opened 2024-03-09
support volume mapping into the home of a container

maurerle:
It is supported in docker compose to have a volume mapping like `/foo:~/data` which should also be supported by podman compose.

This PR allows using this too.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -143,8 +143,9 @@ def parse_short_mount(mount_str, basedir):
     elif len(mount_a) == 2:
         mount_src, mount_dst = mount_a
         # dest must start with / like /foo:/var/lib/mysql
+        # or with ~ like /foo:~/data
         # otherwise it's option like /var/lib/mysql:rw
-        if not mount_dst.startswith("/"):
+        if not mount_dst.startswith("/") and not mount_dst.startswith("~") :
             mount_dst, mount_opt = mount_a
             mount_src = None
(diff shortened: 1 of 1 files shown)


p12tic:
@maurerle Thanks for the PR. Is tilde volume mapping specified in the compose spec? I couldn't find it.

p12tic (changes requested):
Thanks for the PR, it looks good.

This will need unit tests for the `parse_short_mount()` function. But before spending time on these let's wait for me to understand what part of compose spec or undocumented docker compose behavior we are dealing with.

maurerle:
I think that this is not documented here:
https://github.com/compose-spec/compose-spec/blob/master/07-volumes.md

It just says that the destination can be a part inside the docker container - so any valid unix path would be fine, including those starting with a tilde `~`
If it does not start with an absolute path - say plain `data`, `docker compose` says `invalid mount path: 'data' mount path must be absolute` - so ~ is interpreted as a valid absolute path..

p12tic:
OK, replicating undocumented behavior is fine. I was just concerned that this is documented in the spec but I can't find it.

maurerle:
I don't know what happened but somehow this does not work with docker compose either.. `~/data:rw': invalid mount config for type "bind": invalid mount path: '~/data' mount path must be absolute`
Weird - problem some wrong assumptions i had when testing before..
Sorry for that.

PR #869, opened 2024-03-07
Fix CI tests

baszoetekouw:
This should fix the pylint and flake8 checks in CI.

PR #887, opened 2024-03-09
Fix support for older python versions

baszoetekouw:
This PR replaces #869 
It retains the changes needed to support older python versions.

Also, it makes some changes to the CI, notably:
 - re-enable the pylint tests, as this catch some more stuff than ruff on its own.  I can remove this part, if you prefer
 - run the tests for all upstream-supported versions of python (3.8+) 
 - split out the tests in separate GA steps, and reduce verbosity, so that it's easier to spot which tests fail.

As a side note: because coverage now is checked in the CI, it might make sense to enable a tool like https://codecov.io/, which provides a nice web interface showing where coverage is missing, and also adds an automatic CI check to check that new PRs don't decrease coverage.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = tab
+tab_width = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 100
+
+[*.{yml,yaml}]
+indent_style = space
+indent_size = 2
+
+[*.py]
+indent_style = space
+
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,11 +26,15 @@ jobs:
         python -m pip install --upgrade pip
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
         if [ -f test-requirements.txt ]; then pip install -r test-requirements.txt; fi
-    - name: Test with unittest
+    - name: Run tests in tests/
       run: |
-        coverage run --source podman_compose -m unittest pytests/*.py
         python -m unittest tests/*.py
-        coverage combine
-        coverage report
       env:
         TESTS_DEBUG: 1
+    - name: Run tests in pytests/
+      run: |
+        coverage run --source podman_compose -m unittest pytests/*.py
+    - name: Report coverage
+      run: |
+        coverage combine
+        coverage report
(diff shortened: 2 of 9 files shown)


.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+
+[*]
+indent_style = space
+indent_size = tab
+tab_width = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 120
```
p12tic, line 11: 100

.github/workflows/test.yml
```diff
@@ -21,11 +26,13 @@ jobs:
...
+    - name: Run tests in pytests/
       run: |
         coverage run --source podman_compose -m unittest pytests/*.py
-        python -m unittest tests/*.py
+    - name: Report coverage
+      run: |
         coverage combine
         coverage report
-      env:
-        TESTS_DEBUG: 1
```
p12tic, line 31: This is needed to spot errors in case tests fail in CI.

.github/workflows/test.yml
```diff
@@ -21,11 +26,13 @@ jobs:
         python -m pip install --upgrade pip
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
         if [ -f test-requirements.txt ]; then pip install -r test-requirements.txt; fi
```
p12tic, line 28: If pylint is added to CI then it should be added to test-requirements.txt for installation.

.github/workflows/test.yml
p12tic: The failing tests usually print a nice log in the end. Is that not sufficient?

p12tic:
In `Fix python>3.11 compatibility` what was mean is probably ``Fix python<3.11 compatibility`.

`Split out the different tests in separate CI steps ...` commit message is too long, should be max 72 characters. Now it overflows.

p12tic:
Overall looks good. I stand corrected, can bring back pylint.

baszoetekouw:
Thanks for the feedback!  This should fix the outstanding issues.

p12tic:
[Match editorconfig max line length with ruff config](https://github.com/containers/podman-compose/pull/887/commits/63d0be0403e6e4f3132c5ab1fb86591a218dd7ab) should be part of 
[add editorconfig](https://github.com/containers/podman-compose/pull/887/commits/ce4ddce833fa173efc1eeaff20b5d6e6607e3ccc). Aside from that, all is good. Thanks!

baszoetekouw:
done!

p12tic (approved):

PR #898, opened 2024-03-20
Add regression testing for repositories that support pipelines 

MikeKlem:
All regression tests are passing for dockerhub images:
<https://gitlab.com/mikeklem/python-podman-compose/-/pipelines/1220271500>


p12tic:
Not sure what this PR is trying to achieve. podman-compose uses Github Actions as its CI.

Issue #704, opened 2023-06-07
Error mount volume /run/podman/podman.sock:/var/run/docker.sock becomes E:\run\podman\podman.sock:/var/run/docker.sock:Z


SamLowryMOI:
**Describe the bug**
I try to use portainer. On ```podman-compose up -d``` ```- /run/podman/podman.sock:/var/run/docker.sock``` in docker-compose.yml becomes ```-v E:\run\podman\podman.sock:/var/run/docker.sock:Z```

**To Reproduce**
docker-compose.yml on E:\Docker\portainer
```
version: '3'

services:
  portainer:
    image: portainer/portainer-ce:alpine
    restart: unless-stopped
    container_name: portainer_master
    ports:
      - 13000:9000
      - 13443:9443
      - 8000:8000
    volumes:
      - /run/podman/podman.sock:/var/run/docker.sock:Z
      - ./portainer_data:/data
```

**Output**
```
E:\Docker\portainer>dir /s
 Datenträger in Laufwerk E: ist Data
 Volumeseriennummer: XXXX-XXXX

  Verzeichnis von E:\Docker\portainer

07.06.2023  22:35    <DIR>          .
07.06.2023  22:35    <DIR>          ..
07.06.2023  22:21               563 docker-compose.yml
07.06.2023  22:23    <DIR>          portainer_data
               1 Datei(en),            563 Bytes

 Verzeichnis von E:\Docker\portainer\portainer_data3

07.06.2023  22:35    <DIR>          .
07.06.2023  22:35    <DIR>          ..
               0 Datei(en),              0 Bytes
[...]

E:\Docker\portainer>podman-compose version
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.5.1
podman-compose version 1.0.6
podman --version
podman version 4.5.1
exit code: 0

E:\Docker\portainer>podman-compose up -d
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.5.1
** excluding:  set()
['podman', 'ps', '--filter', 'label=io.podman.compose.project=portainer', '-a', '--format', '{{ index .Labels "io.podman.compose.config-hash"}}']
['podman', 'network', 'exists', 'portainer_default']
[...]

PR #905, opened 2024-04-07
FIX issue https://github.com/containers/podman-compose/issues/704 - Windows fix sock not working.

acotty:
From issue 704:
**Describe the bug**
I try to use portainer. On podman-compose up -d - /run/podman/podman.sock:/var/run/docker.sock in docker-compose.yml becomes -v E:\run\podman\podman.sock:/var/run/docker.sock:Z

.....snuip....

**Expected behavior**

-v \run\podman\podman.sock:/var/run/docker.sock:Z

instead of

**Actual behavior**

-v E:\run\podman\podman.sock:/var/run/docker.sock:Z

The issue is for portainer, but it also affects all sock on Windows due to the code not being windows aware.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -159,7 +159,8 @@ def parse_short_mount(mount_str, basedir):
         # User-relative path
         # - ~/configs:/etc/configs/:ro
         mount_type = "bind"
-        mount_src = os.path.abspath(os.path.join(basedir, os.path.expanduser(mount_src)))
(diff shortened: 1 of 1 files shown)


muayyad-alsadi:
the problem is that `.sock` is just a convention.

>  but it also affects all sock on Windows due to the code not being windows aware.

I would love the code to be OS agnostic as much as possible.

p12tic:
@muayyad-alsadi I think this PR should have included a test before being merged.

PR #906, opened 2024-04-08
Fix, use "service.build.ulimits" when it is present instead of "service.ulimits"

camspiers:
https://github.com/compose-spec/compose-spec/pull/433

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -441,8 +441,7 @@ def mount_desc_to_mount_args(compose, mount_desc, srv_name, cnt_name):  # pylint
     raise ValueError("unknown mount type:" + mount_type)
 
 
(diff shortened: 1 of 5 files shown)


p12tic (changes requested):
Looks good in principle, but needs unit tests.

camspiers:
Hey @p12tic thanks for reviewing the PR. I have updated with unit tests, but I am not 100% on what I have done.

At the moment I am testing both that the ulimits are passed correctly to the build command, and that the ulimits are applied during the build. However I could solely test that the right arguments are passed to the build command using the dry-run flag and remove the actual execution of the ulimit command in the build.

p12tic (approved):

p12tic:
Rebased and shortened commit message to max 72 chars allowed by git.

PR #909, opened 2024-04-08
Fixes #908: Add support for multi-line environment files

### LINKED ISSUE #908: Multi-line environment files are not correctly processed
**Describe the bug**
For environment files, podman-compose passes them to `podman` with its `--env-file` option. However, `podman/docker` and `docker-compose` process environment files differently: `podman/docker` doesn't support multi-line environment files, while docker compose supports them (https://github.com/containers/podman/issues/18724#issuecomment-1757162444). 

**Expected behavior**
Properly support multi-line env files

**Actual behavior**
Just passes env-files to podman, which doesn't support multi-line env vars, as apparently docker also doesn't support them too.

** Implementation **
I found that it can be easily implemented (parse env file with dotenv, and pass variables to podman), and I'll open a PR to fix it too.

hedayat:
TBH, I don't find the part of compose spec which specifies this. Actually, it might not be according to the spec. However, this is what docker compose supports... !

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -987,7 +987,12 @@ async def container_to_args(compose, cnt, detached=True):
         env_file = [env_file]
     for i in env_file:
         i = os.path.realpath(os.path.join(dirname, i))
-        podman_args.extend(["--env-file", i])
+        dotenv_dict = {}
+        dotenv_dict = dotenv_to_dict(i)
+        env = norm_as_list(dotenv_dict)
+        for e in env:
+            podman_args.extend(["-e", e])
+
     env = norm_as_list(cnt.get("environment", {}))
     for e in env:
         podman_args.extend(["-e", e])


podman_compose.py
```diff
@@ -987,7 +987,12 @@ async def container_to_args(compose, cnt, detached=True):
         env_file = [env_file]
     for i in env_file:
         i = os.path.realpath(os.path.join(dirname, i))
-        podman_args.extend(["--env-file", i])
+        dotenv_dict = {}
```
p12tic, line 990: Needs a test. See pytests directory

p12tic (changes requested):
Looks good, but needs tests

p12tic:
This PR has been merged via https://github.com/containers/podman-compose/pull/949.

mokibit:
Tests added in https://github.com/containers/podman-compose/pull/949.

PR #911, opened 2024-04-10
Add support for selinux in verbose bind mount specification

charliemirabile:
As specified here in the compose schema: https://github.com/compose-spec/compose-spec/blob/c35d19323fdfd7dbf9db30022498e29ef5ebeab2/schema/compose-spec.json#L382-L388

and described here in the compose spec: https://github.com/compose-spec/compose-spec/blob/master/05-services.md#long-syntax-5

it should be possible to specify selinux relabeling for a bind mount specified using the verbose syntax (as opposed to the terse syntax which already works). This PR adds support for this syntax in podman compose to achieve parity with docker compose in this area.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -431,6 +431,11 @@ def mount_desc_to_mount_args(compose, mount_desc, srv_name, cnt_name):  # pylint
         tmpfs_mode = tmpfs_opts.get("mode", None)
         if tmpfs_mode:
             opts.append(f"tmpfs-mode={tmpfs_mode}")
+    if mount_type == "bind":
+        bind_opts = mount_desc.get("bind", {})
+        selinux = bind_opts.get("selinux", None)
+        if selinux is not None:
+            opts.append(selinux)
     opts = ",".join(opts)
     if mount_type == "bind":
         return f"type=bind,source={source},destination={target},{opts}".rstrip(",")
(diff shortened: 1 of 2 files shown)


podman_compose.py
```diff
@@ -431,6 +431,11 @@ def mount_desc_to_mount_args(compose, mount_desc, srv_name, cnt_name):  # pylint
         tmpfs_mode = tmpfs_opts.get("mode", None)
         if tmpfs_mode:
             opts.append(f"tmpfs-mode={tmpfs_mode}")
+    if mount_type == "bind":
```
p12tic, line 434: We should add a test for the function this code lives in in pytests/

podman_compose.py
```diff
@@ -486,6 +491,12 @@ def mount_desc_to_volume_args(compose, mount_desc, srv_name, cnt_name):  # pylin
     read_only = mount_desc.get("read_only", None)
     if read_only is not None:
         opts.append("ro" if read_only else "rw")
+    if mount_type == "bind":
```
p12tic, line 494: We should add a test for the function this code lives in in pytests/

p12tic (changes requested):
Looks good, but needs tests.

p12tic:
@muayyad-alsadi Please don't merge PRs that have unaddressed concerns raised

PR #913, opened 2024-04-17
Use x- prefix for x-podman.mac_address

p12tic:
Docker compose documents x- prefix as standard vendor prefix.

https://github.com/compose-spec/compose-spec/pull/18


p12tic:
cc @baszoetekouw

PR #918, opened 2024-04-28
Fix handling of --in-pod argument

p12tic:
Currently --in-pod handling is broken because the only way to set False is by providing empty argument like "--in-pod=". As of Python 3.7 the solution is to accept string and parse manually.


p12tic:
This PR reuses ideas by @StaticRocket in https://github.com/containers/podman-compose/pull/712. Thanks a lot for opening that PR.

PR #920, opened 2024-04-30
Add supoort for enabling GPU access

### LINKED ISSUE #919: FR: Enabling GPU access to service containers
Currently, the following configuration is not work on podman-compose.

```yaml
services:
  test:
    image: nvidia/cuda:12.3.1-base-ubuntu20.04
    command: nvidia-smi
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: 1
              capabilities: [gpu]
```

https://docs.docker.com/compose/gpu-support/

mokeyish:
closes #919

### DIFF
--- a/pytests/test_container_to_args.py
+++ b/pytests/test_container_to_args.py
@@ -325,3 +325,106 @@ class TestContainerToArgs(unittest.IsolatedAsyncioTestCase):
                 "busybox",
             ],
         )
+
+    async def test_gpu(self):
+        c = create_compose_mock()
+
+        cnt = get_minimal_container()
+        cnt["command"] = ["nvidia-smi"]
+        cnt["deploy"] = {"resources": {"reservations": {"devices": [{}]}}}
+
+        # count: all
+        cnt["deploy"]["resources"]["reservations"]["devices"][0] = {
+            "driver": "nvidia",
+            "count": "all",
+            "capabilities": ["gpu"],
+        }
+
+        args = await container_to_args(c, cnt)
+        self.assertEqual(
+            args,
+            [
+                "--name=project_name_service_name1",
+                "-d",
+                "--network=bridge",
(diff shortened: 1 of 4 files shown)


p12tic (changes requested):
Looks good, thanks.

Could you add unit tests? There's a bunch already in pytests/test_container_to_args.py. It should be easy to extend the test suite for the GPU options. The existing tests test `container_to_args()` which I think is appropriate level of detail for the additional GPU tests.

mokeyish:
@p12tic  I looked at the existing unit tests and it was a bit difficult to understand. It seems to be output after execution. But Github Actions does not have a GPU and cannot be tested.

p12tic:
@mokeyish Tests in pytests/test_container_to_args.py don't actually run any containers.

The test structure is as follows:
 - create compose definition in parsed form (i.e. python dicts, lists, etc. are used)
 - run it through container_to_args function
 - the result is arguments which would be otherwise sent to podman. The test simply compares them to what's expected and does not run podman.
 
With such tests any feature of podman-compose can be tested without any requirements from hardware.

p12tic:
So in GPU tests I would expect tests that check whether correct GPU-related options are added to podman_args.

pytests/test_container_to_args.py
```diff
@@ -325,3 +325,106 @@ async def test_env_file_obj_optional(self):
                 "busybox",
             ],
         )
+
+    async def test_gpu(self):
```
p12tic, line 329: This test is actually 4 tests. Could you split it. It's ok if the preamble

```
        c = create_compose_mock()

        cnt = get_minimal_container()
        cnt["command"] = ["nvidia-smi"]
        cnt["deploy"] = {"resources": {"reservations": {"devices": [{}]}}}
```

repeats in each test.

PR #923, opened 2024-05-01
Don't create pods/container when registering systemd unit

### LINKED ISSUE #922: Registering a systemd unit should not create pods or containers
#### Problem:
The command `podman-compose systemd -a register` internally issues an `up --no-start` command. This is an unexpected side effect. There is no need to bring up a the pod during registration, and the code does not use the result of it in any way except for logging it.

In my case this lead to naming conflicts since I first had a dummy `compose.yml` and then later changed it's contents. Some names were now already in use by containers that were created but not started, and therefore also hard to spot.

The output of the `up --no-start` command is also not useful when re-generating the systemd unit file while the compose stack is already up, as it will just interfere with the running compose setup and show errors for the container names already in use.

https://github.com/containers/podman-compose/blob/27e27e9fe9f1fb0adba5c2c7e1517d126ad5dec1/podman_compose.py#L2125-L2126

#### Proposed Solution:
Remove the two lines linked above. According to the git blame they were added in order to show more helpful information in the systemd command logs. However I feel like this side effect does more harm than good. Debugging whether or not the actual compose file works should not be part of generating a systemd unit. Plus, the journalctl logs and systemctl status can always be used to debug any issues.

#### Possible Alternatives:
I currently simply skip the full command due to its side effects. Instead I manually create the `.env` file that this command would otherwise write to.

An alternative to removing the lines completely would be to add a command line flag.

winston-yallow:
Closes #922 by removing the side effect from the systemd register action.

## Contributor Checklist:
- this PR is outside of the scope of the compose spec as it is systemd related.
- no unit tests have been changes since the code that was removed wasn't tested before.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2122,8 +2122,6 @@ async def compose_systemd(compose, args):
                     f.write(f"{k}={v}\n")
         log.debug("writing [%s]: done.", fn)
         log.info("\n\ncreating the pod without starting it: ...\n\n")
-        process = await asyncio.create_subprocess_exec(script, ["up", "--no-start"])
-        log.info("\nfinal exit code is %d", process)
         username = getpass.getuser()
         print(
             f"""


winston-yallow:
It seems like the DCO bot is not happy with my "Signed-off-by" line since I used the anonymized github mail there, which works for signing commits (you can see it is verified) but apparently does not satisfy the bot.

p12tic (approved):

p12tic:
Looks like I can override DCO bot, which is reasonable in this case.

winston-yallow:
Already figured it out, my local git client did the signing wrong. Edited it manually, sorry for the hassle :sweat_smile:

p12tic:
Ah, now it's signed as expected by the bot, thanks.

winston-yallow:
thanks for the fast merge!

PR #928, opened 2024-05-06
add await for create_pods call

schugabe:
fixes log output podman_compose.py:2527: RuntimeWarning: coroutine 'create_pods' was never awaited


p12tic:
Thanks!

Issue #795, opened 2023-10-19
Defaulting to --pod on breaks compose file with --userns


deliciouslytyped:
https://github.com/containers/podman-compose/commit/7f5ce26b1b8103f48d5702dce7d52afe8b76daa5 breaks one of my compose files because `Error: --userns and --pod cannot be set together`.  Passing --in-pod=False or something, would presumably fix this, but 

Is there a way to disable --pod in the yaml? If not, can there be a way added to do it from here on out, or can this somehow be fixed?

I tried looking at the source but couldn't figure out if there is an existing way to do it.

MrSSvard:
I think the check needs to be updated too. The only way to have it not be in a pod that I figured out is --in-pod ""
This worked before when the default was False as any "--in-pod" argument meant you wanted a pod. 

A completely new argument probably needs to be created unless we either revert the default or occupy a special pod name "False" that means no pod will be used.

uosis:
`podman compose --in-pod= up` works, but this change of behavior is not great. Why is it desirable to default to running in pod? It seems like a needless deviation from `docker compose` behavior.

camporter:
Same issue here, strange that the default behavior is to break userns options, which is incredibly necessary when dealing with a mix of rootful and rootless containers under podman.

schugabe:
Same issue here that change and breaks pretty much all my projects.

mokibit:
@deliciouslytyped  thanks for opening an issue.
The issue was fixed in https://github.com/containers/podman-compose/pull/964 by adding a custom section in the compose file so the user can set the preferred value for `--in-pod`.
@p12tic the problem can be closed.

Ep0chalypse:
Using x-podman in-pod: false does not seem to work on latest version of podman-compose
```
podman-compose --version
podman-compose version 1.4.0  
podman version 5.2.2
```

I put this in my compose file
```
x-podman:
  in-pod: false
*SNIP*
userns_mode: "keep-id"
```

When I go to start the containers it says:
```
podman-compose up
1e91fa55041e3e5089c7d0d663a72d561e521a657ef3e77999e1eb0ce98828d1
Error: --userns and --pod cannot be set together
Error: --userns and --pod cannot be set together
[mariadb]    | Error: no container with name or ID "mariadb" found: no such container
```

p12tic:
@Ep0chalypse could you copy your comment to a new bug?

guesant:
[...]

PR #932, opened 2024-05-07
Allow providing default value for `--in-pod=` using environment variable

uosis:
With the new default of true, it is quite unergonomic and error prone to have to pass `--in-pod=` everywhere. This simple change allows providing default value using environment variable.

This should help mitigate https://github.com/containers/podman-compose/issues/795.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1943,7 +1943,7 @@ class PodmanCompose:
             help="pod creation",
             metavar="in_pod",
             type=str,
-            default="true",
+            default=os.getenv("PODMAN_COMPOSE_USE_PODS", "true"),
         )
         parser.add_argument(
             "--pod-args",


ggrandou:
Instead of using an environment variable, how about adding a custom section in the compose file ? like:
```
services:
  my_app:
    userns_mode: keep-id
x-podman:
  in-pod: false
```

p12tic:
> Instead of using an environment variable, how about adding a custom section in the compose file ? like:

Yes, I though the same. We should reduce the amount of implicit state.

mokibit:
@uosis  thanks for opening an issue.
The issue was fixed in  https://github.com/containers/podman-compose/pull/964 by adding a custom section in the compose file as @ggrandou  suggested.
@p12tic the problem can be closed.

uosis:
Sounds good, thanks for a proper fix!

PR #941, opened 2024-05-18
Update podman_compose.py

HernandoR:
fix #940

## Contributor Checklist:

All changes require additional unit tests.

PR #943, opened 2024-05-21
Fix the test_include for multi subcomposes

HernandoR:
## Contributor Checklist:

correct the fail of `test_include` introduced in #940

PR #944, opened 2024-05-21
Revert "Update podman_compose.py"

p12tic:
Reverts containers/podman-compose#941

The PR broke tests.


p12tic:
Test fixed in https://github.com/containers/podman-compose/pull/943.

PR #949, opened 2024-05-26
Add support for multi-line environment files

mokibit:
This PR adds support for several multi-line environment files.
PR fixes issue https://github.com/containers/podman-compose/issues/908.
Credit for actual solution goes to @hedayat. I only added tests as code reviewer @p12tic  requested.

From https://github.com/containers/podman-compose/pull/909:
"TBH, I don't find the part of compose spec which specifies this. Actually, it might not be according to the spec. However, this is what docker compose supports... !"

### DIFF
--- a/pytests/test_container_to_args.py
+++ b/pytests/test_container_to_args.py
@@ -352,6 +352,15 @@ class TestContainerToArgs(unittest.IsolatedAsyncioTestCase):
             ],
         )
 
+    async def test_env_file_obj_required_non_existent_path(self):
+        c = create_compose_mock()
+
+        cnt = get_minimal_container()
+        cnt['env_file'] = {'path': 'not-exists', 'required': True}
+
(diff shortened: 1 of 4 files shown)


pytests/test_container_to_args.py
```diff
@@ -302,14 +340,27 @@ async def test_env_file_obj_required(self):
...
         )
 
+    async def test_env_file_obj_required_non_existent_path(self):
+        c = create_compose_mock()
+
+        cnt = get_minimal_container()
+        cnt['env_file'] = {'path': 'not-exists', 'required': True}
+
+        with self.assertRaises(ValueError):
+          await container_to_args(c, cnt)
```
p12tic, line 362: Bad indentation

hedayat:
@mokibit 
Thanks for adding the tests, it could take me a while before I come back to the PR. :)

PR #950, opened 2024-05-26
Test if `Check for spelling errors` fail without changes

mokibit:
This is a draft PR only to check if the test `Check for spelling errors` fails without real changes.
@p12tic  please check why tests fail.


p12tic:
Seems like test changed under our feet..

Issue #747, opened 2023-08-23
Not able to implement health-on-failure via podman-compose


tim13kh:
Hello ,

Seems it is not possible to setup option '--health-on-failure' via podman-compose

`podman run --replace -d --name test-container --health-cmd /healthcheck --health-on-failure=kill --health-retries=1 health-check-actions`

BlaiseLebreton:
I think it's mainly due to the fact the [compose specificiation](https://github.com/compose-spec/compose-spec/blob/master/spec.md#healthcheck) doesn't have any way to specify an action when the container is unhealthy 

Regardless it would be helpful to have something to automatically restart containers based on their health. Either in the compose file or an external service
On Docker there's the [docker-autoheal](https://github.com/willfarrell/docker-autoheal) but it uses the docker socket

antdking:
As a workaround, I'm running podman-compose with `--podman-run-args` set:

`podman-compose --verbose --podman-run-args='--health-on-failure=restart' up --build -d`

This requires that all services have a healthcheck defined, either with a `test` command, or be explicitly disabled:
```yaml
service:
  s:
    healthcheck:
      disable: true
```

hunter86bg:
I've proposed a [PR#955](https://github.com/containers/podman-compose/pull/955) that should allow to specify that.

PR #925, opened 2024-05-03
Fixes #897: support `env_file` as objects

GerkinDev:
Signed-off-by: Alexandre Germain <alexandre.j.p.germain@gmail.com>

PR #955, opened 2024-05-31
Add --health-on-failure, fixes containers/podman-compose#747

hunter86bg:
## Contributor Checklist:

This PR adds the option to stop/restart container when healtcheck fails.


hunter86bg:
@p12tic , what do you think about this one ?

p12tic:
@hunter86bg As I understand, this is not part of docker-compose specification. Accordingly, `on_failure` key should be `x-podman.on_failure` and also it should be documented in https://github.com/containers/podman-compose/blob/main/docs/Extensions.md. Also please add unit tests. There are a lot of examples how to write tests for `container_to_args` in repository history. For example, see pytests/ part of https://github.com/containers/podman-compose/pull/925.

Issue #935, opened 2024-05-10
Unable to start containers with userns on v1.1.0


fubarhouse:
**Describe the bug**

After upgrading to `v1.1.0`, I am unable to start the container stack which has been forcing me to revert to `v1.0.6` for a couple of weeks. This change has occurred independently of Podman upgrades - and it may be that I need to add additional configuration but at this time I do not how to get around it.

I should not that I am not using Podman with Kubernetes right now as I suspect that is related.

I have found an approachable way to replicate the issue, and the error does not occur without the `userns` attribute set.

**To Reproduce**

1. Observe the minimalist busybox `docker-compose.yml` file below.
2. Add it to your file system in an empty directory
3. Run `podman-compose up -d`
4. Observe the error

**docker-compose.yml**

```yamlversion: "2"
services:
    frontend:
      userns_mode: keep-id:uid=1000
      image: busybox
      command: ["/bin/busybox", "httpd", "-f", "-p", "8080"]
      working_dir: /
```

**Expected behavior**

I would expect the container to start, or fail differently. It seems there is something wrong with the configuration which works in the previous release of podman-compose (`v1.0.6`)

**Actual behavior**

Error observed:

```
--userns and --pod cannot be set together
```

**Output**

```
# Fresh git install
...
Cloning into 'podman-compose'...
done.
==> Starting pkgver()...
==> Updated version: podman-compose-git 1:1.1.0.r31.7a2da76-1
==> Sources are ready.
==> Making package: podman-compose-git 1:1.1.0.r31.7a2da76-1 (Fri May 10 21:14:23 2024)
...

$ podman-compose version
podman-compose version 1.1.0
podman version 5.0.2

$ podman-compose up -d
9aa63fcc8cce556c70ea97cd7acdc157f8915efb912f519e0891af50380fedbd
Error: --userns and --pod cannot be set together
Error: no container with name or ID "podmanprojects_frontend_1" found: no such container
```

**Output on v1.0.6**

```
$ wget -O ./podmancompose-1.0.6 https://raw.githubusercontent.com/containers/podman-compose/v1.0.6/podman_compose.py
--2024-05-10 21:40:28--  https://raw.githubusercontent.com/containers/podman-compose/v1.0.6/podman_compose.py
Loaded CA certificate '/etc/ssl/certs/ca-certificates.crt'
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.111.133, 185.199.110.133, 185.199.108.133, ...
[...]

Issue #673, opened 2023-04-12
pod creation disabled by default


mawe42:
Since commit f6dbce36181c44d0d08b6f4ca166508542875ce1, podman-compose has switched the default behaviour for creating a pod. Before this change, there was a `--no-pod` option with default `False`. Now there is a `--in-pod` option with default `False`.

If this change was intentional, then it might be good to advertise / document it somewhere IMHO. But I would think having `--in-pod` default to `True` to keep backwards compatible would be very good for existing users (including me...)

muayyad-alsadi:
The last stable 1.0.3 and current stable 1.0.6 have same behavior which is no pod by default 

The next stable would have pod creation enabled by default. I believe pod creation is good default.

Any suggestion?

mawe42:
Ah I see, thanks! Yes, we - as many others - have been using the devel branch because the stable 1.0.3 release wasn't working for us. I would definitely vote for pod creation by default.

HerrFrutti:
I don't understand this... I've been reading though the issues and there are suggested solutions that are not working (anymore?). 
 
For example this is not working (for me):
> @jayache80 you can no choose to not create pod or create pod with whatever arguments you want
> 
> [#442 (comment)](https://github.com/containers/podman-compose/issues/442#issuecomment-1062067920)
> 
> ```
> podman-compose --no-pod up -d
> # or
> podman-compose --pod-args='--infra=false --share=""' up -d
> # or
> podman-compose --pod-args='--infra=true --share=""' up -d
> ```

I'm new to podman and podman-compose and want that my compose file creates a pod. Nothing I've found helped me so far.

Running on stable, installed trough dnf install.

```
using podman version: 4.4.4
podman-composer version  1.0.3
podman --version
podman version 4.4.4
```

Sorry if this is the wrong place, but I did not want to open another issue, with the same problem.

muayyad-alsadi:
> podman-composer version  1.0.3

we have just released a newer stable version 1.0.6, please update to that version
using 

```
sudo pip install --upgrade podman-compose
```

you can use `--user` if you don't want to use `sudo`

in the latest stable version

```
podman-compose --in-pod=0 up -d # .... if you don't want pod  # this is the the current 1.0.x default
podman-compose --in-pod=1 up -d # .... if you want pod
```

HerrFrutti:
that does the trick. Thank you.

When running `podman-compose down` the containers get removed, but the pod remains. Is that on purpose?

[...]

Issue #166, opened 2020-05-04
Does podman-compose have a way to pass the "--userns keep-id" argument?


coreyryanhanson:
If I add a line to a yaml to be read by podman-compose with "userns_mode: keep-id", it ignores it completely. I've tried a few variations like "userns: keep-id" (matching the podman command line argument), but every time the code generated to call the containers does not change. Is this argument not supported yet in podman-compose?

pm98zz-c:
You should be able to it with:
```
    environment:
      PODMAN_USERNS: keep-id
```
As per http://docs.podman.io/en/latest/markdown/podman-create.1.html for –userns: "It defaults to the PODMAN_USERNS environment variable".

pm98zz-c:
Only workaround I found is to set it in .bashrc/.zshrc:
```
export PODMAN_USERNS=keep-id
```
As per http://docs.podman.io/en/latest/markdown/podman-create.1.html for –userns, "It defaults to the PODMAN_USERNS environment variable."


_That said, I then had other perms issues creating the container (container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:59: mounting "sysfs" to rootfs at "/sys" caused: operation not permitted: OCI runtime permission denied error), but that's off topic._

muayyad-alsadi:
you can pass it using `userns_mode: keep-id` in the YAML

https://github.com/containers/podman-compose/blob/105c27c8dc902a72fcf1cd6784159b43fed6f808/podman_compose.py#L866

here is an example

https://github.com/containers/podman-compose/blob/devel/examples/nodeproj/docker-compose.yml#L17
https://github.com/containers/podman-compose/tree/devel/examples/nodeproj

the example out of the box passes UID from .env not keep-id

you can also put `PODMAN_USERNS=keep-id` in your `.env` because all `PODMAN_*` are passed

miquecg:
I've just tried adding that line to a `docker-compose.yml` but it doesn't work for me, as @coreyryanhanson pointed out. podman-compose generates a `podman create` command which doesn't include the flag `--userns=keep-id`, so the only way to change this seems to be through the environment.

`PODMAN_USERNS=keep-id podman-compose up database` works perfectly.

josharrington:
The `userns_mode` config in podman-compose has not yet been released, it was added after the latest 1.0.3 release.

jasalt:
[...]

PR #964, opened 2024-06-14
Allow providing custom in_pod argument as a global compose file variable

mokibit:
Default command line argument `in_pod` was set to True, but this breaks the compose file for users who want to use `--userns` argument. 
This PR sets default `in_pod` value to None, and later resolves whether to create a pod by checking compose file, as new argument in compose file x-podman is now available (documentation is updated and tests added). 
Now it is convenient for users to pass custom `in_pod` value (True or False) as a compose file argument when command line value of `in_pod` is not provided.
This PR solves issues: https://github.com/containers/podman-compose/issues/795, https://github.com/containers/podman-compose/issues/935, https://github.com/containers/podman-compose/issues/673, https://github.com/containers/podman-compose/issues/695, https://github.com/containers/podman-compose/issues/166.

### DIFF
--- a/docs/Extensions.md
+++ b/docs/Extensions.md
@@ -89,3 +89,23 @@ In addition, podman-compose supports the following podman-specific values for `n
 
 The options to the network modes are passed to the `--network` option of the `podman create` command
 as-is.
+
+
+## Custom pods management
+
+Podman-compose can have containers in pods. This can be controlled by extension key x-podman in_pod.
+It allows providing custom value for --in-pod and is especially relevant when --userns has to be set.
+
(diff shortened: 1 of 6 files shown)


deliciouslytyped:
I forgot how any of this works but shouldn't this be a per-container setting?

p12tic:
> I forgot how any of this works but shouldn't this be a per-container setting?

Pod is grouping of containers. At least currently all services in compose file are put in a single pod if this is enabled. Is there use case when some containers are put in a pod and the rest aren't?

p12tic (approved):

Issue #269, opened 2021-01-25
Read compose file from URL and/or stdin


rvansa:
For demo purposes (blog...) it would be useful if I could run
```
podman-compose -f https://my.domain/docker-compose.yaml
```
or at least
```
curl -s https://my.domain/docker-compose.yaml | podman-compose -f -
```
Neither of that works with my  `podman-composer version  0.1.6dev` (apologies if this is already fixed in a more recent version).

tobiaseisenschenk:
Also an issue for me. The same command breaks once I set `PODMAN_COMPOSE_PROVIDER=podman-compose`

ArthoPacini:
We really need this functionality! How could this be implemented?

ArthoPacini:
is this the place it needs to get as an stdin?

`@staticmethod
    def _init_global_parser(parser):
    #...
    
    parser.add_argument(
            "-f",
            "--file",
            help="Specify an alternate compose file (default: docker-compose.yml)",
            metavar="file",
            action="append",
            default=[],
        )
   `

ArthoPacini:
I've modified it, its not pending on pull request:

https://github.com/containers/podman-compose/pull/966

mokibit:
This issue has been resolved by (https://github.com/containers/podman-compose/pull/966).
Minimal reproduction:
```
Debian GNU/Linux 13 (trixie)
Python 3.13.5
podman-compose version 1.6.0
podman version 5.4.2
```
Create a `compose.yml`:
```
services:
  hello:
    image: busybox
```
Pipe it to `podman-compose`:
```
$ cat compose.yml | podman-compose -f - config
# success
services:
  hello:
    image: busybox
```

Direct URL support (`-f https://...`) is not planned at this time. The official `docker compose` CLI also does not support reading compose files from URLs.

A simple and effective substitution already exists — download the file and pipe it via stdin:
```
curl -s https://example.com/docker-compose.yml | podman-compose -f - up
```
If there is strong community demand for native URL support in the future, a separate feature request can be opened. For now, the stdin functionality covers the primary use case described in this issue.

PR #966, opened 2024-06-16
Add ability to input docker-compose.yaml via stdin

ArthoPacini:
Simple modification to add the ability to input docker-compose.yaml via stdin, that was lacking...

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2051,7 +2051,7 @@ class PodmanCompose:
         parser.add_argument(
             "-f",
             "--file",
-            help="Specify an alternate compose file (default: docker-compose.yml). If '-' is specified, the input will be read from standard input (stdin).",
+            help="Specify an alternate compose file (default: docker-compose.yml) or '-' for reading from stdin.",
             metavar="file",
             action="append",
             default=[],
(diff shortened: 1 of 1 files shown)


ArthoPacini:
That would solve https://github.com/containers/podman-compose/issues/269#issue-793229346

p12tic:
Looks good, thanks. Could you also document this enhancement here? https://github.com/containers/podman-compose/blob/main/podman_compose.py#L2051 (e.g. `When file is "-", standard input is read` or anything better you can think of.

ArthoPacini:
The final help message is "Specify an compose file (default: docker-compose.yml) or '-' to read from stdin."

It seems that the static check didn't allowed larger messages.

p12tic:
> It seems that the static check didn't allowed larger messages.

I think it would have allowed a string broken across multiple lines. Anyway, current message is enough.

ArthoPacini:
It seems that on the `resolve_extends` function it also reads the docker-compose.yaml file using the `yaml.safe_load()` function. I don't know exactly what this function is doing, it seems to be called somewhere on the `_parse_compose_file()` function, so it might be wise to check that out and see if it needs the same kind of modification. For now, it works great I can pass docker-compose.yaml files via stdin. Keep up the good work!

PR #978, opened 2024-06-26
Restore support for alt dockerfiles (removing hardcoded "Dockerfile")

### LINKED ISSUE #929: Broken support for Containerfile
**Describe the bug**

https://github.com/containers/podman-compose/pull/686 broke support for using `Containerfile` instead of `Dockerfile` for builds. `Dockerfile` hardcoded here:

https://github.com/containers/podman-compose/pull/686/files#diff-956dd8b5765533c7e8f5bb94c8a1b4735dfcfd9cff54829f3494b87460fab50bR1297

which causes this logic to be skipped:

https://github.com/containers/podman-compose/blob/main/podman_compose.py#L2238

**To Reproduce**

Try to build any compose file that has `build: .` or similar and `Containerfile` instead of `Dockerfile`.

**Expected behavior**
`Containerfile` to be detected as before.
**Actual behavior**
`Containerfile` not detected

**Output**

```
$ podman-compose version
podman-compose version 1.1.0
podman version 5.0.2
```

**Environment:**
 - OS: Linux

### LINKED ISSUE #977: 1.1.0 (and perhaps above) broke the use of Containerfiles
**Describe the bug**
Installed podman-compose 1.1.0 (from pypi), it refused to build my image (raise OSError : Dockerfile not found in +ctx (/my/path)).
Downgraded to 1.0.6, it works like a charm.
Also, podman build works well. 

**To Reproduce**
podman-compose.yml file including a build instruction to the current folder, with a Containerfile in it. Renaming the containerfile Dockerfile makes it work, but defeats the purpose of the generic name. Then run "podman compose up -d".

**Expected behavior**
Successful build.

**Actual behavior**
OSError: Dockerfile not found in /my/path

**Output**
If needed, I can edit this issue at a later point, I do not have the output right now due to the emergency of deployment. The error was occuring in site-packages/podman-compose.py somewhere in the lines 2400-2800, I can't remember. It displayed a full python stacktrace.

**Additional context**
Debian Linux 12, Podman 5.1.1 installed from Homebrew, podman-compose 1.1.0 installed from pypi.

PlugNPush:
Resolves #977 
Resolves  #929 
This commit restores setting a build directory as a string and looking for all available dockerfile alt names.
I strongly believe this is a bugfix, not modifying a compatibilty feature, as this was working that way in the older version 1.0.6.

The code for checking all the dockerfile alt names is already there, but "Dockerfile" was hardcoded in the normalize function, instead of being set to None to allow further alt dockerfile names search later in build_one().

build_one() already supports dockerfile being set to None (and that is visibly supposed to be that way).

With this PR, having a podman-compose.yaml with:
`build: .`
No longer throws "No Dockerfile found in CTX" when using a Containerfile (or other alt name).

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1527,12 +1527,8 @@ def normalize_service_final(service: dict, project_dir: str) -> dict:
         build = service["build"]
         context = build if is_str(build) else build.get("context", ".")
         context = os.path.normpath(os.path.join(project_dir, context))
-        dockerfile = (
-            "Dockerfile" if is_str(build) else service["build"].get("dockerfile", "Dockerfile")
-        )
         if not is_dict(service["build"]):
             service["build"] = {}
-        service["build"]["dockerfile"] = dockerfile
         service["build"]["context"] = context
     return service
 
--- a/newsfragments/978.bugfix
+++ b/newsfragments/978.bugfix
@@ -1,0 +1,1 @@
+Fixed support for de-facto alternative `Dockerfile` names (e.g. `Containerfile`)
--- a/tests/unit/test_normalize_final_build.py
+++ b/tests/unit/test_normalize_final_build.py
@@ -22,7 +22,7 @@ class TestNormalizeFinalBuild(unittest.TestCase):
         (
             {"build": "."},
             {
-                "build": {"context": cwd, "dockerfile": "Dockerfile"},
+                "build": {"context": cwd},
             },
         ),
         (
@@ -30,7 +30,6 @@ class TestNormalizeFinalBuild(unittest.TestCase):
             {
                 "build": {
                     "context": os.path.normpath(os.path.join(cwd, "../relative")),
-                    "dockerfile": "Dockerfile",
                 },
             },
         ),
@@ -39,7 +38,6 @@ class TestNormalizeFinalBuild(unittest.TestCase):
             {
                 "build": {
                     "context": os.path.normpath(os.path.join(cwd, "./relative")),
-                    "dockerfile": "Dockerfile",
                 },
             },
         ),
@@ -48,7 +46,6 @@ class TestNormalizeFinalBuild(unittest.TestCase):
             {
                 "build": {
                     "context": "/workspace/absolute",
-                    "dockerfile": "Dockerfile",
                 },
             },
         ),
@@ -74,7 +71,6 @@ class TestNormalizeFinalBuild(unittest.TestCase):
             {
                 "build": {
                     "context": cwd,
-                    "dockerfile": "Dockerfile",
                 },
             },
         ),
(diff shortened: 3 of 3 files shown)


PlugNPush:
Currently modifying the tests, as we now expect None when no dockerfile is provided.

podman_compose.py
```diff
@@ -1528,7 +1528,7 @@ def normalize_service_final(service: dict, project_dir: str) -> dict:
         context = build if is_str(build) else build.get("context", ".")
         context = os.path.normpath(os.path.join(project_dir, context))
         dockerfile = (
-            "Dockerfile" if is_str(build) else service["build"].get("dockerfile", "Dockerfile")
+            None if is_str(build) else service["build"].get("dockerfile", None)
         )
         if not is_dict(service["build"]):
             service["build"] = {}
```
larsks, line 1534: This works, but aren't we doing unnecessary work? It seems we would get to the same place if we just removed the `dockerfile` logic completely, like this:

```
def normalize_service_final(service: dict, project_dir: str) -> dict:
    if "build" in service:
        build = service["build"]
        context = build if is_str(build) else build.get("context", ".")
        context = os.path.normpath(os.path.join(project_dir, context))
        if not is_dict(service["build"]):
            service["build"] = {}
        service["build"]["context"] = context
    return service
```

This will not add a `dockerfile` key if one did not already exist, and that will allow the logic in `build_one` to find the correct file.

PlugNPush: This would prevent setting a custom dockerfile like this:
`build:
  context: .
  dockerfile: webapp.Dockerfile`

[spec](https://docs.docker.com/compose/compose-file/build/)

larsks: No it doesn't, because of:

```
build = service["build"]
```

If `service['build']` was a dictionary at the start of the function, it will still be one at the end of the function.

PlugNPush: Yes you are right, I read the code a bit too fast. I'm removing it now then.

PlugNPush: The tests are expecting a dockerfile to be present in every case, they should also be updated to reflect this change.

p12tic:
Thanks a lot for the PR. It looks good from the first sight, didn't look in depth yet.

We will need tests that cover all 4 cases: 
 - `build: .`
 - build dictionary but not dockerfile
 - build dictionary with `dockerfile: Dockerfile`
 - build dictionary with `dockerfile: Dockerfile.custom` 

Therefore it's not appropriate just to change existing tests, we will need additional cases. Hopefully this is mostly copy-paste and can be done without much effort.

Thanks again!

p12tic:
> All these cases are already included in test_normalize_final_build.py, have a look at the changes :)
They used to expect "dockerfile": "Dockerfile" in every test case, which is incorrect when trying to support alt dockerfile names.

You're right. Turns out that the caveat of `didn't look in depth yet` was worth writing.

p12tic (changes requested):
Looks great, thanks!

Could you please add a newsfragment with a description of what this fixes? This is just one line file in newsfragments directory, see newsfragments/README.md for an explanation.

Also, could you combine the commits into one? Whole PR is semantically one change with tests, so there's no need to split it across multiple commits to show how the code evolved until it arrived to correct conclusion.

PlugNPush:
Doing this now!

PlugNPush:
All done! @p12tic

p12tic (approved):

p12tic:
Thanks!

Issue #972, opened 2024-06-24
Secrets handling needs more unit tests


p12tic:
get_secret_args() is almost completely not tested except for the few tests added in https://github.com/containers/podman-compose/pull/971.

mokibit:
Fixed in https://github.com/containers/podman-compose/pull/981.

p12tic:
Thanks

PR #981, opened 2024-06-30
unittests: Added unit tests for podman-compose secrets

mokibit:
This PR adds unit tests for secret handling in podman_compose.get_secret_args().
PR solves issue https://github.com/containers/podman-compose/issues/972.

### DIFF
--- a/tests/unit/test_container_to_args_secrets.py
+++ b/tests/unit/test_container_to_args_secrets.py
@@ -1,11 +1,18 @@
 # SPDX-License-Identifier: GPL-2.0
 
+import os
 import unittest
+
+from parameterized import parameterized
 
 from podman_compose import container_to_args
 
 from .test_container_to_args import create_compose_mock
 from .test_container_to_args import get_minimal_container
+
+
+def repo_root():
+    return os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 
 
 class TestContainerToArgsSecrets(unittest.IsolatedAsyncioTestCase):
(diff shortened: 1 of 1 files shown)


tests/unit/test_container_to_args_secrets.py
```diff
@@ -89,3 +92,317 @@ async def test_pass_secret_as_env_variable_no_external(self):
...
+                "uid": "103",
+                "gid": "103",
+                "mode": "400",
+            }
+        ]
+
+        with self.assertLogs() as cm:
+            args = await container_to_args(c, cnt)
+        self.assertEqual(len(cm.output), 1)
+        self.assertEqual('That is un-supported and a no-op and is ignored.' in cm.output[0], True)
```
p12tic, line 183: assertTrue

tests/unit/test_container_to_args_secrets.py
```diff
@@ -89,3 +92,317 @@ async def test_pass_secret_as_env_variable_no_external(self):
...
+    async def test_file_secret(
+        self, test_name, declared_secrets, add_to_minimal_container, volume_ref
+    ):
+        c = create_compose_mock()
+        c.declared_secrets = declared_secrets
+        cnt = get_minimal_container()
+        cnt["_service"] = "test-service"
+        cnt["secrets"] = [add_to_minimal_container]
+        args = await container_to_args(c, cnt)
+        base_dir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
```
p12tic, line 356: Extract to separate function because it's likely to have repeat uses (at least that's the case in other tests).

tests/unit/test_container_to_args_secrets.py
```diff
@@ -89,3 +92,317 @@ async def test_pass_secret_as_env_variable_no_external(self):
...
+    @parameterized.expand([
+        (
+            "no_secret_target",
+            {
+                "file_secret": {
+                    "file": "./my_secret",
+                }
+            },
+            "file_secret",
+            "/test_dirname/my_secret:/run/secrets/file_secret:ro,rprivate,rbind",
```
p12tic, line 306: I think we should calculate the final path here, otherwise it's surprising to the reader that the expected value is going to be modified in the test.

tests/unit/test_container_to_args_secrets.py
```diff
@@ -89,3 +92,317 @@ async def test_pass_secret_as_env_variable_no_external(self):
...
+            },
+            {
+                "source": "file_secret",
+                "target": "/etc/custom_location",
+            },
+            "/test_dirname/my_secret:/etc/custom_location:ro,rprivate,rbind",
+        ),
+    ])
+    async def test_file_secret(
+        self, test_name, declared_secrets, add_to_minimal_container, volume_ref
```
p12tic, line 348: Last argument should be something like expected_volume_ref

p12tic (approved):

PR #994, opened 2024-07-07
Pass build description labels to podman build

swick:
The Compose Specification supports adding labels to the build image which is also used in practice. Support this and pass the labels to `podman build`.

https://docs.docker.com/compose/compose-file/build/#labels

One of the integration tests suggests that this is a known limitation but there is no indication anywhere as to why.

### DIFF
--- a/newsfragments/build-labels.feature
+++ b/newsfragments/build-labels.feature
@@ -1,0 +1,1 @@
+Added support for build labels.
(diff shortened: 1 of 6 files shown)


p12tic (changes requested):
Looks good, thanks.

Please add a release note to newsfragments directory (see README there for documentation and existing release notes snippets for examples).

Also, please add a test for this feature. https://github.com/containers/podman-compose/commit/3296c8d34f4f8f774d6d857e5e2ff3fe91dd24b3 is a good example. I think you can just take the output of `podman inspect` and do some `assertIn`s to verify that the label has been passed correctly.

The integration tests are a bit of a mess right now, so putting your new test in a separate directory such as tests/integration/build_labels would be a fine solution. I or someone else will categorize and move tests around later when we have time.

swick:
Added an integration test and added support for labels as dict.

Issue #983, opened 2024-07-02
All integration tests should be run automatically


p12tic:
Currently there are a lot of integration tests in tests/integration that are supposed to be run manually. I.e. call podman-compose and inspect output. This can be fully automated.

mokibit:
Working on it.

PR #997, opened 2024-07-08
Automate manual `build` tests

mokibit:
This PR automates manual `build` tests.
It is a partial fix for https://github.com/containers/podman-compose/issues/983.


requirements.txt
```diff
@@ -4,4 +4,5 @@
 
 pyyaml
 python-dotenv
+requests
```
p12tic, line 7: Should be part of test-requirements.txt

tests/integration/test_podman_compose_build.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_build(self):
+        try:
+            command_build = [
+                podman_compose_path(),
+                "-f",
+                os.path.join(compose_yaml_path(), "docker-compose.yml"),
+                "build",
+                "--no-cache",
+            ]
+            self.run_subprocess_assert_returncode(command_build)
```
p12tic, line 29: Nit - command could be put directly into arguments. Same below.

tests/integration/test_podman_compose_build.py
```diff
@@ -0,0 +1,65 @@
...
+            self.run_subprocess_assert_returncode(command_up)
+
+            request = requests.get('http://localhost:8080/index.txt')
+            self.assertEqual(request.status_code, 200)
+
+            # FIXME: suspicious behaviour, too often ends up in error
+            alt_request = requests.get('http://localhost:8000/index.txt')
+            self.assertEqual(alt_request.status_code, 200)
+            self.assertTrue("ALT buildno=2 port=8000 " in alt_request.text)
+        except Exception:
```
p12tic, line 47: Not sure what this dues. It will skip the rest of the block.

tests/integration/test_podman_compose_build_fail.py
```diff
@@ -0,0 +1,31 @@
...
+        command_build = [
+            podman_compose_path(),
+            "-f",
+            os.path.join(compose_yaml_path(), "docker-compose.yml"),
+            "build",
+        ]
+        output, error = self.run_subprocess_assert_returncode(
+            command_build, expected_returncode=127
+        )
+        self.assertTrue("RUN this_command_does_not_exist" in str(output))
```
p12tic, line 29: assertIn should be used

p12tic: output is already a string, no?

tests/integration/test_podman_compose_build.py
p12tic: Ah no, it will execute the inspect check only when exception happens.

tests/integration/test_podman_compose_build.py
```diff
@@ -0,0 +1,65 @@
...
+    return os.path.join(test_path(), "build")
+
+
+class TestComposeBuild(unittest.TestCase, RunSubprocessMixin):
+    def test_build(self):
+        try:
+            command_build = [
+                podman_compose_path(),
+                "-f",
+                os.path.join(compose_yaml_path(), "docker-compose.yml"),
```
p12tic, line 25: Duplicate calculation many times in the function.

PR #1004, opened 2024-07-19
Substitution with service environment

lisongmin:
This pr add the ability to substitute the variable with the service's environment.
This allows the use the service's environment within the service block itself.

For instance, in the provided example:

```yaml
services:
  app:
    image: "abc:${version}"
    build:
      context: .
      args:
        VERSION: "${version}"
    environment:
      version: v1.0
```

The `version` environment is defined within the `app` service and
it is then used to define the `image` and `VERSION` build argument
for the same service.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -264,6 +264,16 @@ def rec_subs(value, subs_dict):
     do bash-like substitution in value and if list of dictionary do that recursively
     """
     if is_dict(value):
+        if 'environment' in value and is_dict(value['environment']):
+            # Load service's environment variables
+            subs_dict = subs_dict.copy()
+            svc_envs = {k: v for k, v in value['environment'].items() if k not in subs_dict}
+            # we need to add `svc_envs` to the `subs_dict` so that it can evaluate the
+            # service environment that reference to another service environment.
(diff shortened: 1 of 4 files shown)


podman_compose.py
```diff
@@ -264,6 +264,14 @@ def rec_subs(value, subs_dict):
     do bash-like substitution in value and if list of dictionary do that recursively
     """
     if is_dict(value):
+        if 'environment' in value and is_dict(value['environment']):
+            # Load service's environment variables
+            subs_dict = subs_dict.copy()
+            svc_envs = {k: v for k, v in value['environment'].items() if k not in subs_dict}
+            subs_dict.update(svc_envs)
```
p12tic, line 271: I think it's enough to do a dict update two lines below. rec_subs won't delete any of the entries in `svc_envs`, so everything what this `update` does will be overwritten to lines below.

If I'm wrong, then the reason needs to be added as a comment because others will have the same question when reading this code.

p12tic (changes requested):
Looks great, thanks. I had one comment, but aside from that the PR can go in.

p12tic:
Rebased.

p12tic (approved):

PR #1009, opened 2024-07-22
podman_compose.py: support "platform" property in the build command

bkueng:
This was already added to container_to_args() in https://github.com/containers/podman-compose/pull/470 which is used for the 'up' and 'run' commands.

Definition in the schema: https://github.com/compose-spec/compose-spec/blob/main/schema/compose-spec.json#L329

Should I add a unit test? Are there any for the build command already?

### DIFF
--- a/newsfragments/build-platform.feature
+++ b/newsfragments/build-platform.feature
@@ -1,0 +1,1 @@
+Added support for "platform" property in the build command.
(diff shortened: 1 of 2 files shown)


p12tic:
Rebased and added newsfragment.

There are no unit tests for build command and I will not ask to create such tests for now.

p12tic (approved):
Thanks for adding this feature.

Issue #1013, opened 2024-07-30
Support changing `io.podman` label domain namespace


legobeat:
**Is your feature request related to a problem? Please describe.**

It is possible that we have two different processes where one needs to use `docker-compose` and the other `podman-compose` but interacting with the same container socket, containers, and compose project. For example, maybe certain configuration is only available in one or the other and no single implementation covers the entire use-case.

Unfortunately, containers started by one is not visible from the other.

This is due to project-matching being done by matching the label `io.podman.compose.project`. Containers started by docker-compose instead carry the `io.podman.compose.project`. It seems to be no way to make them work together due to this.

**Describe the solution you'd like**

Introduce a new environment variable like `COMPOSE_LABEL_NAMESPACE` defaulting to `io.podman`. Users can then interact with projects managed by docker-compose via `COMPOSE_LABEL_NAMESPACE=com.docker`, or set their own namespace.

**Describe alternatives you've considered**
Transparently filter for both currently known namespaces. This seems to not be worth it.

**Additional context**

PR #1015, opened 2024-07-31
Support changing io.podman label domain namespace

legobeat:
Add global cli flag `--label-domain` and env var `COMPOSE_LABEL_DOMAIN` (default `io.podman`).
If set, it changes the out-of-spec root domain set and filtered for alongside the spec-mandated `com.docker.compose`.

### Related:
- Resolves https://github.com/containers/podman-compose/issues/1013
- Based on https://github.com/containers/podman-compose/pull/1014

#### Compose spec
  - https://github.com/compose-spec/compose-spec/blob/main/05-services.md#labels
  - https://github.com/compose-spec/compose-spec/blob/main/02-model.md#the-compose-application-model

### DIFF
--- a/newsfragments/label_domain.feature
+++ b/newsfragments/label_domain.feature
@@ -1,0 +1,1 @@
+Added support to change default resource label prefix from io.podman via --label-domain cli arg
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -401,7 +401,7 @@ async def assert_volume(compose, mount_dict):
     log.debug("podman volume inspect %s || podman volume create %s", vol_name, vol_name)
     # TODO: might move to using "volume list"
     # podman volume list --format '{{.Name}}\t{{.MountPoint}}' \
-    #     -f 'label=io.podman.compose.project=HERE'
+    #     -f f"label={compose.label_domain}.compose.project=HERE"
     try:
         _ = (await compose.podman.output([], "volume", ["inspect", vol_name])).decode("utf-8")
     except subprocess.CalledProcessError as e:
@@ -411,7 +411,7 @@ async def assert_volume(compose, mount_dict):
         args = [
             "create",
             "--label",
-            f"io.podman.compose.project={compose.project_name}",
+            f"{compose.label_domain}.compose.project={compose.project_name}",
             "--label",
             f"com.docker.compose.project={compose.project_name}",
         ]
@@ -804,11 +804,11 @@ def norm_ports(ports_in):
     return ports_out
 
 
-def get_network_create_args(net_desc, proj_name, net_name):
+def get_network_create_args(net_desc, proj_name, net_name, label_domain='io.podman'):
     args = [
         "create",
         "--label",
-        f"io.podman.compose.project={proj_name}",
+        f"{label_domain}.compose.project={proj_name}",
         "--label",
         f"com.docker.compose.project={proj_name}",
     ]
@@ -871,7 +871,9 @@ async def assert_cnt_nets(compose, cnt):
         except subprocess.CalledProcessError as e:
             if is_ext:
                 raise RuntimeError(f"External network [{net_name}] does not exists") from e
-            args = get_network_create_args(net_desc, compose.project_name, net_name)
+            args = get_network_create_args(
+                net_desc, compose.project_name, net_name, compose.label_domain
+            )
             await compose.podman.output([], "network", args)
             await compose.podman.output([], "network", ["exists", net_name])
 
@@ -1461,7 +1463,7 @@ class Podman:
                     "ls",
                     "--noheading",
                     "--filter",
-                    f"label=io.podman.compose.project={self.compose.project_name}",
+                    f"label={self.compose.label_domain}.compose.project={self.compose.project_name}",
                     "--format",
                     "{{.Name}}",
                 ],
@@ -1677,6 +1679,7 @@ class PodmanCompose:
         self.commands = {}
         self.global_args = argparse.Namespace()
         self.project_name = None
+        self.label_domain = None
         self.dirname = None
         self.pods = None
         self.containers = []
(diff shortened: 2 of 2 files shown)


p12tic:
I think this feature makes sense in principle, but instead of environment variable and command-line option I would add a customization point to the compose file. For example, we already have [default_net_name_compat](https://github.com/containers/podman-compose/blob/main/docs/Extensions.md#compatibility-of-default-network-names-between-docker-compose-and-podman-compose).

Environment variable and command line option are error prone, because a single compose file will need to have the same values passed on every invocation of podman-compose, otherwise it will break horribly.

p12tic:
A couple of nitpicks:

> The existing io.podman labels are already "undefined behavior"/"out of spec" (not in itself necessarily an issue!)

This is a conscious decision due to the fact that podman-compose will always be missing some features of docker-compose. Therefore it is impossible to claim full compatibility with docker-compose and thus podman-compose should not try to manage resources created by docker-compose.

> If users prefer editing the Compose file, this can mostly already be achieved today by configuring labels on resources explicitly - adding the new feature here is really just sugar/convenience 

I think it's not possible to set the labels with the prefix used by the compose itself. From the spec:

>> The com.docker.compose label prefix is reserved. Specifying labels with this prefix in the Compose file results in a runtime error.

Going back to whole argument I think I understand what you're saying, we are arriving to different conclusions due to different points of view.

I consider both command line arguments and the supported syntax of the compose file as the API provided by podman-compose. If we want to expose a feature to the users we should consider both command line arguments and the compose file as potential points where users could configure the feature. Flexibility, ease of use, whether API is error prone, maintainability and so on are arguments for and against a particular approach.

In this particular case, resources created using different domain namespace will be invisible to both podman-compose and docker-compose. It is easy to miss a command line argument in some invocation of podman-compose and this will result in error that is not immediately visible but may result in confusing errors later. Even `podman-compose down` has no chance to work properly. Therefore I view changing io.podman label domain namespace via command line option as error prone with high impact in terms of lost time in case of an error.

Putting the configuration of the feature into compose file as a vendor extension significantly reduces the chances of errors because all invocations of podman-compose will be guaranteed to use the same label domain namespace. Therefore this looks like a preferable way to expose the feature.

I am personally weakly against exposing this feature in general on the grounds of error-proneness. However I don't want to be a maintainer that limits a project to the limits of my personal imagination. Therefore I will accept this feature with some modifications to reduce impact that I don't like but preserving ability to do what you want. I think this should be a reasonable compromise.

Are there limitations in the approach that I proposed that would prevent your use case? Do you just want to configure `com.docker.compose` and that's it? Maybe exposing `use_canonical_label_domain: (true|false)` would be enough for your use case?

p12tic:
> I'm thinking of the Compose file as something we expect to conform to spec, whereas CLI flags is something podman-compose can change independently without considering spec-compliance in the same way?

This is where our point of view differs.

The fact is that the default io.podman domain namespace is already used. The reason why a value different from com.docker has been selected was most likely to not break docker-compose. Regardless of reasons, that's the current situation.

Whether default io.podman domain namespace is compatible with compose spec or not does not matter because we are not going to change it. Firstly, a lot of deployments rely on default io.podman domain namespace not being changed. Additionally, it is not good idea to change it right now because podman-compose is still lacking features compared to docker-compose.

You present a good argument why user may want to change the default domain namespace. I agree with it.

The way how this is going to be exposed is completely up to podman-compose. Whether this is exposed as vendor setting in compose file or via command line option does not affect compatibility to compose spec in practice because out of the box podman-compose is incompatible anyway.

I agree that in theory it is a bit nicer to be able to specify a custom command line option to podman-compose and make it use the same domain namespace as docker-compose. However given practical considerations I think better idea is to have a setting in compose file instead.

Finally, the reason why compose exists at all was that it was inconvenient to setup a bunch of arguments to docker command when creating containers. Specifying them declaratively turned out way more convenient and less error prone. I think adding new setting keys to compose file (especially important ones such as domain namespace) is in the spirit of compose ecosystem as a whole.

> Yes, so, this PR does not touch these labels, as it is it does not allow changing/removing/overriding these in any new way. What it concerns itself with is the io.podman counterparts.

Here I was responding to "If users prefer editing the Compose file, this can mostly already be achieved today by configuring labels on resources explicitly". I interpreted that you were suggesting that users may add `com.docker.compose.project` and similar labels. In any case, I think this is peripheral to the discussion.

> Does any of this change the picture a bit? :)

Unfortunately no, sorry. Having said that, it was a very useful discussion.

I do wonder though, does my proposal not work in your use case for some reason? If it works then perhaps the best path forward is just to implement that solution, get it merged and move on. Compatibility of default domain namespace will become more important once podman-compose has relative feature parity with docker-compose.

p12tic:
> I would hope that this still reflects current intentions: for podman-compose to be an implementation of Compose Spec, and that any violations/deviations of the spec are to be treated as bugs?

This is true with a temporary exception of default domain namespace. That bug will fixed once podman-compose is relatively compatible with docker-compose.

p12tic:
By the way, one incompatibility between podman-compose and docker-compose has already been fixed by introducing a setting into the compose file. See https://github.com/containers/podman-compose/blob/main/docs/Extensions.md#compatibility-of-default-network-names-between-docker-compose-and-podman-compose

p12tic:
@legobeat I changed my mind. I think this feature makes sense assuming that users can be blamed if they use the feature incorrectly. If ability to set the domain namespace is added to compose file as well (like [default_net_name_compat](https://github.com/containers/podman-compose/blob/main/docs/Extensions.md#compatibility-of-default-network-names-between-docker-compose-and-podman-compose) then this feature can go in.

Ninos:
Any news about this MR? It's frustrating switching from docker to podman & seeing such dynamic labeling is not possible anymore... :'(

PR #1016, opened 2024-07-31
fix: improve type resolutions

legobeat:
Running [pyright]() yields many typing violations - some indicating actual issues and some due to lack of hints.

This does a minor refactor top remove a lot of the false-positive errors.

No user-facing changes in this PR. Actual bugfixes will be submitted as follow-ups after merge of this PR.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1667,11 +1667,11 @@ class PodmanCompose:
         self.environ = {}
         self.exit_code = None
         self.commands = {}
-        self.global_args = None
+        self.global_args = argparse.Namespace()
         self.project_name = None
         self.dirname = None
         self.pods = None
-        self.containers = None
+        self.containers = []
         self.vols = None
         self.networks = {}
         self.default_net = "default"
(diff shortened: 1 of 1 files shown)


podman_compose.py
```diff
@@ -1960,7 +1967,9 @@ def _parse_compose_file(self):
         container_names_by_service = {}
         self.services = services
         for service_name, service_desc in services.items():
-            replicas = try_int(service_desc.get("deploy", {}).get("replicas", "1"))
+            spec_replicas = try_int(service_desc.get("deploy", {}).get("replicas", "1"))
+            replicas = spec_replicas if spec_replicas is not None else 1
```
p12tic, line 1971: Seems like this changes behavior, doesn't it?

p12tic: I'm looking more from local perspective. In broader picture I agree that changing the behavior makes sense.

In such case we should put this in a separate commit at least and not mix with other refactoring work that doesn't change behavior and can be reviewed without considering wider implications.

By the way, wouldn't this achieve the same and be easier to understand?

`replicas = try_int(service_desc.get("deploy", {}).get("replicas"), fallback=1)`

PR #1018, opened 2024-08-03
Minimum Python version 3.9

legobeat:
Suggesting to drop advertised support for Python 3.7 and 3.8, increasing minimum supported version to 3.9.

- Python 3.7 went EOL 2023-06-27 and is already untested.
- Python 3.8 is nearing EOL (ETA 2024-10).
- 3.9 matches distro versions of:
  - Debian bullseye (oldstable)
  - At least down to Fedora 37

(I'd personally go as far as suggesting bumping as far as 3.10 or even 3.11 (Debian stable/Fedora 38) as the type annotation features there could be nice. Starting with this more conservative change to take the temperature on how conservative podman-compose intends to be wrt runtime backwards-compatibility :))

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12' ]
+        python-version: [ '3.9', '3.10', '3.11', '3.12' ]
 
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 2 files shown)


p12tic:
Agreed regarding upgrading to Python 3.9. Older distributions don't even have podman, so supporting older python is moot. If newer podman can be installed, then Python can be as well.

Regarding dropping Python 3.9 I think let's wait until it's EOL itself. The benefits of typing are small to a tiny project like podman-compose. The downsides of some commercial customer of Red Hat being unable to upgrade to get a bugfix they need are larger.

p12tic (approved):

PR #1043, opened 2024-09-21
Download and build podman-compose using docker/podman locally

iruzo:
## Reason for this PR
As a user of Fedora Silverblue I found that the easiest ways to install podman-compose are:
- Open a toolbox/distrobox container, install there the python dependencies, use pyinstaller, generate a podman-compose binary and move it outside of the container to use it.
- Create a virtual env, install dependencies and use pyinstaller to compile it.
- Install Podman Desktop and let it install and manage podman-compose.
- [Not actually podman-compose] Manually set podman.socket and downloading docker-compose.

I just wanted to execute a single command and obtain a binary I can save on my home directory without modifying anything outside my $HOME.

## Changes
- Adds Dockerfile that compiles podman-compose and let the binary on /result.
- Adds a script to generate the binary and drop it on the current directory.
- Adds script to simplify this process so people can just "download" the binary by executing a single shell command.

## Usage
```bash
sh -c "$(curl -sSL https://raw.githubusercontent.com/containers/podman-compose/main/scripts/download_podman-compose.sh)"
```

> Using this approach, everything happens within the build steps of the Dockerfile, so no dependencies or temporary files should be left on your system.

Thanks in advance. Any suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,12 @@ pip3 install https://github.com/containers/podman-compose/archive/main.tar.gz
 brew install podman-compose
 ```
 
+### Generate binary using docker/podman locally
+This script will download the repo, generate the binary using [this Dockerfile](https://github.com/containers/podman-compose/blob/main/Dockerfile.md) and let the binary in the directory where you called this script.
+```bash
+sh -c "$(curl -sSL https://raw.githubusercontent.com/containers/podman-compose/main/scripts/download_and_build_podman-compose.sh)"
+```
+
 ### Manual
 
 ```bash
--- a/scripts/generate_binary_using_dockerfile.sh
+++ b/scripts/generate_binary_using_dockerfile.sh
@@ -1,0 +1,57 @@
+#!/bin/sh
+
+# Find an available container tool (docker or podman)
+find_container_tool() {
+    if command -v docker > /dev/null 2>&1; then
+        echo "sudo docker"
+    elif command -v podman > /dev/null 2>&1; then
+        echo "podman"
+    else
+        echo "Error: Neither docker nor podman is available." >&2
+        exit 1
+    fi
+}
+
+# Determine which container tool to use
+CONTAINER_TOOL=$(find_container_tool)
+
+# Locate the directory containing dockerfile (root)
+PROJECT_ROOT_DIR="$(cd "$(dirname "$0")" && pwd)/.."
+
+# Check SELinux status and set appropriate mount option
+check_selinux() {
+    if command -v getenforce > /dev/null 2>&1; then
+        SELINUX_STATUS=$(getenforce)
+        if [ "$SELINUX_STATUS" = "Enforcing" ] || [ "$SELINUX_STATUS" = "Permissive" ]; then
+            echo ":z"
+        else
+            echo ""
+        fi
+    elif [ -f /sys/fs/selinux/enforce ]; then
+        if [ "$(cat /sys/fs/selinux/enforce)" = "1" ]; then
+            echo ":z"
+        else
+            echo ""
+        fi
+    else
+        echo ""
+    fi
+}
+
+# Get the SELinux option for volume mounts if SELinux is enforcing or permissive
+SELINUX=$(check_selinux)
+
+# Build binary
+$CONTAINER_TOOL image rm build-podman-compose
+
+if expr "$CONTAINER_TOOL" : '.*docker.*' >/dev/null; then
+    $CONTAINER_TOOL build -t build-podman-compose "$PROJECT_ROOT_DIR"
+    $CONTAINER_TOOL run --name build-podman-compose build-podman-compose
+    $CONTAINER_TOOL cp build-podman-compose:/result/podman-compose "$PROJECT_ROOT_DIR/podman-compose"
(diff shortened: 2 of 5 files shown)


p12tic:
Thanks for the PR. It will take some time to review it. Ideally I would like to have this as part of release process so that the executable is added to release artifacts.

iruzo:
Hey @p12tic,

Thank you for being so fast to reply to my PR.
I just added in a commit a `release.yml` that add the binary to the next release (in case GITHUB_TOKEN is enabled in the repo).

Please, take your time to review the PR, there is no need to rush, thanks for just taking a look at it :smiley:

README.md
```diff
@@ -74,6 +74,12 @@ pip3 install https://github.com/containers/podman-compose/archive/main.tar.gz
 brew install podman-compose
 ```
 
+### Generate binary using docker/podman locally
+This script will download the repo, generate the binary using [this Dockerfile](https://github.com/containers/podman-compose/blob/main/Dockerfile.md) and let the binary in the directory where you called this script.
+```bash
+sh -c "$(curl -sSL https://raw.githubusercontent.com/containers/podman-compose/main/scripts/download_podman-compose.sh)"
```
p12tic, line 80: I think this should better named as `build_podman-compose.sh`, because we're not just downloading. Also add a mention that the script will call sudo in case of docker-based build because this is rather sensitive to some people.

scripts/generate_binary_using_dockerfile.sh
```diff
@@ -0,0 +1,61 @@
...
+        echo "Error: Neither docker nor podman is available." >&2
+        exit 1
+    fi
+}
+
+# Determine which container tool to use
+CONTAINER_TOOL=$(find_container_tool)
+
+# Find the directory containing the dockerfile by traversing up the directory tree
+find_dockerfile_dir() {
```
p12tic, line 19: I think it would be simpler to get the path to the script using `$(dirname "$0")` and then take `../Dockerfile`. Traversing whole directory tree is not needed.

scripts/generate_binary_using_dockerfile.sh
```diff
@@ -0,0 +1,61 @@
...
+    else
+        echo ""
+    fi
+}
+
+# Get the SELinux option for volume mounts if SELinux is enforcing or permissive
+SELINUX=$(check_selinux)
+
+# Build binary
+$CONTAINER_TOOL image rm podman-compose
```
p12tic, line 59: Use something like `build-podman-compose` so that it has less chances to conflict with container names already on the system.

p12tic:
I actually forgot about this PR because I somehow thought that I'm waiting for something from you. Thanks for the ping. I have reviewed the PR now, looks good except several comments that should be simple to address.

scripts/generate_binary_using_dockerfile.sh
iruzo: You are right!, I just pushed the changes.

scripts/generate_binary_using_dockerfile.sh
iruzo: The reason for this is to be able to call the script from where ever you want without being worried to break the `podman build` command.
Thanks to this I can call the script from the download one while being outside the repository dir.

Also, if for some reason you want to move the scripts to other dir, you can be sure that they will not break.

README.md
iruzo: Hmmm, yeah, you are probably right, but since there is already a script called `generate_binary_using_dockerfile.sh`, having one with "build" could lead to confusion.

What do you think about naming it "build_and_download_podman-compose.sh" ?

p12tic: Makes sense. But maybe  download_and_build_podman-compose.sh ? So that reading left to right the sequence of events is correct.

scripts/generate_binary_using_dockerfile.sh
iruzo: If you use `$(dirname "$0")/../Dockerfile`, this will work as long as your keep the script in a subdir at the same level as it is right now.

e.g. If you change the script to `/root/scripts/docker-scripts/generate_binary_using_dockerfile.sh`, the script will break.

But yeah, you are right that it should still work if you call it from any other directory. So, if you are fine with keeping the script always under `/scripts`, I can make the change knowing that it will never break.

Since I didn't know if you will ever move the script to any other dir, I decided to do the find on the directory tree. Now that you are suggesting this, I imagine the directory structure will always be the same, so I'm going to make the change and ping you once done :smile: .

README.md
iruzo: I agree with you. I just pushed the changes :smile:

iruzo:
@p12tic 
I just pushed the change to stop doing a recursive search on the directory tree.
Also, I noticed that `docker build` does not have a `-v` parameter (oopsie, since I use podman, I didn't notice until I try it today), I changed the script so it would also work with docker.

p12tic (approved):

p12tic:
Thanks a lot for fixes, they look great.

p12tic:
I rebased to squash fixup commits into the commit they are fixing.

PR #1058, opened 2024-10-12
feat: add support for ssh property in the build command

### LINKED ISSUE #705: Support ssh key in building images
podman-compose does not seem to support the ssh key as part of the build specification for images (https://github.com/compose-spec/compose-spec/blob/master/build.md#ssh).

Supporting the ssh key should boil down to an addition to build_one along the lines of

```#!python
    for agent_or_key in build_desc.get("ssh", []):
        build_args.extend(["--ssh", agent_or_key])
```

That should extend the arguments provided to podman build to include the proper key or agent socket. I can prepare a pull request if this is interesting.

It is possible to provide keys on a global level by adding something like `--podman-build-args \\--ssh=default`, but that needs to be added explicitly and does not easily allow to use different keys or agents for different images.

banditopazzo:
Fixes #705: Add support for ssh property in the build command

Schema definition: https://github.com/compose-spec/compose-spec/blob/6f87f9328842307d33dc777ca243cf92acf7aea4/schema/compose-spec.json#L107


tests/integration/build_ssh/test_build_ssh.py
```diff
@@ -0,0 +1,226 @@
...
+                compose_path,
+                "build",
+                "test_build_ssh_map",
+                "test_build_ssh_array",
+            ])
+
+        finally:
+            # Now we send the stop command to gracefully shut down the server
+            agent.stop_agent()
+            # agent_thread.join()
```
p12tic, line 48: Seems like old code, no?

tests/integration/build_ssh/test_build_ssh.py
```diff
@@ -0,0 +1,226 @@
...
+
+            self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_path,
+                "build",
+                "test_build_ssh_map",
+                "test_build_ssh_array",
+            ])
+
```
p12tic, line 44: I think it would make sense to produce some output that shows that information was really acquired from the agent and then assert it in the test.

tests/integration/build_ssh/test_build_ssh.py
```diff
@@ -0,0 +1,226 @@
...
+        """Start the mock SSH agent and create a Unix domain socket."""
+        self.sock_path = sock_path
+        if os.path.exists(self.sock_path):
+            os.remove(self.sock_path)
+
+        self.server_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.server_sock.bind(self.sock_path)
+        self.server_sock.listen(5)
+
+        # print(f"Mock SSH agent running with socket: {self.sock_path}")
```
p12tic, line 118: Old code.

tests/integration/build_ssh/test_build_ssh.py
```diff
@@ -0,0 +1,226 @@
...
+            if self.agent_thread:
+                self.agent_thread.join()  # Wait for the thread to finish
+                self.agent_thread = None  # Reset thread reference
+
+            # Remove the socket file only after the server socket is closed
+            if self.server_sock:  # Check if the server socket exists
+                self.server_sock.close()  # Close the server socket
+                os.remove(self.sock_path)
+
+            # print("Mock SSH agent stopped.")
```
p12tic, line 226: No longer needed code.

p12tic (changes requested):
This PR is great, thank you! I had a couple of suggestions for the test, but overall looks great.

tests/integration/build_ssh/test_build_ssh.py
banditopazzo: do you mean not having the mock agent at all?

banditopazzo: sorry, got it, I found a solution to check the key

p12tic: Sorry for not being explicit, but your solution was pretty similar to what I was thinking about.

A more simplier solution would be to do:
```
RUN echo "No id:" >> /result.log 
RUN --mount=type=ssh ssh-add -L >> /result.log

RUN echo "id1:" >> /result.log 
RUN --mount=type=ssh,id=id1 ssh-add -L >> /result.log

RUN echo "id2:" >> /result.log 
RUN --mount=type=ssh,id=id2 ssh-add -L >> /result.log
```

Then in docker compose set command to `cat /result.log` and then check the logs in the test via `podman compose logs` (there are a few tests that do this to use for reference).

banditopazzo: Done

p12tic (approved):

p12tic:
Could you please combine the commit into one and remove `feat:` prefix as we don't use them in this repository? Thanks!

PR #1078, opened 2024-11-27
First attempt at providing support for conditional dependencies

flixman:
Currently all services dependencies are treated unconditionally. However, [docker-compose supports](https://docs.docker.com/compose/how-tos/startup-order/#control-startup) specifying three possible conditions (started, healthy, successfully finished). This PR attempts to provide such support, extending the existing unit tests for dependencies.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1337,7 +1337,7 @@ def flat_deps(services, with_extends=False):
             ext = srv.get("extends", {}).get("service", None)
             if ext:
                 if ext != name:
-                    deps.add(ext)
+                    deps.add(ServiceDependency(ext))
                 continue
         deps_ls = srv.get("depends_on", [])
         if isinstance(deps_ls, str):
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1348,6 +1348,7 @@ def flat_deps(services, with_extends=False):
             deps_ls = [ServiceDependency(k, v.get("condition")) for k, v in deps_ls.items()]
         else:
             raise RuntimeError("depends_on should be a string, a list of strings or a dict")
+        deps.update(deps_ls)
 
         # parse link to get service name and remove alias
         links_ls = srv.get("links", [])
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -25,6 +25,7 @@ import signal
 import subprocess
 import sys
 from asyncio import Task
+from enum import Enum
 
 try:
     from shlex import quote as cmd_quote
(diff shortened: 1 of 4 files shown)


podman_compose.py
```diff
@@ -1273,33 +1274,63 @@ async def container_to_args(compose, cnt, detached=True):
...
+        if isinstance(other, ServiceDependency):
+            return self._name == other.name and self._condition == other.condition
+        return False
 
 
 def flat_deps(services, with_extends=False):
     """
     create dependencies "_deps" or update it recursively for all services
     """
+    def rec_deps(services, service_name, start_point=None):
```
p12tic, line 1312: Moving this function made reviewing way harder.

podman_compose.py
```diff
@@ -1048,8 +1049,8 @@ async def container_to_args(compose, cnt, detached=True):
     if pod:
         podman_args.append(f"--pod={pod}")
     deps = []
-    for dep_srv in cnt.get("_deps", None) or []:
-        deps.extend(compose.container_names_by_service.get(dep_srv, None) or [])
+    for dep_srv in cnt.get("_deps", []):
+        deps.extend(compose.container_names_by_service.get(dep_srv.name, []))
```
p12tic, line 1053: I would move this one into separate commit as well. If the PR breaks something, then it's possible to bisect in smaller increments.

podman_compose.py
```diff
@@ -1308,14 +1339,16 @@ def flat_deps(services, with_extends=False):
                 if ext != name:
                     deps.add(ext)
                 continue
-        deps_ls = srv.get("depends_on", None) or []
-        if isinstance(deps_ls, str):
```
p12tic, line 1312: I think we should support `str` still and raise a warning, otherwise we will break a number of configurations already using this code path.

podman_compose.py
flixman: Noted, but... do you want me to move it out? Because it makes full sense to have it there...

podman_compose.py
flixman: But I am lost now, with respect to this (and the previous) comment: if this PR breaks something, it should be entirely reverted, and done. I can agree to the statement "these refactorings should not be here, but in a separate PR", and I will remove them (and, honestly, most likely not bring them back anymore... because this is just a cosmetic change aimed at simplifying the code, but I will not spend the time to go through a full review process for changes like this... I hope you do not take me wrong).

podman_compose.py
p12tic: I would say it makes things more confusing. When functions are moved to other functions it's often because they need to refer to variables of parent function. The dependencies are less clear.

flixman: Ok: I will move it out, because it's not the point to have a discussion about this :-) You are right when you say that "it's often because they need to refer to variables of parent function" (although this is not the case here), but inner functions are also used when you want to encapsulate functionality on its own function but still not make it public to be called from anywhere else.

podman_compose.py
flixman: Noted. Considering the different flavours of coding / contributing out there, maybe is a good idea to add this to the CONTRIBUTING.md file?

flixman:
@p12tic I am closing the PR to put it back when all is working with the changes you requested.

podman_compose.py
p12tic: Yes, good idea, thanks.

PR #1080, opened 2024-11-29
docker compat: use hyphens to join container names

mattp-:
docker-compose uses - not _ to join container names to project. not a huge thing, but for services making assumptions based on this it can catch off guard, and is easy to make the same.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -223,7 +223,7 @@ def fix_mount_dict(compose, mount_dict, srv_name):
             elif external:
                 vol["name"] = f"{source}"
             else:
-                vol["name"] = f"{compose.project_name}_{source}"
+                vol["name"] = f"{compose.project_name}-{source}"
     return mount_dict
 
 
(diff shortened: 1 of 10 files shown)


p12tic:
Good catch.

I think we should make this configurable, like with network names. Otherwise people will upgrade podman-compose and break their existing deployments.

Documentation for compatibility of network names between podman-compose and docker-compose: https://github.com/containers/podman-compose/blob/main/docs/Extensions.md#compatibility-of-default-network-names-between-docker-compose-and-podman-compose

mattp-:
@p12tic makes sense, i can definitely add that. is there an env var analogue for those switches ? if not I can add that too. in my use-case im dealing with tens-hundreds of compose.yml's wrapped in a single executor, so i would want to influence the behavior there with env var rather than update each yaml. thanks 👍

p12tic:
I'm usually against this, because environment variables increase risk that environment setting will not be set for some invocation and then podman-compose will do bad things.

However I see your point. The project should be flexible enough to satisfy user requirements, so let's do like you say.

Let's do environment variable plus compose setting. If both are present and have different values, podman-compose must throw an exception.

Could you also do the same for network names for consistency?

Thanks

mattp-:
Sure thing 👍

p12tic:
This feature has been implemented via https://github.com/containers/podman-compose/pull/1241.

PR #1085, opened 2024-12-05
Add device_cgroup_rules support

jeremyvisser:
`device_cgroup_rules: []` is defined in the Compose file specification, and corresponds to `podman run --device-cgroup-rule=`

The relevant part of the spec: https://github.com/compose-spec/compose-spec/blob/main/05-services.md#device_cgroup_rules

This recently became necessary for getting passthrough devices to work for me after upgrading to Fedora 41.


jeremyvisser:
Thanks for the prompt. Done.

PR #1087, opened 2024-12-17
Support network scoped service aliases

lisongmin:
To support network scoped aliases, we using the `alias` option in `--network`.
Since the `--network-alias` will override the `alias` set by `--network`,
we do not using the `--network-alias` any more. All aliases will be set
via `--network`.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -936,13 +936,13 @@ def get_net_args_from_networks(compose, cnt):
     mac_address = cnt.get("mac_address")
     service_name = cnt["service_name"]
 
-    aliases = [service_name]
+    aliases_on_container = [service_name]
     # NOTE: from podman manpage:
     # NOTE: A container will only have access to aliases on the first network
     #       that it joins. This is a limitation that will be removed in a later
     #       release.
-    if cnt.get("_aliases"):
-        aliases.extend(cnt.get("_aliases"))
+    aliases_on_container.extend(cnt.get("_aliases", []))
+    aliases_on_net = []
 
     # TODO: add support for per-interface aliases
     #  See https://docs.docker.com/compose/compose-file/compose-file-v3/#aliases
@@ -983,7 +983,7 @@ def get_net_args_from_networks(compose, cnt):
             ipv6 = net_config_.get("ipv6_address")
             # custom extension; not supported by docker-compose v3
             mac = net_config_.get("x-podman.mac_address")
-            aliases.extend(norm_as_list(net_config_.get("aliases")))
+            aliases_on_net = norm_as_list(net_config_.get("aliases", []))
 
             # if a mac_address was specified on the container level, apply it to the first network
             # This works for Python > 3.6, because dict insert ordering is preserved, so we are
@@ -1023,7 +1023,7 @@ def get_net_args_from_networks(compose, cnt):
             net_config = list(multiple_nets.values())[0]
             ipv4 = net_config.get("ipv4_address")
             ipv6 = net_config.get("ipv6_address")
-            aliases.extend(norm_as_list(net_config.get("aliases")))
+            aliases_on_net = norm_as_list(net_config.get("aliases"))
         if ipv4:
             net_args.append(f"--ip={ipv4}")
         if ipv6:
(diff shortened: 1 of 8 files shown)


tests/integration/test_podman_compose_network_scoped_aliases.py
```diff
@@ -0,0 +1,84 @@
+# SPDX-License-Identifier: GPL-2.0
```
p12tic, line 1: It is fine to not have integration test, if the unit tests cover the code like in this case. Thanks for adding it anyway.

p12tic:
@lisongmin Thanks for the PR. I will write several comments, but overall it looks good.

Please note, that such a PR where all changes are added as a single commit is extremely hard to review. Refactoring is mixed with functionality changes and as a result I would have to replay everything in my head.

I've split the PR in small commits that refactor one thing at a time. The end result is almost the same except several [deliberate changes](https://github.com/containers/podman-compose/compare/873c1c0f18cb59402a466001bdb39769233a95d5..681d636eb70a93294be97b39ebd462bdd5fdb742). The functionality changes are then way smaller and way easier to review.

podman_compose.py
```diff
@@ -855,8 +855,9 @@ async def assert_cnt_nets(compose, cnt):
     create missing networks
     """
     net = cnt.get("network_mode")
-    if net and not net.startswith("bridge"):
+    if net:
```
p12tic, line 858: Could you please tell me why is this change needed. It's not obvious to me. If there's some requirement in compose spec that I'm not aware of, please point me to it.

podman_compose.py
```diff
@@ -910,12 +911,9 @@ def get_net_args_from_network_mode(compose, cnt):
         aliases_on_container = [service_name]
         if cnt.get("_aliases"):
             aliases_on_container.extend(cnt.get("_aliases"))
-        net_args.append("--network=bridge")
-        mac_address = cnt.get("mac_address")
-        if mac_address:
-            net_args.append(f"--mac-address={mac_address}")
```
p12tic: I think we lose mac-address here. Is this deliberate? If yes, please split to separate commit with explanation.

podman_compose.py
```diff
@@ -937,29 +935,16 @@ def get_net_args_from_networks(compose, cnt):
...
+
     multiple_nets = cnt.get("networks", {})
     if not multiple_nets:
         if not compose.default_net:
             # The bridge mode in podman is using the `podman` network.
             # It seems weird, but we should keep this behavior to avoid
             # breaking changes.
-            net_args.append("--network=bridge")
-            if mac_address:
-                net_args.append(f"--mac-address={mac_address}")
```
p12tic: I think we lose mac-address here. Is this deliberate? If yes, please split to separate commit with explanation.

podman_compose.py
```diff
@@ -984,17 +969,17 @@ def get_net_args_from_networks(compose, cnt):
...
         default_net_name = default_network_name_for_project(compose, net_, is_ext)
         net_name = ext_desc.get("name") or net_desc.get("name") or default_net_name
 
         ipv4 = net_config_.get("ipv4_address")
         ipv6 = net_config_.get("ipv6_address")
-        # custom extension; not supported by docker-compose v3
+
         mac = net_config_.get("x-podman.mac_address")
-        aliases_on_net = norm_as_list(net_config_.get("aliases", []))
+        aliases_on_net = net_config_.get("aliases")
```
p12tic, line 982: Probably makes sense to move to separate commit, as it's yet another cleanup.

p12tic:
I'm looking into https://github.com/containers/podman-compose/pull/1087/commits/22454fcf7c1b28b9030c5cff2ff1121ec555688f to review functionality changes.

podman_compose.py
lisongmin: I just missing the mac_address here.

The top-level mac_address should be part of networks, and docker >= 25.0 will ignore this attribute
in favor of the network-scoped mac_address. see [compose spec](https://github.com/compose-spec/compose-spec/blob/main/05-services.md#mac_address-1) for more detail.
    
I'll set the mac_address in the bridge mode here for backward compatibility.

podman_compose.py
lisongmin: When using bridge mode, podman-compose will use the default `podman` network.
We do not need to create this network, so just return as other network_mode.

podman_compose.py
lisongmin: Just rollback changes here.

lisongmin:
Done. 

Sorry for the poor PR, thanks for your patience in checking the codes. I learn a lot from it.

PR #1092, opened 2024-12-28
Provide support for cache_from and cache_to fields

flixman:
Currently the fields cache_from and cache_to on the build are ignored. This PR provides support for them in the build block, just by propagating them to the call to podman.

### DIFF
--- a/newsfragments/cache-fields-support.feature
+++ b/newsfragments/cache-fields-support.feature
@@ -1,0 +1,1 @@
+Added support for cache_from and cache_to fields in build section.
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2519,6 +2519,10 @@ def container_to_build_args(compose, cnt, args, path_exists):
             "--build-arg",
             build_arg,
         ))
+    for cache_img in build_desc.get("cache_from", []):
+        build_args.extend(["--cache-from", cache_img])
+    for cache_img in build_desc.get("cache_to", []):
+        build_args.extend(["--cache-to", cache_img])
     build_args.append(ctx)
     return build_args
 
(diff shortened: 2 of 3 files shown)


p12tic (changes requested):
Thanks for the PR. It looks good in principle.

What's missing:
 - unit tests. I did some refactoring and added unit tests for image building functionality in https://github.com/containers/podman-compose/pull/1094
 - release note. Please see newsfragments/README.txt file

Also, podman build [documentation](https://docs.podman.io/en/v5.3.0/markdown/podman-build.1.html#cache-from-image) says that cache-from is ignored unless `--layers` is specified. I suppose this PR won't do what user expects as is, what am I missing?

p12tic:
Also, please rebase instead of merging main, thanks!

flixman:
@p12tic About the --layers flag: I also saw the documentation, but I made a couple of tests and caches are used regardless on --layers being set or not. I have just reported this to [podman](https://github.com/containers/podman/issues/24916), because looks like might be an error on the documentation.

flixman:
@p12tic about the branch update (merging vs rebasing): Can't the "update with merge" be disabled on the GH UI? I just clicked on "update" and it did a merge. Now I have seen I had the option to "update with rebase", but if that is not valid... maybe should just not appear at all?

p12tic:
> Can't the "update with merge" be disabled on the GH UI? I just clicked on "update" and it did a merge.

Unfortunately github UI does not support this. They have different options for actual merging, but for the update button - no.

p12tic:
Still needs a release note.

> - release note. Please see newsfragments/README.txt file

p12tic (approved):

flixman:
@p12tic I have seen there were changes to main, and I have rebased my branch on top of it. Do you need to approve again?

p12tic:
I also combined the fixup commits into the feature commit, so that a commit groups one logical change. See https://metal-k8s.readthedocs.io/en/2.4.1/developer/best_practices/commit.html for an example of commit best practices. Podman-compose should probably have a link to similar page in PR template.

p12tic (approved):

PR #1104, opened 2025-01-11
Add support for disable_dns, dns and ignore on network creation

rjeffman:
This patch adds options to enhance definition of networks which are not available in the docker-compose spec but are useful for creating temporary clusters for experiments and research.

It adds support for setting DNS entries for the network or disable DNS for the network.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -833,6 +833,8 @@ def get_network_create_args(net_desc, proj_name, net_name):
     ipam_config_ls = ipam.get("config", [])
     if net_desc.get("enable_ipv6"):
         args.append("--ipv6")
+    if net_desc.get("x-podman.disable_dns"):
+        args.append("--disable-dns")
 
     if isinstance(ipam_config_ls, dict):
         ipam_config_ls = [ipam_config_ls]
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -835,6 +835,11 @@ def get_network_create_args(net_desc, proj_name, net_name):
         args.append("--ipv6")
     if net_desc.get("x-podman.disable_dns"):
         args.append("--disable-dns")
+    if net_desc.get("x-podman.dns"):
+        args.extend((
+            "--dns",
+            ",".join(norm_as_list(net_desc.get("x-podman.dns"))),
+        ))
 
     if isinstance(ipam_config_ls, dict):
         ipam_config_ls = [ipam_config_ls]
--- a/docs/Extensions.md
+++ b/docs/Extensions.md
@@ -32,6 +32,18 @@ For explanations of these extensions, please refer to the [Podman Documentation]
 The following extension keys are available under network configuration:
 
 * `x-podman.disable-dns` - Disable the DNS plugin for the network when set to 'true'.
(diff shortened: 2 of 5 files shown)


podman_compose.py
```diff
@@ -833,6 +833,12 @@ def get_network_create_args(net_desc, proj_name, net_name):
     ipam_config_ls = ipam.get("config", [])
     if net_desc.get("enable_ipv6"):
         args.append("--ipv6")
+    if net_desc.get("ignore"):
+        args.append("--ignore")
```
p12tic, line 837: Could you tell the use case for this? It should be easy enough to set the network to external, so there must be something I don't understand.

podman_compose.py
```diff
@@ -833,6 +833,12 @@ def get_network_create_args(net_desc, proj_name, net_name):
...
+    if net_desc.get("disable_dns"):
```
p12tic, line 838: Should be called `x-podman.disable_dns`.

podman_compose.py
```diff
@@ -833,6 +833,12 @@ def get_network_create_args(net_desc, proj_name, net_name):
...
+        args.append("--disable-dns")
+    if net_desc.get("dns"):
```
p12tic, line 840: Should be called `x-podman.dns`.

p12tic:
Thanks a lot for the PR.

`disable-dns` and `dns` look good in principle. Please add documentation to https://github.com/containers/podman-compose/blob/main/docs/Extensions.md and a release note to https://github.com/containers/podman-compose/tree/main/newsfragments directory. See the README on how release notes should be written and removed files in https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be for examples.

Also, currently all features are in single commit. Could these be split so that each feature is separate at commit level?

rjeffman:
@p12tic thank you for the review! I'll update the PR with your suggestions later this week (as other tasks have taken me away from this for a while).

podman_compose.py
rjeffman: When you run `podman-compose up` it will create the pod, the containers and the network. If you run `podman-compose down` it will destroy the pod, the containers, but not the network.

By allowing the use of `--ignore` one can ensure that multiple up/down will work and not fail because network already exists.

Yes, it is possible (and easy) to use the external network, but this means that the first `up` is different than another one.

It's most useful while experimenting with service composes rather than in production (which is my primary use for `podman-compose`).

rjeffman: @p12tic I removed the "--ignore" setting from the patch. In the end, using "name: <network>" is enough to re-use an existing network (at least in the case I had).

podman_compose.py
rjeffman: Fixed.

podman_compose.py
rjeffman: Fixed.

rjeffman:
@p12tic, I split the PR in two commits, one for '--disable-dns' and one for '--dns'.

Let me know if the text for `docs/Extensions.md` and `newsfragment` are ok or should be improved.

p12tic (approved):
Looks great, thanks!

PR #1106, opened 2025-01-13
Fix CMD healthchecks running with /bin/sh

### LINKED ISSUE #23: WIP: proper healthchecks support
continue on #22 

https://docs.docker.com/compose/compose-file/#healthcheck
https://docs.docker.com/engine/reference/builder/#healthcheck
https://docs.docker.com/engine/reference/run/#healthcheck

and

https://github.com/containers/libpod/issues/3507

ben-krieger:
Fixes #23 

This change causes `["CMD", "exe", "arg1", "arg2"]` style healthchecks to run using the given command directly and NOT using `/bin/sh -c`.

This is compatible with Podman 1.5.0 and later.

- [ ] Unit tests
- [ ] Release note

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -28,14 +28,8 @@ import sys
 from asyncio import Task
 from enum import Enum
 
-try:
-    from shlex import quote as cmd_quote
-except ImportError:
-    from pipes import quote as cmd_quote  # pylint: disable=deprecated-module
-
 # import fnmatch
 # fnmatch.fnmatchcase(env, "*_HOST")
-
 import yaml
 from dotenv import dotenv_values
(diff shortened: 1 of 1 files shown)


p12tic:
Looks good. Please add unit test for `container_to_args` [here](https://github.com/containers/podman-compose/blob/main/tests/unit/test_container_to_args.py) and a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like).

PR #1108, opened 2025-01-14
Fixes #398: exclude deps on up if --no-deps

### LINKED ISSUE #398: option "--no-deps" is ignored for "podman-compose up"
The option has no effect:
```
$ podman-compose up --help
...
  --no-deps             Don't start linked services.

### PR DESCRIPTION
**Pull request by bailsman**
Fixes #398 by excluding dependent services if --no-deps is given on the command line.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2600,7 +2600,8 @@ def get_excluded(compose, args):
     if args.services:
         excluded = set(compose.services)
         for service in args.services:
-            excluded -= set(x.name for x in compose.services[service]["_deps"])
+            if not args.no_deps:
+                excluded -= set(x.name for x in compose.services[service]["_deps"])
             excluded.discard(service)
     log.debug("** excluding: %s", excluded)
     return excluded
@@ -2658,6 +2659,12 @@ async def run_container(
     return await compose.podman.run(*command, log_formatter=log_formatter)
 
 
+def deps_from_container(args, cnt):
+    if args.no_deps:
+        return set()
+    return cnt['_deps']
+
+
 @cmd_run(podman_compose, "up", "Create and start the entire stack or some of its services")
 async def compose_up(compose: PodmanCompose, args):
     excluded = get_excluded(compose, args)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by p12tic** [comment]
Thanks for the PR. It looks good.

 Please a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like). 

Also the PR needs an integration test. In this particular case I think `tests/integration/test_podman_compose_deps.py` is most relevant for enhancement. Note that it uses compose files from `tests/integration/deps`.

**Comment by bailsman** [comment]
A bit more involved that I thought initially.

* Run (through container_to_args) adds --requires parameters to the podman run command line, which does not work if the dependency is not created/started. Worked around this by adding extra boolean parameter to container_to_args, as is already done for detach.
* Up tries to check if dependencies have successfully started, worked around this by passing in an empty deps set if no-args is given.

File: podman_compose.py [hunk 1]
```diff
@@ -2699,10 +2700,13 @@ async def compose_up(compose: PodmanCompose, args):
...
             continue
-        podman_args = await container_to_args(compose, cnt, detached=args.detach)
+        podman_args = await container_to_args(
+            compose, cnt, detached=args.detach, no_deps=args.no_deps
+        )
         subproc = await compose.podman.run([], podman_command, podman_args)
+        deps = set() if args.no_deps else cnt["_deps"]
         if podman_command == "run" and subproc is not None:
-            await run_container(compose, cnt["name"], cnt["_deps"], ([], "start", [cnt["name"]]))
+            await run_container(compose, cnt["name"], deps, ([], "start", [cnt["name"]]))
```
p12tic, line 2709: Add ```
def deps_from_container(args, cnt):
    if args.no_deps:
        return set()
    return cnt['deps']
```

and call it here and in the call site below.

p12tic:
> The integration test exposed that it is a bit more involved than I thought initially

Great to see tests had good impact :)

The PR looks good overall, I had one nit comment.

Another nit comment, could you move `Fixes #issue number` from commit title to commit description? This is the style used throughout recent history.

Thanks a lot!

p12tic (approved):

Issue #1105, opened 2025-01-12
Environment Variable Default Not Working Correctly v1.3.0


damaestro:
**Describe the bug**

When using the shell default value syntax, the default is not always interpreted. Oddly, it does work if the source variable name doesn't match the destination variable name.

This works as expected with `docker compose`.

**To Reproduce**

`Dockerfile`:

```
FROM docker.io/library/busybox:latest
CMD ["/bin/sh"]
```

`docker-compose.yaml`:

```
version: '3'
services:
  test:
    build: .
    environment:
      TEST_EXPECT_WORKING: ${TEST_EXPECT_WORKING:-working}
      TEST_EXPECT_EMPTY: ${DOES_NOT_EXIST}
      TEST_EXPECT_ONE: 1
      TEST_ODDLY_WORKING: ${HOW_STRANGE:-working}
    command: ["/bin/sh", "-c", "printenv|grep TEST"]
```

**Expected behavior**

`TEST_EXPECT_WORKING: ${TEST_EXPECT_WORKING:-working}` is a common pattern for allowing user-set environment variables to be passed through, while also setting a default. This is working fine with `docker compose`.

**Actual behavior**

`TEST_EXPECT_WORKING` is set to `${TEST_EXPECT_WORKING:-working}` not `working` or previously set `$TEST_EXPECT_WORKING`.

**Output**

Versions:
```
$ podman-compose version
podman-compose version 1.3.0
podman version 5.3.1
```

Defaults expected:
```
$ podman compose run test
TEST_EXPECT_EMPTY=
TEST_ODDLY_WORKING=working
TEST_EXPECT_ONE=1
TEST_EXPECT_WORKING=${TEST_EXPECT_WORKING:-working}
```

User-set expected:
```
$ TEST_EXPECT_WORKING='cli-working' podman compose run test
TEST_EXPECT_EMPTY=
TEST_ODDLY_WORKING=working
TEST_EXPECT_ONE=1
TEST_EXPECT_WORKING=cli-working
```

**Environment:**
 - OS: Linux
 - podman version: 5.3.1
 - podman compose version: 1.3.0

**Additional context**

This might have been working previously as I'm just now noticing it.

damaestro:
I've confirmed this is a bug with podman-compose `1.3.0`. The following is working with `1.2.0`:

Versions:
```
$ podman-compose version
podman-compose version 1.2.0
podman version 5.3.1
```

Defaults expected:
```
$ podman compose run test
[...]

PR #1110, opened 2025-01-15
expand service environment_variables before adding to subs_dict

indrat:
Addresses #1105

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -270,7 +270,6 @@ def rec_subs(value, subs_dict):
             svc_envs = {k: v for k, v in value['environment'].items() if k not in subs_dict}
             # we need to add `svc_envs` to the `subs_dict` so that it can evaluate the
             # service environment that reference to another service environment.
-            subs_dict.update(svc_envs)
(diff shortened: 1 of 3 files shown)


indrat:
Not sure the failing integration test is correct as `docker-compose` warns that `ZZVAR3` is not set and defaults it to an empty string per the acutal output here.

```yaml
› docker-compose -f container-compose.load-.env-in-project.yaml config
WARN[0000] The "ZZVAR3" variable is not set. Defaulting to a blank string.
name: project
services:
  app:
    command:
      - /bin/busybox
      - sh
      - -c
      - env | grep ZZ
    environment:
      ZZVAR1: This value is loaded but should be overwritten
      ZZVAR2: This value is loaded from .env in project/ directory
      ZZVAR3: ""
...
```

p12tic:
@indrat Thanks for the PR. Could you include whole https://github.com/containers/podman-compose/pull/1110#issuecomment-2594169639 comment into the commit description, so that it's part of git history and is clear why the change was done? Thanks!

p12tic (approved):

PR #1117, opened 2025-01-23
Don't raise exception on inexistent services in 'down' command

rjeffman:
When running 'podman-compose down <service>', if service is not part of the compose, a KeyError exception was raised it function 'get_excluded'.

By only allowing evaluation of services that exist in the compose provides a cleaner and gentler exit for this case.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2648,7 +2648,8 @@ def get_excluded(compose, args):
     if args.services:
         excluded = set(compose.services)
         for service in args.services:
-            if not args.no_deps:
+            # we need 'getattr' as compose_down_parse dose not configure 'no_deps'
+            if service in compose.services and not getattr(args, "no_deps", False):
                 excluded -= set(x.name for x in compose.services[service]["_deps"])
             excluded.discard(service)
     log.debug("** excluding: %s", excluded)
--- a/newsfragments/compose_down_inexistent_service.bugfix
+++ b/newsfragments/compose_down_inexistent_service.bugfix
@@ -1,0 +1,1 @@
+- Fixed KeyError in case podman-compose down was called with an inexistent service


podman_compose.py
```diff
@@ -2606,7 +2606,7 @@ def get_excluded(compose, args):
     if args.services:
         excluded = set(compose.services)
         for service in args.services:
-            if not args.no_deps:
+            if service in compose.services and not getattr(args, "no_deps", False):
```
p12tic, line 2609: Any reason to change `args.no_deps` to use getattr?

rjeffman: This is one of the reasons I kept this PR in "work in progress", there were some cases (which I don't know if are valid or not) that `args.no_deps` was undefined.

I'm sorry, but I've been drowning in work lately. I plan to get back to podman-compose and try to help with its development, but it will still take a few weeks.

p12tic: Oh I see - compose_up_parse configures `no-deps` argument, but compose_down_parse doesn't.

p12tic:
Looks good. Please also add:

 -  a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like).

 - integration test. For a basic example you can look into https://github.com/containers/podman-compose/blob/main/tests/integration/env-tests/test_podman_compose_env.py

podman_compose.py
p12tic: Maybe it makes sense to add a comment saying that `compose_down_parse` does not configure `no-deps` command argument. This would make it clear why the code is how it is.

rjeffman:
@p12tic sorry for the long delay (work and a needed PTO were in place).

I fixed the PR with your requests. Let me know if anything else is needed.

p12tic (approved):

Issue #1109, opened 2025-01-14
Relative Paths To Compose File Fail in 1.3.0


LArkema:
**Describe the bug**
Starting in podman-compose 1.3.0, compose files fail to open when provided as a relative path. 

**To Reproduce**
1. what is the content of the current working directory: 

```
- project/
   - docker-compose.yaml
 ```

2. what is the sequence of commands you typed
```
[user]# pwd
/home/user
[user]# podman-compose -f project/docker-compose.yml build
```

**Expected behavior**
The build process begins following the instructions in the compose file

**Actual behavior**
A FileNotFoundError is returned
```
[user]# podman-compose -f project/docker-compose.yml build
Traceback (most recent call last):
  File "/usr/local/bin/podman-compose", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/podman_compose.py", line 3711, in main
    asyncio.run(async_main())
  File "/usr/lib64/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/podman_compose.py", line 3707, in async_main
    await podman_compose.run()
  File "/usr/local/lib/python3.11/site-packages/podman_compose.py", line 1859, in run
    self._parse_compose_file()
  File "/usr/local/lib/python3.11/site-packages/podman_compose.py", line 1954, in _parse_compose_file
    with open(filename, "r", encoding="utf-8") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'project/docker-compose.yml'
```

However, both local directory references ( `podman-compose -f docker-compose.yml build` executed from `/home/user/project/` ) and absolute paths ( `podman-compose -f /home/user/project/docker-compose.yml build` executed from anywhere) work and read the compose file.

**Output**

```
# podman-compose version
podman-compose version 1.3.0
podman version 4.9.4-rhel
```

**Environment:**
[...]

PR #1124, opened 2025-01-27
remove incorrect os.chdir, handle relative extends/includes

notdian:
Running `docker-compose -f containers/base` resulted in a `FileNotFoundError`. However, it worked after `cd containers` and then running `docker-compose -f base`. It appears that the removal of `chdir` operation was overlooked.

### DIFF
--- a/newsfragments/1125.bugfix
+++ b/newsfragments/1125.bugfix
@@ -1,0 +1,1 @@
+- Fix handling of relative includes and extends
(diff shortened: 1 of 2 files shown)


p12tic:
> Seems like this broke integration tests, maybe the fix should be to make all files absolute before the loop?

Probably. This needs checking.

p12tic:
> the chdir and services/includes are related, you can't have one without the other, if you think otherwise, I can split them.

You're right, my head is not working.

cfunkhouser:
I have confirmed that this fixes #1109 for my use cases.

p12tic:
Rebased.

flexagoon:
@p12tic could you please tag a release with this fix?

PR #1133, opened 2025-02-07
Implement x-podman.pod_args to override --pod-args default

### LINKED ISSUE #1057: podman-compose extension for --pod-args
The recent PR https://github.com/containers/podman-compose/pull/964 implemented the ability to set `--in-pod` in the compose file. However, there still does not appear to be a way to set `--pod-args` in the compose file. 
This would be pretty useful as it would allow for options such as `--infra`, `--infra-command`, `--infra-image`, `--infra-name`, `--share` and more to be set in the compose file instead of needing to pass them in with `--pod-args`.

Compose example:
```
x-podman:
    in_pod: true
    pod_args: '--infra=true --infra-name=pod_unipi_infra --share=""'
```

Possible partial implementation:
```diff --git a/podman_compose.py b/podman_compose.py
index 6b974f5..43305fb 100755
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -1762,6 +1762,12 @@ class PodmanCompose:
         # otherwise use `in_pod` value provided by command line
         return self.global_args.in_pod_bool
 
+    def resolve_pod_args(self):
+        if self.global_args.pod_args == "--infra=false --share=":
+            self.global_args.pod_args = self.x_podman.get("pod_args", True)
+        # otherwise use `pod_args` value provided by command line
+        return self.global_args.pod_args
+
     def _parse_compose_file(self):
         args = self.global_args
         # cmd = args.command
@@ -2008,6 +2014,7 @@ class PodmanCompose:
         self.x_podman = compose.get("x-podman", {})
 
         args.in_pod_bool = self.resolve_in_pod()
+        args.pod_args = self.resolve_pod_args()
         pods, containers = transform(args, project_name, given_containers)
         self.pods = pods
         self.containers = containers

```

neocturne:
Allow setting an argument list as x-podman.pod_args to override the default value `--infra=false --share=`. `--pod-args` passed on the command line takes precedence over the value set in docker-compose.yml; the values are not merged.

Fixes #1057.
Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>


podman_compose.py
```diff
@@ -2133,6 +2145,7 @@ def _parse_compose_file(self):
         self.x_podman = compose.get("x-podman", {})
 
         args.in_pod_bool = self.resolve_in_pod()
+        args.pod_arg_list = self.resolve_pod_args()
```
p12tic, line 2148: I think it makes sense to attach `pod_arg_list` to self and then do all processing in `resolve_pod_args()`. That is, include the following in that function too:

```
        if self.global_args.pod_args is None:
            pod_arg_list = None
        else:
            pod_arg_list = shlex.split(self.global_args.pod_args)
```

docs/Extensions.md
```diff
@@ -174,3 +174,15 @@ services:
 x-podman:
     in_pod: false
 ```
+
+It is also possible to override the default arguments for pod creation that are
+used when --pod-args is not passed on the command line:
```
p12tic, line 179: Makes sense to document the default too.

tests/integration/pod_args/test_podman_compose_pod_args.py
```diff
@@ -0,0 +1,176 @@
...
+                "podman",
+                "pod",
+                "inspect",
+                pod_name,
+            ])
+            pod_info = json.loads(output.decode('utf-8'))
+            # Podman 5.0 changed pod inspect to always output a list.
+            # Check type to support both old and new version.
+            if isinstance(pod_info, list):
+                pod_info = pod_info[0]
```
p12tic, line 67: Makes sense to extract the above 5 lines into separate function, e.g. `load_pod_info`.

p12tic (changes requested):
Thanks for the PR, it looks good in principle. I had several comments about how it could be improved before merging.

docs/Extensions.md
neocturne: Done

tests/integration/pod_args/test_podman_compose_pod_args.py
neocturne: Done

podman_compose.py
neocturne: Makes sense. I also restructured the code of `resolve_pod_args` a bit after including the piece you suggested.

p12tic (approved):

PR #1098, opened 2025-01-01
Fixes #1096: read log output by chunks if a line is too long

whym:
This is meant to solve the freeze issue described in #1096.

The approach taken here is the more conservative, but less simple version I mentioned in the issue.

PR #1140, opened 2025-02-11
Fix comment, add tests, improve coding style

whym:
This is a follow up to https://github.com/containers/podman-compose/pull/1098 to fix some wrong comment statements, and minor coding style issues. The runtime code remains unchanged.

### DIFF
--- a/tests/unit/test_compose_run_log_format.py
+++ b/tests/unit/test_compose_run_log_format.py
@@ -8,10 +8,10 @@ from podman_compose import Podman
 
 
 class DummyReader:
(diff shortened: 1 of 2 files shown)


tests/unit/test_compose_run_log_format.py
```diff
@@ -24,34 +24,44 @@ def setUp(self):
         self.buffer = io.StringIO()
 
     async def test_single_line_single_chunk(self):
-        reader = DummyReader([b'hello, world\n'])
```
p12tic, line 27: Could we keep single quotes?

Issue #1142, opened 2025-02-18
Ruff and black disagree on what the code should look like


italomaia:
Just noticed `ruff podman_compose.py` and `black podman_compose.py` generate two different outputs that fail for checks against one-another. As pre-commit checks against black and the ci checks against ruff, and the contribution guide suggests setting up pre-commit, it would be reasonable to either alter pre-commit by replacing black or the ci by replacing ruff. 

**To Reproducte**

```
black podman_compose.py
ruff podman_compose.py
```

**Expected behavior**
It should be possible to align the code against a single code formatter.

**Actual behavior**
Ruff and Black disagree on what "properly formatted" looks like.

**Environment:**
 - OS: Linux
 - podman-compose version 1.3.0
 - podman version 4.9.3

**Additional context**

Tested against main branch.

PR #1143, opened 2025-02-18
Bug: replaced black with ruff on pre-commit

italomaia:
## Contributor Checklist:

As black and ruff have a different idea on how the formatted code should look like, by using pre-commit, which is advised in the contributing guide, we block any contribution, as ruff is used in the ci format check and black is used in the pre-commit. This MR fixes this: https://github.com/containers/podman-compose/issues/1142

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,17 +1,9 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 23.3.0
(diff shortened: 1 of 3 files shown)


p12tic:
Looks good. Please limit commit message line length to 72 chars and split it into message subject and message body parts.

italomaia:
Can do; minute

italomaia:
Done.

p12tic (approved):

PR #1149, opened 2025-02-26
Fix event loop handling for Windows platform in compose_up function

AlexandreAANP:
Hi,
I tried to run podman-compose on Windows and always break the script because of NotImplementedError in asyncio/events.
After a fast search, I noted that Windows doesn't have support for signals, and I found a code solution to solve this issue in this stack overflow link: https://stackoverflow.com/questions/58014623/python-asyncio-notimplementederror

I added an if condition to check if it's a Windows system or not, not sure if the best way to compare, but it works.

### DIFF
--- a/newsfragments/windows_not_implemented_error.bugfix
+++ b/newsfragments/windows_not_implemented_error.bugfix
@@ -1,0 +1,1 @@
+- Fixed NotImplementedError in case script is interrupted on Windows
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2776,9 +2776,11 @@ async def compose_up(compose: PodmanCompose, args):
         max_service_length = curr_length if curr_length > max_service_length else max_service_length
 
     tasks = set()
-
-    loop = asyncio.get_event_loop()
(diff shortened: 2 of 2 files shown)


p12tic:
Reading through the stackoverflow post and linked pages, seems like WindowsSelectorEventLoopPolicy was needed for aio-dns library which we don't use. 

Maybe it is enough for podman-compose to simply not call add_signal_handler and not touch the event loop policy?

p12tic:
Looks good, please add release note in newsfragments directory.

AlexandreAANP:
Sorry, I'm kind of new to project contribution, how should be the news fragment type for this modification?

p12tic:
@AlexandreAANP No problems at all, I have been newbie some time ago as well :-)

What you need to do is to use interactive rebase to combine the commits into one. Then you would need to amend the commit to adjust its commit message so that the `signed-off-by` line can be added.

Here is tutorial on how to use amend and rebase https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history

p12tic (approved):

p12tic:
Thanks!

PR #1152, opened 2025-02-27
Add quiet flag to podman-compose config

IamTheFij:
This skips printing and is useful for validating config files.

This is present in `docker-compose` and `docker compose` as well: https://docs.docker.com/reference/cli/docker/compose/config/

Do you think this warrants news? I could add that, but it seems pretty minor.

Also, since it is testing the print, the unit test didn't make sense. I added an integration test instead.

### DIFF
--- a/tests/integration/profile/test_podman_compose_config.py
+++ b/tests/integration/profile/test_podman_compose_config.py
@@ -80,3 +80,20 @@ class TestComposeConfig(unittest.TestCase, RunSubprocessMixin):
             actual_services[service] = service in actual_output
 
         self.assertEqual(expected_services, actual_services)
+
+    def test_config_quiet(self):
+        """
+        Tests podman-compose config command with the --quiet flag.
+        """
+        config_cmd = [
+            "coverage",
+            "run",
+            podman_compose_path(),
+            "-f",
+            profile_compose_file(),
+            "config",
+            "--quiet",
+        ]
+
+        out, _ = self.run_subprocess_assert_returncode(config_cmd)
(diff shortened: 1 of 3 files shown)


tests/integration/profile/test_podman_compose_config.py
```diff
@@ -80,3 +80,22 @@ def test_config_profiles(self, profiles, expected_services):
...
+            "-f",
+            profile_compose_file(),
+            "config",
+            "--quiet",
+        ]
+
+        out, _ = self.run_subprocess_assert_returncode(config_cmd)
+
+        string_output = out.decode("utf-8")
+        self.assertEqual("", string_output)
```
p12tic, line 101: I think `string_output` is not needed and can be inlined - code is obvious anyway. Also, assertEqual(actual, expected) is more readable in my opinion, though yes, this is not followed consistently in the codebase.

p12tic:
> Do you think this warrants news? I could add that, but it seems pretty minor.

It's best to expose all user-visible changes. If only so that project appears more active.

> Also, since it is testing the print, the unit test didn't make sense. I added an integration test instead.

Agreed, thanks.

Separately, please add signed-off-by line, there's a policy on the containers organization that requires it. More info: https://github.com/containers/podman-compose/pull/1152/checks?check_run_id=37949903107

IamTheFij:
Done! I also created a new PR (#1153) to catch committing without sign-off for future new contributors. Hopefully that bubbles things up earlier to them so you don't have to be the one to tell them on the PR.

p12tic (approved):

p12tic:
Thanks!

PR #1153, opened 2025-02-28
Add hook to check for signoff in commit messages

IamTheFij:
Since this is checked on PR, it could also be checked at commit so users can avoid making commits to the tree without expected documentation.


p12tic:
Note that commit message body should be wrapped at 72 characters per convention.

Issue #267, opened 2021-01-19
Podman not supporting --scale flag 


unvaare:
I have been facing this issue where i am trying to scale down or up the services using --scale option but podman-compose is ignoring the altogether and spinning up containers as given in compose file.

Can someone please help?

metal3d:
A Pod shares the same IP for all containers.
Scaling the same service will try to open several times the same ports on the Pod. So it's not possible.
Compose seems to not propose a way to create one pod per service. That would be the solution. But at this time, it's impossible.

And it's a pitty.

heitorPB:
Is there a workaround for this issue?

me-coder:
@metal3d,
@unvaare is curious regarding `podman-compose` ignoring `--scale`.
`podman-compose` as default doesn't create pods unless directed to do so with `--in-pod`, but creates containers just like `docker compose` does (`godman-compose` more of a drop-in replacement for `docker compose`).
So `--scale` seems like a necessary feature support, which is missing from `podman-compose`.

Typically after `podman-compose up -d` we don't see any pods:
```
# podman pod ps
POD ID      NAME        STATUS      CREATED     INFRA ID    # OF CONTAINERS
```

Unless the command is accompanied by `podman-compose --in-pod hosting up -d`.
```
# podman pod ps
POD ID        NAME         STATUS      CREATED         INFRA ID    # OF CONTAINERS
d391e937fe90  pod_hosting  Running     46 seconds ago              1
```

In later case, what you have explained might apply, but for the former case there would be different containers with individually separate IPs that would be created and shouldn't be an issue.

metal3d:
It seems that there is now a `x-podman` directive that may set `in-pod` to false.

epag:
I also just ran into this same issue, did anyone find out a workaround?

me-coder:
@unvaare, @epag,
I've submitted #1159 to address this, which is a partial working fix.
If you are still active and interested in this, could you help me test and review this change-set?

epag:
Happy to help out in what ways I can later this week when I get some free time, thanks for addressing this! Fingers crossed my IT department is able to upgrade the version to include the change 🤞

mokibit:
I can no longer reproduce the issue. I ran `podman-compose` as follows: 
 Check the version
```
podman-compose --version
podman-compose version 1.4.0
podman version 5.4.2
```
1.1.  `compose.yml`
```
services:
  test1:
    image: busybox
    command: ["echo", "test1"]
[...]

PR #1159, opened 2025-03-05
Updates handling of scale/replicas parameter in CLI and compose file

me-coder:
# Purpose
Addresses issue: #267

## Contributor Checklist:

> If this PR adds a new feature that improves compatibility with docker-compose, please add a link

This PR attempts to address missing (the code to accommodate this was already present, just broken) `--scale <SERVICE_NAME>=<NUMBER_OF_REPLICAS>` option for `podman-compose up`:
https://docs.docker.com/reference/cli/docker/compose/up/#options

# Testing
**Test file**:
```docker-compose.yaml
name: podman-compose

services:
  app:
    image: docker.io/library/busybox:latest
    tty: true
    # scale: 3
    # deploy:
    #   mode: replicated
    #   replicas: 2

  app1:
    image: docker.io/library/busybox:latest
    tty: true
    depends_on:
      - app

  app2:
    image: docker.io/library/busybox:latest
    tty: true

  app3:
    image: docker.io/library/busybox:latest
    tty: true

```

**With compose file**:
```
$ python ./podman_compose.py -f ./test.yml up -d
1283f00b651460df07514d3be6d01e0ea76529782ead0292c977077b72229937
d3b036808fff09a6e107c3ba769ad03a2c024d148581a24bdd5e3128b952d976
663cb01d16c4e7194145b63a306d6e225f7c05b1ff97103ae8e2df57144233b7
8dd9e7f5588ec2d1ecd16565cdf371be8fce87442f902e48bec7bf97276026b4

$ python ./podman_compose.py -f ./test.yml ps
CONTAINER ID  IMAGE                             COMMAND     CREATED        STATUS        PORTS       NAMES
d3b036808fff  docker.io/library/busybox:latest  sh          7 seconds ago  Up 7 seconds              podman-compose_app_1
663cb01d16c4  docker.io/library/busybox:latest  sh          6 seconds ago  Up 6 seconds              podman-compose_app_2
8dd9e7f5588e  docker.io/library/busybox:latest  sh          5 seconds ago  Up 5 seconds              podman-compose_app_3

$ python ./podman_compose.py -f ./test.yml down
podman-compose_app_1
podman-compose_app_3
podman-compose_app_2
podman-compose_app_3
podman-compose_app_2
podman-compose_app_1
1283f00b651460df07514d3be6d01e0ea76529782ead0292c977077b72229937
podman-compose_default
```

**Scale existing services using CLI**:
```bash
$ python ./podman_compose.py -f ./test.yml up -d --scale app=3
5a76ea112148b71be792cd4c9e77f7b505cbdc2432118eaa2bff8003753b1bb9
ca5b2c05a5198392d64560b63a57e8db0cf2ecdb7f1171f10e800824cc1974d3
8f835d6759a2614c59ee7d21abd0859e6cc96ffec06bdceae3dc55ab096bd488
1fbc4958fa4439d8e0349459f77f1d59026ed3414b815026992611ee8518e141

$ python ./podman_compose.py -f ./test.yml ps
CONTAINER ID  IMAGE                             COMMAND     CREATED        STATUS        PORTS       NAMES
ca5b2c05a519  docker.io/library/busybox:latest  sh          5 seconds ago  Up 5 seconds              podman-compose_app_1
8f835d6759a2  docker.io/library/busybox:latest  sh          4 seconds ago  Up 4 seconds              podman-compose_app_2
1fbc4958fa44  docker.io/library/busybox:latest  sh          3 seconds ago  Up 3 seconds              podman-compose_app_3

$ python ./podman_compose.py -f ./test.yml down
podman-compose_app_3
podman-compose_app_2
podman-compose_app_1
podman-compose_app_3
podman-compose_app_2
podman-compose_app_1
5a76ea112148b71be792cd4c9e77f7b505cbdc2432118eaa2bff8003753b1bb9
podman-compose_default
```

# To-Do
- [x] Scale-up with CLI works (although with errors on command line, which should be warnings). But, Scale-down doesn't work.
=> That part of code yet needs to be addressed, may be through a separate issue.
- [x] Command `down` (to destroy the scaled containers/replicas) works only based on value assigned in compose yaml file. This instead, needs to be addressed to stop all running replicas.
=> Done

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2107,7 +2107,23 @@ class PodmanCompose:
         container_names_by_service = {}
         self.services = services
         for service_name, service_desc in services.items():
-            replicas = try_int(service_desc.get("deploy", {}).get("replicas"), fallback=1)
+            replicas = 1
+            if "scale" in args and args.scale is not None:
+                # Check `--scale` args from CLI command
+                scale_args = args.scale.split('=')
+                if service_name == scale_args[0]:
+                    replicas = try_int(scale_args[1], fallback=1)
+            elif "scale" in service_desc:
+                # Check `scale` value from compose yaml file
+                replicas = try_int(service_desc.get("scale"), fallback=1)
+            elif (
+                "deploy" in service_desc
+                and "replicas" in service_desc.get("deploy", {})
+                and "replicated" == service_desc.get("deploy", {}).get("mode", '')
+            ):
+                # Check `deploy: replicas:` value from compose yaml file
+                # Note: All conditions are necessary to handle case
+                replicas = try_int(service_desc.get("deploy", {}).get("replicas"), fallback=1)
 
             container_names_by_service[service_name] = []
             for num in range(1, replicas + 1):
@@ -3423,12 +3439,13 @@ def compose_up_parse(parser):
         action="store_true",
         help="Remove containers for services not defined in the Compose file.",
     )
+    # `--scale` argument needs to store as single value and not append,
+    # as multiple scale values could be confusing.
     parser.add_argument(
         "--scale",
         metavar="SERVICE=NUM",
-        action="append",
-        help="Scale SERVICE to NUM instances. Overrides the `scale` setting in the Compose file if "
-        "present.",
+        help="Scale SERVICE to NUM instances. "
+        "Overrides the `scale` setting in the Compose file if present.",
     )
     parser.add_argument(
         "--exit-code-from",
--- a/newsfragments/267.bugfix
+++ b/newsfragments/267.bugfix
@@ -1,0 +1,1 @@
+- Fixed the --scale flag as described in issue #267 [Podman not supporting --scale flag]
--- a/tests/integration/service_scale/scaleup_cli/docker-compose.yml
+++ b/tests/integration/service_scale/scaleup_cli/docker-compose.yml
@@ -1,0 +1,6 @@
+name: podman-compose
+
+services:
+  service1:
+    image: docker.io/library/busybox:latest
+    tty: true
--- a/tests/integration/service_scale/scaleup_deploy_replicas_parameter/docker-compose.yml
+++ b/tests/integration/service_scale/scaleup_deploy_replicas_parameter/docker-compose.yml
@@ -1,0 +1,9 @@
+name: podman-compose
+
+services:
+  service1:
+    image: docker.io/library/busybox:latest
+    tty: true
+    deploy:
+      mode: replicated
+      replicas: 2
--- a/tests/integration/service_scale/scaleup_scale_parameter/docker-compose.yml
+++ b/tests/integration/service_scale/scaleup_scale_parameter/docker-compose.yml
@@ -1,0 +1,7 @@
+name: podman-compose
+
+services:
+  service1:
+    image: docker.io/library/busybox:latest
+    tty: true
+    scale: 2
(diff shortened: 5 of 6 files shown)


podman_compose.py
```diff
@@ -2080,7 +2080,17 @@ def _parse_compose_file(self):
         container_names_by_service = {}
         self.services = services
         for service_name, service_desc in services.items():
-            replicas = try_int(service_desc.get("deploy", {}).get("replicas"), fallback=1)
+            replicas = try_int(args.scale, fallback=1) if "scale" in args else 1
```
p12tic, line 2083: I think we should use `None` to signify that no value has been supplied, and then handle the None case at the very end. Right now if explicit 1 value is passed it may be overridden by next steps.

me-coder: I've updated the code to handle the cases:
1. `--scale` from the CLI `up` command
2. `scale` from compose yaml file
3. `{"deploy": {"replicas": <NUM>}` from compose yaml file

Was tempted to use the new fancy match-cases, but didn't to maintain compatibility with pre-Python-2.10 versions.

Let me know if the new code is fine.

podman_compose.py
```diff
@@ -2870,38 +2904,44 @@ async def compose_down(compose: PodmanCompose, args):
             timeout = str_to_seconds(timeout_str)
         if timeout is not None:
             podman_stop_args.extend(["-t", str(timeout)])
-        down_tasks.append(
-            asyncio.create_task(
-                compose.podman.run([], "stop", [*podman_stop_args, cnt["name"]]), name=cnt["name"]
-            )
-        )
-    await asyncio.gather(*down_tasks)
+        await compose.podman.run([], "stop", [*podman_stop_args, cnt["name"]])
```
p12tic, line 2907: This will way on all services to stop one by one. This is problematic in case of large number of services. One deployment I have is 40 services for example.

p12tic:
No need to add `(Issue: containers#267)` in commit description.

podman_compose.py
me-coder: They would stop in parallel as await yields from the `compose.podman.run([], "stop", [*podman_stop_args, cnt["name"]])`.
So rather than waiting for `asyncio.gather` to schedule the future tasks/co-routines, here they get scheduled as soon as they are passed.

However, the concern is valid. It won't wait for all the stop tasks to finish before proceeding to `await compose.podman.run([], "rm", [cnt["name"]])`. I'll try to fix it. (Ideally should be upgraded to TaskGroup, but for that there has to be an upgrade mandate to Python 3.11 for compatibility reasons).

me-coder: Done. Updated the code to use `asyncio.gather` for `stop` and `rm` command tasks.

p12tic:
Overall code looks good.

We need integration tests. You've already done most of the work and provided compose file and expected output in the PR description. What's left is to automate it. Check out tests/integration for examples of how this can be done.

[Updates to compose_down command to match docker compose behavior](https://github.com/containers/podman-compose/pull/1159/commits/902454d73fda2aee371309019a7594273f9d388a) should probably live in another PR because the change is unrelated. This would make it easier to land "[Updates handling of scale/replicas through CLI & compose file](https://github.com/containers/podman-compose/pull/1159/commits/2edd4f376c898106adaab21c1cb8ce51a0b2b924)" commit.

Talking about [Updates to compose_down command to match docker compose behavior](https://github.com/containers/podman-compose/pull/1159/commits/902454d73fda2aee371309019a7594273f9d388a), it also needs tests and release note (probably several, because more than one thing is fixed). Also, I would say that the commit should itself be split into several, because right now it's hard to understand what the commit is trying to achieve. It's clear that it does improve something, but it contains several improvements merged into one.

p12tic:
I removed `Updates to compose_down command to match docker compose behavior` commit because it needs a separate release note and tests. Also I think that it's probably enough to add `-f` to podman rm.

Issue #1073, opened 2024-11-11
unsupported interpolation for COMPOSE_PROJECT_NAME


europ:
**Describe the bug**
podman-compose does not support interpolation for COMPOSE_PROJECT_NAME variable - anywhere in compose file.

**To Reproduce**
1. Create docker-compose.yml
```yaml
name: custom_project-name
services:
  test:
    container_name: '${COMPOSE_PROJECT_NAME}-test'
    image: container-registry.oracle.com/os/oraclelinux:8-slim
    command:
      - sleep
      - "5"
```
1. Run `podman-compose up`

**Expected behavior**

```bash
# created container
```

**Actual behavior**

```console
Error: running container create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
[test] | Error: unknown shorthand flag: 't' in -test
[test] | See 'podman start --help'
```

**Output**

```
$ podman-compose version
podman-compose version 1.2.0
podman version 5.2.5

$ podman-compose up
Error: running container create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
[test] | Error: unknown shorthand flag: 't' in -test
[test] | See 'podman start --help'

```

**Environment:**

 - OS: Mac
 - podman version: 5.2.5
 - podman compose version: 1.2.0

**Additional context**

- value for COMPOSE_PROJECT_NAME is loaded correctly https://github.com/containers/podman-compose/blob/4ae6ccb5c07d6e69fa4529323742217ec21c0744/podman_compose.py#L1903
- https://docs.docker.com/reference/compose-file/interpolation/
- https://docs.docker.com/compose/how-tos/environment-variables/envvars/
- https://docs.docker.com/reference/compose-file/version-and-name/#name-top-level-element
- https://github.com/compose-spec/compose-spec/blob/main/04-version-and-name.md
    
    ```
    Name top-level element
    
    The top-level name property is defined by the Specification as the project name
    to be used if you don't set one explicitly. Compose offers a way for you to override
    this name, and sets a default project name to be used if the top-level name element 
    is not set.
    
    Whenever a project name is defined by top-level name or by some custom
    mechanism, it is exposed for interpolation and environment variable resolution
    as COMPOSE_PROJECT_NAME
    ```

PR #1165, opened 2025-03-12
Fix `COMPOSE_PROJECT_NAME` interpolation

r10r:
This fixes the interpolation of the environment variable **COMPOSE_PROJECT_NAME**
when it is generated from the  top-level`name` value within the *compose.yaml*.

See #1073 and the integration test description for details.

This PR was developed independently from #805 (because I did not look for related open PRs first) but addresses the same issue and contains the required integration test. I think #805 can be closed if this PR is merged.

### DIFF
--- a/tests/integration/env-tests/test_podman_compose_env.py
+++ b/tests/integration/env-tests/test_podman_compose_env.py
@@ -67,3 +67,23 @@ class TestComposeEnv(unittest.TestCase, RunSubprocessMixin):
                 compose_yaml_path(),
                 "down",
             ])
+
+    def test_project_name_override(self):
+        try:
+            output, _ = self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_yaml_path(),
+                "run",
+                "-e",
+                "COMPOSE_PROJECT_NAME=project-name-override",
+                "project-name-test",
+            ])
+            self.assertIn("project-name-override", str(output))
+        finally:
+            self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_yaml_path(),
+                "down",
+            ])
--- a/tests/integration/env-tests/test_podman_compose_env.py
+++ b/tests/integration/env-tests/test_podman_compose_env.py
@@ -36,3 +36,34 @@ class TestComposeEnv(unittest.TestCase, RunSubprocessMixin):
                 compose_yaml_path(),
                 "down",
             ])
+
+    """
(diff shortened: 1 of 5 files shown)


r10r:
After looking at #805 I noticed that apart from the interpolation the evaluation order is wrong. 
The compose spec defines the evaluation order here https://docs.docker.com/compose/how-tos/project-name/#set-a-project-name

I'll address add a separate commit that addresses this issue including another integration test.

tests/integration/env-tests/test_podman_compose_env.py
```diff
@@ -49,6 +49,7 @@ def test_env(self):
     - https://docs.docker.com/compose/how-tos/environment-variables/envvars/
     - https://github.com/compose-spec/compose-spec/blob/main/04-version-and-name.md
     """
+
```
p12tic, line 52: Unnecessary change.

p12tic:
@r10r Thanks for the PR, code changes look good.

Please add a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like). In this PR I think there are not one but two fixes.

tests/integration/env-tests/test_podman_compose_env.py
r10r: This was required to make `ruff` happy. Otherwise the formatting check fails. Changed in https://github.com/containers/podman-compose/compare/198623df38e818d8f1f0baac35380d01841d286c..6fd5b545f163a69dd0c45b88d57e3cbead5620d0

r10r:
> @r10r Thanks for the PR, code changes look good.
> 
> Please add a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like). In this PR I think there are not one but two fixes.

Thanks for the review. I'll add the release note as soon as possible.

tests/integration/env-tests/test_podman_compose_env.py
p12tic: Understood. I suppose that new line should have been added to the commit that added the comment and the test. Probably not worth the effort at this point.

p12tic (approved):

p12tic:
Thank you very much

Issue #1163, opened 2025-03-12
ValueError when merging args with different syntax


piotr-kubiak:
**Describe the bug**
When trying to merge compose files with `args` property in a dict and a list form, the following error occurs: 

`ValueError: can't merge value of [args] of type <class 'list'> and <class 'dict'>`

**To Reproduce**
Create two files:

```
# docker-compose-1.yml
services:
  www:
    build:
      dockerfile: Dockerfile
      args:
        FOO: foo
```

```
# docker-compose-2.yml
services:
  www:
    build:
      dockerfile: Dockerfile
      args:
        - BAR=BAR
```

Then run:
`podman-compose -f docker-compose-1.yml -f docker-compose-2.yml config`

**Expected behavior**
It should be possible to merge above configs, as both syntaxes are valid.

**Version**
```
$ podman-compose --version
podman-compose version 1.3.0
podman version 5.2.2
```

PR #1166, opened 2025-03-12
Allow merging of args in both list and dict syntax

piotr-kubiak:
Fixes https://github.com/containers/podman-compose/issues/1163.

### DIFF
--- a/newsfragments/1163.bugfix
+++ b/newsfragments/1163.bugfix
@@ -1,0 +1,1 @@
+- Fixed error when merging args in list and dict syntax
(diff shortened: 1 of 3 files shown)


p12tic (changes requested):
Looks great, thanks.

Could you please add a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like).

PR #1168, opened 2025-03-17
Properly surface errors from build commands

### LINKED ISSUE #476: podman-compose build exits 0 when build fails
Hello!

I have been using podman-compose for a bit and love it but I noticed that when I run `podman-compose build` it always exits 0 regardless of if the build failed or not. 

It does print out the exit code of the underlying podman command but I don't want to parse the output for that. Would it be possible to have the build command exit with the underlying exit code if it runs into a failure?

I have my local version of podman-compose patched with:
```
obj = types.SimpleNamespace()
compose.podman.run([], "build", build_args, sleep=0, obj=obj)
if obj.exit_code != 0:
  sys.exit(obj.exit_code)
```
on https://github.com/containers/podman-compose/blob/devel/podman_compose.py#L1898. I'm not sure if there's a better solution.

charliemirabile:
fixes #476

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2613,7 +2613,7 @@ async def compose_build(compose, args):
     status = 0
     for t in asyncio.as_completed(tasks):
         s = await t
-        if s is not None:
+        if s is not None and s != 0:
             status = s
 
     return status
--- a/tests/integration/build_fail_multi/bad/Dockerfile
+++ b/tests/integration/build_fail_multi/bad/Dockerfile
@@ -1,0 +1,3 @@
+FROM busybox
+
+RUN false
--- a/tests/integration/build_fail_multi/docker-compose.yml
+++ b/tests/integration/build_fail_multi/docker-compose.yml
@@ -1,0 +1,8 @@
+version: "3"
+services:
+  bad:
+    build:
+      context: bad
+  good:
+    build:
+      context: good
--- a/tests/integration/build_fail_multi/good/Dockerfile
+++ b/tests/integration/build_fail_multi/good/Dockerfile
@@ -1,0 +1,3 @@
+FROM busybox
+#ensure that this build finishes second so that it has a chance to overwrite the return code
+RUN sleep 0.5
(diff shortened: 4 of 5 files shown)


charliemirabile:
Not sure if this needs unit tests / how to test if the maintainers think it does.

p12tic:
I think an integration test could be handy.

Surprisingly, we test build command already. Probably https://github.com/containers/podman-compose/blob/main/tests/integration/build_fail/test_podman_compose_build_fail.py test could be enhanced

p12tic:
Other than this, this is great find, thanks for submitting a PR.

charliemirabile:
Ok, I have added an integration test. I fear that it isn't perfect, because while it will definitely pass 100% with the changes in this PR, it is possible that it could pass erroneously on the current main branch because it is basically relying on a race condition to surface the original bug (the image that fails needs to finish first and then the successful one finishes second so that the failing status is erased by the new status code of 0). I can't really think of any better way to test this - you could make the sleep command in the successful image arbitrarily long in order to increase the chances of it winning the race, but you can't ever get it 100% because the scheduler could just choose not to run the failing build, and the longer you make it, the longer the test takes. the 0.5s sleep is plenty on my maching to have a 100% failure rate on the main branch, but I have a reasonably fast computer.

charliemirabile:
I just reread you message more carefully
> I think an integration test could be handy.
> 
> Surprisingly, we test build command already. Probably https://github.com/containers/podman-compose/blob/main/tests/integration/build_fail/test_podman_compose_build_fail.py test could be enhanced to test this situation as well in separate test.
If you would prefer that I merge the new test I created in a folder `build_fail_multi` into the existing `build_fail` one, I can, though I am not sure if it would actually be cleaner.

p12tic:
Thanks for writing a test, looks great. Having test in separate directory or not is something not worth arguing, especially since the rest of tests don't have consistency in this regard. Regarding sleep, in this case it's the best way to solve the issue even if we have a race condition in the test. The integration test suite is slow anyway, so a second here or there does not matter much. If we start to care about integration test performance, better avenue would be to parallelize it.

p12tic (approved):

PR #1179, opened 2025-04-07
Fix port command

### LINKED ISSUE #778: podman-compose port should consult podman port command

As of today `podman-compose` consults the actual Compose YAML file to get information about ports and assumes that ports are expected to have direct mappings to host, like `8080:80` etc. Otherwise a `ValueError: not enough values to unpack (expected 2, got 1)` exception is [thrown after unpacking](https://github.com/containers/podman-compose/blob/bce40c2db30fb0ffb9264b5f51535c26f48fe983/podman_compose.py#L2615).

I presume this was the compromise of providing "at least something", while `podman` had issues with properly publishing exposed ports and supplying this information via `inspect <container>` command.

And starting with `podman 4.6.x` I suspect things can shape up for the better.

Latest versions of **Podman** (>=4.6.0) now correctly supplies all **published** ports information via `podman inspect <container>`. This allows `podman-compose` to take advantage of this, and provide this information as well. A "positive" side-effect of this change - `podman-compose` becomes more aligned with behavior and output of `docker-compose`, making switching from **Docker** to **Podman** even more appealing to the public (same goes for `alias docker-compose=postman-compose`).

Another benefit: additional tooling unlocks better support for **Podman**.
I can definitively say that with proposed change in place `testcontainers` for `python` (and mb `Java`, but I need to check this claim) will start functioning correctly when querying port information. For compose style containers they rely on information from [`(podman)docker-compose port <container> <port>`](https://github.com/testcontainers/testcontainers-python/blob/928af5a88393d8223a0d5a55e9378941826997ec/compose/testcontainers/compose/__init__.py#L171)

The only other alternative is to use `podman port` command to get this data. This approach does not apply or as easy to implement as soon as you rely on some other tool/abstraction (that has baked-in support for `*-compose` style commands) instead of direct communication with **Docker**/**Podman** clients/sockets: case in point - `testcontainers`.

**Additional context**

In case you're looking for analogous output from `docker-compose` I got you covered

**When container is running**
```sh
$ docker-compose port kong 9090
0.0.0.0:9090

$ docker port kong 9090
0.0.0.0:9090
[::]:9090
```
**When container is not started**
```sh
$ docker-compose port kong 9090
service "kong" is not running

$ docker port kong 9090
Error response from daemon: No such container: kong
```

### LINKED ISSUE #1039: podman-compose port does not work with dynamic host port
If a container does not have a specific host port mapping for its internal port it will be given a dynamic port on the host when it is started (this is useful if you want to have more than one instance of the container on the same host without conflicts).

This is the primary use case for the `podman port` command—precisely to query podman to figure out what random port your service received. While podman-compose does offer a similar command `podman-compose port` for this situation, it seems to only work on port mappings with a specific host port.

**Describe the bug**
If `podman-compose port` is run to query a port for a container that does not have a specific host mapping, podman-compose crashes with a value error because it tries to unpack the result of splitting a string expecting to find at least two pieces.

**To Reproduce**
Create a compose file with a service that has an internal port it listens on and add a port mapping that mentions the number of the internal port but does not specify a host port to map it to. Run podman-compose port with the name of the service and the number of the internal port.

**Expected behavior**
The mapping that podman chose for the internal port on the host is printed

**Actual behavior**
Python stack trace and crash output due to unhandled value error exception

**Output**

example compose file
```yml
services:
  example:
    image: busybox
    command: httpd -f -p 80 -h /tmp
    ports:
      - "80"
```
commands that demonstrate the issue
```
$ podman-compose up -d
2522d6eeed43ad011db9300d901101f53fe6cf30c3b7c97073ce7b90f32974be
Resolved "busybox" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/busybox:latest...
Getting image source signatures
Copying blob 3d1a87f2317d done   | 
Copying config 87ff76f62d done   | 
Writing manifest to image destination
cd58e63445c33df91487869478f7d2b352603ff3e665873dcceb44489d1b982c
foo_example_1
$ podman-compose port example 80
  File "podman_compose.py", line 3530, in <module>
    main()
  File "podman_compose.py", line 3526, in main
    asyncio.run(async_main())
  File "/usr/lib64/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/runners.py", line 118, in run
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 685, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "podman_compose.py", line 3522, in async_main
    await podman_compose.run()
  File "podman_compose.py", line 1755, in run
    retcode = await cmd(self, args)
              ^^^^^^^^^^^^^^^^^^^^^
  File "podman_compose.py", line 2921, in compose_port
    published, target = _published_target(p)
                        ^^^^^^^^^^^^^^^^^^^^
  File "podman_compose.py", line 2910, in _published_target
    published, target = port_string.split(":")[-2:]
    ^^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)

$ podman port foo_example_1 80
0.0.0.0:33737
```

**Environment:**
 - OS: Linux (Fedora 39)
 - podman version: 4.9.3
 - podman compose version:
```
$ git describe HEAD
v1.2.0-72-g229650c
$ git show --oneline
229650c (HEAD -> main, upstream/main) Merge pull request #1032 from shuckc/main
```

knarfS:
This PR fixes #778 and fixes #1039 by using `podman inspect` to get the actual host ports, which is necessary for dynamic host ports.

The tests now have asserts to check the command output.


tests/integration/ports/docker-compose.yml
```diff
@@ -6,9 +6,8 @@ services:
       command: ["dumb-init", "/bin/busybox", "httpd", "-f", "-h", "/var/www/html", "-p", "8001"]
       working_dir: /var/www/html
       ports:
-        - 8001:8001
-      volumes:
-        - ./test1.txt:/var/www/html/index.txt:ro,z
```
p12tic, line 11: I think that if we're removing custom index.txt, then we may as well replace httpd with something more obvious, like `sleep infinity`. Then `working_dir` can be removed as well. `down_cmd` no longer needs `--volumes`, but needs `-t0` to make test not take 10 seconds more.

p12tic (changes requested):
Very nice, thanks. Could you please add a release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like)?

knarfS:
Thanks for you review!

I've fixed your change request and also added the news fragment.

tests/integration/ports/docker-compose.yml
knarfS: I've simplified the compose file.

The `-t0` argument is not needed here, because `dump-init` is used as PID1 and so SIGTERM works when shutting down the container.

PR #1180, opened 2025-04-07
Add rmi argument for down command

### LINKED ISSUE #387: add "--rmi" option for podman-compose down
`docker compose down` has the `--rmi` option:
```
$ docker compose down --help
Usage:  docker compose down

Stop and remove containers, networks

Options:
      --remove-orphans    Remove containers for services not defined in the Compose file.
      --rmi string        Remove images used by services. "local" remove only images that don't have a custom tag ("local"|"all")
  -t, --timeout int       Specify a shutdown timeout in seconds (default 10)
  -v, --volumes volumes    Remove named volumes declared in the volumes section of the Compose file and anonymous volumes attached to containers.

### PR DESCRIPTION
**Pull request by knarfS**
This fixes #387 by implementing the `--rmi` argument (with `all` and `local`). This argument removes the images when stopping the services.

### CODE REVIEW

File: podman_compose.py [hunk 1]
```diff
@@ -2931,13 +2932,15 @@ async def compose_down(compose: PodmanCompose, args):
...
-                        "{{ .Names }}",
+                        "{{ .Image }} {{ .Names }}",
                     ],
                 )
             )
             .decode("utf-8")
             .splitlines()
         )
+        orphaned_images = [item.split()[0] for item in orphaned_containers]
+        names = {item.split()[1] for item in orphaned_containers}
```
p12tic, line 2943: Not sure if you really wanted to have set for names but not for orphaned_images. Needs a comment why.

tests/integration/up_down/test_podman_compose_up_down.py
```diff
@@ -0,0 +1,289 @@
...
+        "run",
+        podman_compose_path(),
+        "-f",
+        os.path.join(test_path(), "up_down", "docker-compose.yml"),
+        "up",
+        "-d",
+        "--force-recreate",
+    ]
+
+    container_exists_cmd = [
```
p12tic, line 30: These commands are simple enough that they can be inlined. Right now we make tests a bit harder to understand.

tests/integration/up_down/test_podman_compose_up_down.py
```diff
@@ -0,0 +1,289 @@
...
+        self.run_subprocess_assert_returncode(self.container_exists_cmd + ["up_down_web1_1"], 1)
+        self.run_subprocess_assert_returncode(self.container_exists_cmd + ["up_down_web2_1"])
+        self.run_subprocess_assert_returncode(self.volume_exists_cmd + ["up_down_web1_vol"], 1)
+        self.run_subprocess_assert_returncode(self.volume_exists_cmd + ["up_down_web2_vol"])
+        self.run_subprocess_assert_returncode(self.image_exists_cmd + ["podman-compose-test"])
+        self.run_subprocess_assert_returncode(
+            self.image_exists_cmd + ["docker.io/library/debian:bookworm-slim"]
+        )
+
+        # Cleanup orphant container
```
p12tic, line 138: typo

tests/integration/up_down/test_podman_compose_up_down.py
```diff
@@ -0,0 +1,289 @@
+# SPDX-License-Identifier: GPL-2.0
+
+"""
+test_podman_compose_up_down.py
+
+Tests the podman compose up and down commands used to create and remove services.
+"""
+
+# pylint: disable=redefined-outer-name
```
p12tic, line 9: Probably no longer needed?

tests/integration/up_down/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+volumes:
+    web1_vol:
+    web2_vol:
+services:
+    web1:
+      image: podman-compose-up-down-test
+      build: .
+      hostname: web1
+      command: ["dumb-init", "/bin/busybox", "httpd", "-f", "-h", "/var/www/html", "-p", "8001"]
```
p12tic, line 10: Better launch `sleep infinity` and not add unnecessary moving parts. Stop commands will need `-t0` in such case.

tests/integration/up_down/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM docker.io/library/debian:bookworm-slim
```
p12tic, line 1: Let's use docker.io/library/busybox as many tests already use that image.

p12tic (changes requested):
Some small comments.

tests/integration/up_down/Dockerfile
p12tic: Please disregard this comment. We want to move from busybox to nopush/podman-compose-test where possible and this will make `docker.io/library/debian:bookworm-slim` the primary image we pull.

podman_compose.py
knarfS: Good catch, thanks!

knarfS:
Fixed your change requests.

I've also added a news fragment.

p12tic (approved):

Issue #806, opened 2023-11-24
Add support of setting PID limit


candleindark:
**Is your feature request related to a problem? Please describe.**
The current version of `podman-compose` doesn't seem to support the setting of the PID limit of a service container. I was unable to set the PID limit of a service using either the [`pids_limit`](https://github.com/compose-spec/compose-spec/blob/master/05-services.md#pids_limit) element or the [`deploy.reservations.pids`](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#pids) element.

**Describe the solution you'd like**
I would like `podman-compose` to support the setting PID limit of a service container either through the [`pids_limit`](https://github.com/compose-spec/compose-spec/blob/master/05-services.md#pids_limit) element or the [`deploy.reservations.pids`](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#pids) element.

**Describe alternatives you've considered**
Currently, I use a separate bash script to call up `podman run` with the `--pids-limit` option to set the PID limit for one of the service containers within a project but use `podman-compose` to spawn other containers in the project. The `podman run` command is unseemly long.

**Additional context**
```
$ podman-compose --version
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.3.1
podman-compose version 1.0.6
podman --version
podman version 4.3.1
exit code: 0
```

rockymtnman:
Are there any plans to add this enhancement to support the "--pids-limit"?

DEvil0000:
`podman update --pids-limit -1 container` works but the docker compose options do both not work

offa:
As a workaround I use `--podman-run-args` to pass the limit to Podman:

```sh
podman-compose --podman-run-args="--pids-limit=<your limit>" run […]
```

rockymtnman:
It would be nice if there was a mechanism to provide unsupported arguments similar to "--podman-run-args" in the compose file itself.  Having these values separate from the compose at the command line makes maintainability harder than having everything in a configurable compose file.

p12tic:
Fixed in https://github.com/containers/podman-compose/pull/1182.

Issue #626, opened 2023-01-24
`podman-compose up` swallows exit code


bronson:
**Describe the bug**

Related to #586 but for `up` instead of `build`.

`podman-compose up` always returns a success result code, even when the command fails.

This is an issue, for example, in [this script](https://gitlab.com/bronson/veloren-serve/-/blob/c0bddb8a905b98f615ff1f259621eb04963686ea/resume#L40): if `podman-compose up` fails, the script should error out (because it's `set -e`) and _not_ run `podman-compose logs` on failed containers.

To reproduce, create this podman-compose.yml:

```yaml
version: "3.5"
services:
  container:
    image: this-image-is-SO-unusable
```

And run these commands:

```sh
sudo podman-compose up || echo FAILED
echo result code: $?
```

"FAILED" should be printed after the podman-compose command, and podman claims the exit code should be 125:

```txt
Error: repository name must be lowercase
exit code: 125
podman start -a tmp_container_1
Error: unable to start container cad728e9678a340b717ef681b6a806d68432b28fcf1c138bc02eb1219c99f44c: netavark: setns: IO error: Operation not permitted (os error 1)
exit code: 125
result code: 0
```

Instead, "FAILED" is _not_ printed, and the result code is 0: success!

```
$ podman-compose version
['podman', '--version', '']
using podman version: 4.4.0-rc2
podman-composer version  1.0.3
podman --version 
podman version 4.4.0-rc2
exit code: 0
```

**Environment:**
 - OS: Linux (Fedora Silverblue)

mluhmann:
In my opinion proper exit codes are needed for any podman-compose command. 
But in that case it need a complete rework of all commands relying on that 0 return value.
https://github.com/containers/podman-compose/issues/501#issuecomment-1160256390

EdwinKM:
Validated your bug and seems a issue.

But i only want to warn you for a mistake in the logic.
If you use this in your actual scripts you get a "0" in all cases.

```
sudo podman-compose up || echo FAILED
echo result code: $?
```
"$?" will show the exit code of the last command. If compose fails you get the exit code of "echo".


```
$ cat /does_not_exist 2>/dev/null; echo $?
1
$ cat /does_not_exist 2>/dev/null || echo "does not exist"; echo $?
does not exist
0
```

bronson:
Oops you’re absolutely right. Thanks for catching that!

PR #1181, opened 2025-04-08
Return non-zero exit_code on failure when doing `up -d`

zeyugao:
When run `podman-compose up -d`, it returns directly without handling the error.

Related issues:

https://github.com/containers/podman-compose/issues/806
https://github.com/containers/podman-compose/issues/626

pids_limit spec:

https://github.com/compose-spec/compose-spec/blob/main/05-services.md#pids_limit and https://github.com/compose-spec/compose-spec/blob/main/deploy.md#pids

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2879,11 +2879,15 @@ def deps_from_container(args, cnt):
 @cmd_run(podman_compose, "up", "Create and start the entire stack or some of its services")
 async def compose_up(compose: PodmanCompose, args):
     excluded = get_excluded(compose, args)
+
     if not args.no_build:
         # `podman build` does not cache, so don't always build
         build_args = argparse.Namespace(if_not_exists=(not args.build), **args.__dict__)
-        if await compose.commands["build"](compose, build_args) != 0:
+        build_exit_code = await compose.commands["build"](compose, build_args)
+        if build_exit_code != 0:
             log.error("Build command failed")
+            if not args.dry_run:
+                return build_exit_code
 
     hashes = (
         (
@@ -2911,19 +2915,29 @@ async def compose_up(compose: PodmanCompose, args):
     # args.no_recreate disables check for changes (which is not implemented)
 
     await create_pods(compose, args)
+    exit_code = 0
     for cnt in compose.containers:
         if cnt["_service"] in excluded:
             log.debug("** skipping: %s", cnt["name"])
             continue
         podman_args = await container_to_args(compose, cnt, detached=False, no_deps=args.no_deps)
-        subproc = await compose.podman.run([], "create", podman_args)
-        if not args.no_start and args.detach and subproc is not None:
-            await run_container(
+        subproc_exit_code = await compose.podman.run([], "create", podman_args)
+        if subproc_exit_code is not None and subproc_exit_code != 0:
+            exit_code = subproc_exit_code
+
+        if not args.no_start and args.detach and subproc_exit_code is not None:
+            container_exit_code = await run_container(
                 compose, cnt["name"], deps_from_container(args, cnt), ([], "start", [cnt["name"]])
             )
 
-    if args.no_start or args.detach or args.dry_run:
-        return
+            if container_exit_code is not None and container_exit_code != 0:
+                exit_code = container_exit_code
+
+    if args.dry_run:
+        return None
+    if args.no_start or args.detach:
+        return exit_code
+
     # TODO: handle already existing
     # TODO: if error creating do not enter loop
     # TODO: colors if sys.stdout.isatty()
--- a/newsfragments/handle-up-detach-exitcode.bugfix
+++ b/newsfragments/handle-up-detach-exitcode.bugfix
@@ -1,0 +1,1 @@
+- Return non-zero exit_code on failure when `up -d`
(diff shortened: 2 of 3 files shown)


zeyugao:
The unitest failed, but the `up -d` returns `stderr: Error: keep-id is only supported in rootless mode`. So it seems that the original unittest is failed unexpectedly

podman_compose.py
```diff
@@ -772,6 +772,27 @@ def container_to_cpu_res_args(cnt, podman_args):
             str(mem_res).lower(),
         ))
 
+    # Handle pids limit from both container level and deploy section
+    pids_limit = cnt.get("pids_limit")
+    deploy_pids = None
```
p12tic, line 777: `limits.get("pids")` would be equivalent, no?

podman_compose.py
```diff
@@ -772,6 +772,27 @@ def container_to_cpu_res_args(cnt, podman_args):
...
+
+    # Ensure consistency between pids_limit and deploy.resources.limits.pids
+    if pids_limit is not None and deploy_pids is not None:
+        if str(pids_limit) != str(deploy_pids):
+            raise ValueError(
+                f"Inconsistent PIDs limit: pids_limit ({pids_limit}) and "
+                f"deploy.resources.limits.pids ({deploy_pids}) must be the same"
+            )
+
+    # Use whichever value is set (prioritizing pids_limit if both are set)
```
p12tic, line 791: `(prioritizing pids_limit if both are set)` is invalid comment, because due to above exception this situation can't happen.

podman_compose.py
```diff
@@ -2760,11 +2760,16 @@ def deps_from_container(args, cnt):
 @cmd_run(podman_compose, "up", "Create and start the entire stack or some of its services")
 async def compose_up(compose: PodmanCompose, args):
     excluded = get_excluded(compose, args)
+    exit_code = 0
+    
     if not args.no_build:
         # `podman build` does not cache, so don't always build
         build_args = argparse.Namespace(if_not_exists=(not args.build), **args.__dict__)
-        if await compose.commands["build"](compose, build_args) != 0:
+        build_result = await compose.commands["build"](compose, build_args)
```
p12tic, line 2789: Let's name build_exit_code

podman_compose.py
```diff
@@ -2802,12 +2807,21 @@ async def compose_up(compose: PodmanCompose, args):
             compose, cnt, detached=args.detach, no_deps=args.no_deps
         )
         subproc = await compose.podman.run([], podman_command, podman_args)
+        if subproc is not None and subproc != 0:
```
p12tic, line 2831: Let's name subproc_exit_code

podman_compose.py
```diff
@@ -2802,12 +2807,21 @@ async def compose_up(compose: PodmanCompose, args):
...
+            exit_code = subproc
+        
         if podman_command == "run" and subproc is not None:
-            await run_container(
+            container_result = await run_container(
```
p12tic, line 2835: Let's name container_exit_code

podman_compose.py
```diff
@@ -2760,11 +2760,16 @@ def deps_from_container(args, cnt):
...
+    
     if not args.no_build:
         # `podman build` does not cache, so don't always build
         build_args = argparse.Namespace(if_not_exists=(not args.build), **args.__dict__)
-        if await compose.commands["build"](compose, build_args) != 0:
+        build_result = await compose.commands["build"](compose, build_args)
+        if build_result != 0:
             log.error("Build command failed")
+            if not args.dry_run:
+                return build_result
```
p12tic, line 2793: This return needs a separate release note, because it fixes unrelated issue that podman-compose would continue even if build step fails.

p12tic (changes requested):
Looks good overall.

Please add release notes to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like).

p12tic:
I think it's worth moving pids_limit change to separate PR, because it will take some time to understand why tests fail for your `up -d` changes. Also they will need an integration test, so splitting PRs will not delay the pids_limit part.

podman_compose.py
zeyugao: Yes, definitely

zeyugao:
https://github.com/containers/podman-compose/pull/1182

p12tic:
Tests are currently failing, do they fail for you locally?

zeyugao:
image

I think that the failure reason is mentioned before. Locally I use rootless podman, but in the github action, podman is run as root. In this unit test expects using keep-id which is only supported in rootless mode as shown in the stderr

https://github.com/containers/podman-compose/blob/main/tests/integration/in_pod/custom_x-podman_false/docker-compose.yml#L5

p12tic:
> I think that the failure reason is mentioned before. Locally I use rootless podman, but in the github action, podman is run as root. In this unit test expects using keep-id which is only supported in rootless mode as shown in the stderr

@zeyugao Sounds fair, thanks for explaining. Could we detect root by e.g. doing `def is_root(): return os.geteuid() == 0` and then adjusting expected return code based on that? This would make integration test work in both rootful and rootless mode.

zeyugao:
I have considered it as a solution, but this will make the CI on GitHub Actions unable to test the functionality correctly, making it less usefully. We have to test it locally.

zeyugao:
I have changed the expected exit code accordingly

p12tic (approved):
Looks great, thank you!

p12tic:
I rebased the PR and fixed merge conflict.

Issue #1178, opened 2025-04-07
Podman Compose 1.3.0 `up -d` command never returns/finishes/ends


ArthoPacini:
**Describe the bug**
When running `podman-compose up -d` the command never returns, it stalls / freezes

**To Reproduce**
1. `docker-compose.yaml`:

```yaml
services:
  service_a:
    image: docker.io/crccheck/hello-world
    container_name: "service_a"
    depends_on:
      service_b:
        condition: service_healthy
    healthcheck:
      test: ["CMD", "wget", "--quiet", "--tries=1", "--spider", "http://localhost:8000"]
      interval: 10s
      start_interval: 15s
      timeout: 5s
      retries: 3

  service_b:
    image: docker.io/crccheck/hello-world
    container_name: "service_b"
    healthcheck:
      test: ["CMD", "wget", "--quiet", "--tries=1", "--spider", "http://localhost:8000"]
      interval: 10s
      start_interval: 15s
      timeout: 5s
      retries: 3

networks:
  default:
    driver: bridge
```

2. Run `docker-compose up -d`:

```bash
podman-compose up -d
9a61f6744544dd5343a3e497579e2aadfbac156936d735e89749d95b520832f2
8d43c893be1ddbda564692e8e7e28f27f8a04c559ac1d38a121cf6b6736e4d29
7a350b27a718f4bda86977f51ef32ae75da0b78c38b7d5c84a1e3a62ad767254
```

**Expected behavior**
The command should end

**Actual behavior**
It requires `CTRL-C` to be called to terminate the program in which it outputs:

```bash
^CTraceback (most recent call last):
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/bin/podman-compose", line 3707, in async_main
    await podman_compose.run()
  File "/usr/local/bin/podman-compose", line 1861, in run
    retcode = await cmd(self, args)
              ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/bin/podman-compose", line 2705, in compose_up
    await run_container(compose, cnt["name"], cnt["_deps"], ([], "start", [cnt["name"]]))
[...]

PR #1184, opened 2025-04-14
Fix service_healthy condition enforcing

### LINKED ISSUE #1176: Detached mode executes entrypoint twice
**Describe the bug**
`podman-compose up -d` executes the container entrypoint twice.

**To Reproduce**
Dockerfile
```
FROM docker.io/debian:bookworm

COPY ./init.sh /root/
```

compose.yaml
```
services:
  test:
    image: test-image:latest
    container_name: test
    build:
      context: .
      dockerfile: Dockerfile
    entrypoint: ["/bin/bash", "/root/init.sh"]
```

init.sh
```
#!/bin/bash
file="/root/store.txt"

if [ ! -f $file ]; then
  echo 1 > $file
  value=1
else
  value=$(cat $file)
  value=$((value + 1))
fi

echo "$value"
```

**Expected behavior**
`podman logs test` should output:
```
1
```

**Actual behavior**
`podman logs test` outputs:
```
1
2
```

**Environment:**
 - OS: Fedora 41
 - podman version: 5.4.1 
 - podman compose version: 1.3.0

**Additional context**

`podman compose up` behaves as expected:

`podman logs test` outputs:
```
1
```

### LINKED ISSUE #1183: podman-compose fails to enforce the `service_healthy` condition
**Describe the bug**
Previously, podman-compose doesn't handle conditions of dependency services properly as reported
by [Fails to handle the service_healthy condition of a depends_on element][1] and [compose fails
with depends_on:service_healthy with various errors][2]. As of podman-compose 1.3.0, with the merge
of [PR #1082 Provide support for conditional dependencies][3], podman-compose supports various
dependency conditions and enforces the constraints. Specifically, it uses the `run` method to create
and start the containers w/o check dependency conditions in first pass. Then it calls the
`run_container()` where the container is started only if the conditions of its dependencies are
satisfied. This implementation is problematic with a few issues:

- Containers with dependencies that should enter into healthy state are started prematurely,
resulting in potential startup failure of the containers that tightly coupled with their
dependencies
- The `check_dep_conditions()` function doesn't take into account the fact that podman prior to
4.6.0 doesn't support --condition=healthy, leading to infinite loop in this function
- A third problem is that podman-compose attempts to stop and remove the containers defined in the
compose file when the `--force-recreate` option is specified when there are no running containers at
all.

An example is included in this issue to help reproduce and debug the aforementioned problem. This
example deploys a Java SpringBoot application backed by MySQL, which can only startup properly when
the MySQL database is ready to connect. The current implementation `run` the MySQL and Java
containers at the same time. Therefore, the Java application fails to startup in its first few
attempts. The example files are attached in [reproduce.zip](https://github.com/user-attachments/files/19736925/reproduce.zip) and organized as follows:

> .
├── docker-compose.yml
└── provision
    ├── mariadb
    │   ├── conf.d
    │   │   └── my.cnf
    │   └── entrypoint-initdb.d
    │       ├── nacos.sql
    └── nacos
        └── init.d
            └── custom.properties

**To Reproduce**
Steps to reproduce the startup failure problem:
1. unzip the attachment to a directory say podman-pr
2. install podman-compose 1.3.0, and its dependencies such as podman
3. run "podman-compose up -d --force-recreate" under podman-pr folder

**Expected behavior**
Both the nacos and mariadb containers start successfully and enter into the healthy status 

**Actual behavior**
The mariadb container starts successfully and enters into the healthy status. However, the nacos container stuck in the starting status. 

**Output**

```
$  podman-compose version
podman-compose version 1.3.0
podman version 5.4.1


$  podman-compose up -d --force-recreate

...
podman-pr_mariadb_1
WARN[0010] StopSignal SIGTERM failed to stop container podman-pr_nacos_1 in 10 seconds, resorting to SIGKILL
podman-pr_nacos_1
podman-pr_nacos_1
podman-pr_mariadb_1
5a637e579a6e7971f9d41a40a133d39e2308406c791254027a7205d6fbee9950
podman-pr_default
1cc8f2da09a14b5fdceed77e4fe813540f6e6fb44c38bcd6ba61b58a7079449b
b8f6d7580c96b48724512b6b25c4d37c868427da862bf6f0e49a6a3a74b51831
podman-pr_mariadb_1
23cf598b1522622a9da8887aa7067f3c2abfca52ef01671f51dfd3cb0b023e4f
podman-pr_nacos_1
```

**Environment:**
 - OS: Linux (5.15.179-1-MANJARO)
 - podman-compose version 1.3.0
 - podman version 5.4.1

**To Reproduce the improper handling of --force-recreate**
1. unzip the attachment to a directory say podman-pr
2. install podman-compose 1.3.0, and its dependencies such as podman
3. run "podman-compose down" under podman-pr folder
4. run "podman-compose up -d --force-recreate" under podman-pr folder

**Expected behavior**
There should be no Error complaining no container found

**Actual behavior**
There are a few confusing error messages stating no container found

**Output**

```
$  /usr/bin/podman-compose up down
$  /usr/bin/podman-compose up -d --force-recreate
Error: no container with name or ID "podman-pr_nacos_1" found: no such container
Error: no container with name or ID "podman-pr_mariadb_1" found: no such container
Error: no container with ID or name "podman-pr_nacos_1" found: no such container
Error: no container with ID or name "podman-pr_mariadb_1" found: no such container
Error: no pod with name or ID pod_podman-pr found: no such pod
adf1669c9bd84744cc9ab4675f005fa73870e0184c5081ec08cf149311c31fa9
dfecff55ce080c38c23f507589dc360e8f83e24fc372c868334f4d2107454144
podman-pr_mariadb_1
005c3cba1953b78d44d4aab5069829e9be2508e7dd139d43485e919302f2e0b4
podman-pr_nacos_1
```

**Environment:**
 - OS: Linux (5.15.179-1-MANJARO)
 - podman-compose version 1.3.0
 - podman version 5.4.1

**To Reproduce the improper check_dep_conditions()**
1. find a debian 12 environment
2. unzip the attachment to a directory say podman-pr
3. install podman-compose 1.3.0, and its dependencies such as podman 4.3.1
4. run "podman-compose up -d" under podman-pr folder

**Expected behavior**
Both the nacos and mariadb containers start successfully and enter into the healthy status 

**Actual behavior**
The podman-compose up command hangs forever

**Output**

```
$  (base) vagrant@debian12:/work/podman-pr$ podman-compose up -d                        
d437edf388aa8bcca550a9b41e949d498376d37f6912c71d4f033a02cd4db7fe
6fb09531ded7fe6316064cd4601e164baa986a679efd6180e7f2c0fa025a5a84
2dde6b30fbf111164be18325812005a5b20a55a2fe50fd87928ed03af5b6dd06

(podman-compose hangs forever here...)

(open a separate terminal and type the following command)
(base) vagrant@debian12:/work/podman-pr$ watch -n 2 podman-compose ps
Every 2.0s: podman-compose ps                                    gauss: 22时48分24秒
                                                                       in 0.157s (0)
CONTAINER ID  IMAGE                                          COMMAND     CREATED
     STATUS                     PORTS
                    NAMES
dfecff55ce08  docker.io/schnell18/mariadb:10.7-focal-2       mariadbd    31 minutes
ago  Up 31 minutes (healthy)    0.0.0.0:3306->3306/tcp
                    podman-pr_mariadb_1
005c3cba1953  docker.io/schnell18/nacos-server:1.4.2-alpine              31 minutes
ago  Up 31 minutes (unhealthy)  0.0.0.0:8848->8848/tcp, 0.0.0.0:9555->9555/tcp, 0.0.
0.0:9848->9848/tcp  podman-pr_nacos_1
```

**Environment:**
 - OS: Linux (Debian 6.1.69-1 x86_64 GNU/Linux)
 - podman-compose version 1.3.0
 - podman version  4.3.1

**Additional information**
The check_dep_conditions() problem can also be reproduced by removing the `--no-deps` argument in the `test_run_nodeps()` and running the integration test by pushing to github. The `Tests` workflow will freeze forever. Here is a similar sample https://github.com/schnell18/podman-compose/actions/runs/14435019500/job/40474677325

[1]: https://github.com/containers/podman-compose/issues/866
[2]: https://github.com/containers/podman-compose/issues/1119
[3]: https://github.com/containers/podman-compose/pull/1082

schnell18:
This PR fixes #1176, #1178, and #1183 by employing a create-and-start approach, where the containers are created in the
first pass, then they are started using the `run_container()` method to make sure the dependencies'
conditions are checked. The second improvement is to add a version check to skip "podman wait
--condition=healthy" in the `check_dep_conditions()` function to prevent podman-compose hang. BTW,
this PR also fixes a minor problem that podman-compose attempts to stop and remove the containers
defined in the compose file when the `--force-recreate` option is specified when there are no
running containers at all.

Specific changes are as follows:
- Change compose-up to create then start container to enforce dependency condition check
- Skip running compose-down when there are no active containers
- Skip dependency health check to avoid compose-up hang for podman prior to 4.6.0, which doesn't support --condition healthy
- Add relevant integration test case and run the healthy state validation only when systemd is available
- Improve robustness for network, ulimit, extends etctest cases
- Relax pylint rules for test code, disable duplicate code check

4.6.0 seems to be the first version to support --condition=healthy, as discovered by this script:

~~~~bash
#!/bin/bash

# find supported wait conditions since v4.3.1 which is podman version used in
# github actions of podman-compose project
for ver in $(git tag -l | sed -e '/-rc[0-9]\+$/d' -e '/^v[0-3]/d' -e '/^v4.[0-2]/d' -e '/v4.3.0/d'); do
  echo $ver
  git show $ver:cmd/podman/common/completion.go | grep -A 3 "func AutocompleteWaitCondition"
done
~~~~

Finally, there are a few integration test cases mentioning this bug: 
- https://github.com/containers/podman-compose/blob/342a39dcfe6944932561bf12150deeb687fb6c52/tests/integration/env-file-tests/test_podman_compose_env_file.py#L129
- https://github.com/containers/podman-compose/blob/342a39dcfe6944932561bf12150deeb687fb6c52/tests/integration/ulimit/test_podman_compose_ulimit.py#L37

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2815,6 +2815,18 @@ async def check_dep_conditions(compose: PodmanCompose, deps: set) -> None:
         deps_cd = []
         for d in deps:
             if d.condition == condition:
+                if (
+                    d.condition
+                    in (ServiceDependencyCondition.HEALTHY, ServiceDependencyCondition.UNHEALTHY)
+                ) and strverscmp_lt(compose.podman_version, "4.6.0"):
+                    log.warning(
+                        "Ignored %s condition check due to podman %s doesn't support %s!",
+                        d.name,
+                        compose.podman_version,
+                        condition.value,
+                    )
+                    continue
+
                 deps_cd.extend(compose.container_names_by_service[d.name])
 
         if deps_cd:
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2891,7 +2891,7 @@ async def compose_up(compose: PodmanCompose, args):
         .splitlines()
     )
     diff_hashes = [i for i in hashes if i and i != compose.yaml_hash]
-    if args.force_recreate or len(diff_hashes):
+    if (args.force_recreate and len(hashes) > 0) or len(diff_hashes):
         log.info("recreating: ...")
         down_args = argparse.Namespace(**dict(args.__dict__, volumes=False, rmi=None))
         await compose.commands["down"](compose, down_args)
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2898,21 +2898,18 @@ async def compose_up(compose: PodmanCompose, args):
         log.info("recreating: done\n\n")
     # args.no_recreate disables check for changes (which is not implemented)
 
-    podman_command = "run" if args.detach and not args.no_start else "create"
-
     await create_pods(compose, args)
     for cnt in compose.containers:
         if cnt["_service"] in excluded:
             log.debug("** skipping: %s", cnt["name"])
             continue
-        podman_args = await container_to_args(
-            compose, cnt, detached=args.detach, no_deps=args.no_deps
-        )
-        subproc = await compose.podman.run([], podman_command, podman_args)
-        if podman_command == "run" and subproc is not None:
+        podman_args = await container_to_args(compose, cnt, detached=False, no_deps=args.no_deps)
+        subproc = await compose.podman.run([], "create", podman_args)
+        if not args.no_start and args.detach and subproc is not None:
             await run_container(
                 compose, cnt["name"], deps_from_container(args, cnt), ([], "start", [cnt["name"]])
             )
+
     if args.no_start or args.detach or args.dry_run:
         return
     # TODO: handle already existing
--- a/tests/integration/deps/test_podman_compose_deps.py
+++ b/tests/integration/deps/test_podman_compose_deps.py
@@ -2,7 +2,9 @@
 import os
 import unittest
 
+from tests.integration.test_utils import PodmanAwareRunSubprocessMixin
 from tests.integration.test_utils import RunSubprocessMixin
+from tests.integration.test_utils import is_systemd_available
 from tests.integration.test_utils import podman_compose_path
 from tests.integration.test_utils import test_path
 
@@ -181,5 +183,84 @@ class TestComposeConditionalDeps(unittest.TestCase, RunSubprocessMixin):
                 podman_compose_path(),
                 "-f",
                 compose_yaml_path(suffix),
+                "down",
+            ])
+
+
+class TestComposeConditionalDepsHealthy(unittest.TestCase, PodmanAwareRunSubprocessMixin):
+    def setUp(self):
+        self.podman_version = self.retrieve_podman_version()
+
+    def test_up_deps_healthy(self):
+        suffix = "-conditional-healthy"
+        try:
+            self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_yaml_path(suffix),
+                "up",
+                "sleep",
+                "--detach",
+            ])
+
+            # Since the command `podman wait --condition=healthy` is invalid prior to 4.6.0,
+            # we only validate healthy status for podman 4.6.0+, which won't be tested in the
+            # CI pipeline of the podman-compose project where podman 4.3.1 is employed.
+            podman_ver_major, podman_ver_minor, podman_ver_patch = self.podman_version
+            if podman_ver_major >= 4 and podman_ver_minor >= 6 and podman_ver_patch >= 0:
+                self.run_subprocess_assert_returncode([
+                    "podman",
+                    "wait",
+                    "--condition=running",
+                    "deps_web_1",
+                    "deps_sleep_1",
+                ])
+
+            # check both web and sleep are running
+            output, _ = self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_yaml_path(),
+                "ps",
+                "--format",
+                "{{.ID}}\t{{.Names}}\t{{.Status}}\t{{.StartedAt}}",
+            ])
+
+            # extract container id of web
+            decoded_out = output.decode('utf-8')
+            lines = decoded_out.split("\n")
+
+            web_lines = [line for line in lines if "web" in line]
+            self.assertTrue(web_lines)
+            self.assertEqual(1, len(web_lines))
+            web_cnt_id, web_cnt_name, web_cnt_status, web_cnt_started = web_lines[0].split("\t")
+            self.assertNotEqual("", web_cnt_id)
+            self.assertEqual("deps_web_1", web_cnt_name)
+
+            sleep_lines = [line for line in lines if "sleep" in line]
+            self.assertTrue(sleep_lines)
+            self.assertEqual(1, len(sleep_lines))
+            sleep_cnt_id, sleep_cnt_name, _, sleep_cnt_started = sleep_lines[0].split("\t")
+            self.assertNotEqual("", sleep_cnt_id)
+            self.assertEqual("deps_sleep_1", sleep_cnt_name)
+
+            # When test case is executed inside container like github actions, the absence of
+            # systemd prevents health check from working properly, resulting in failure to
+            # transit to healthy state. As a result, we only assert the `healthy` state where
+            # systemd is functioning.
+            if (
+                is_systemd_available()
+                and podman_ver_major >= 4
+                and podman_ver_minor >= 6
+                and podman_ver_patch >= 0
+            ):
+                self.assertIn("healthy", web_cnt_status)
+            self.assertGreaterEqual(int(sleep_cnt_started), int(web_cnt_started))
+
+        finally:
+            self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_yaml_path(),
                 "down",
             ])
--- a/tests/integration/deps/test_podman_compose_deps.py
+++ b/tests/integration/deps/test_podman_compose_deps.py
@@ -16,7 +16,7 @@ def compose_yaml_path(suffix=""):
 class TestComposeBaseDeps(unittest.TestCase, RunSubprocessMixin):
     def test_deps(self):
         try:
-            output, error = self.run_subprocess_assert_returncode([
+            output, _ = self.run_subprocess_assert_returncode([
                 podman_compose_path(),
                 "-f",
                 compose_yaml_path(),
@@ -39,7 +39,7 @@ class TestComposeBaseDeps(unittest.TestCase, RunSubprocessMixin):
 
     def test_run_nodeps(self):
         try:
-            output, error = self.run_subprocess_assert_returncode([
+            output, _ = self.run_subprocess_assert_returncode([
                 podman_compose_path(),
                 "-f",
                 compose_yaml_path(),
@@ -73,7 +73,7 @@ class TestComposeBaseDeps(unittest.TestCase, RunSubprocessMixin):
                 "--detach",
                 "sleep",
             ])
-            output, error = self.run_subprocess_assert_returncode([
+            output, _ = self.run_subprocess_assert_returncode([
                 podman_compose_path(),
                 "-f",
                 compose_yaml_path(),
@@ -146,7 +146,7 @@ class TestComposeConditionalDeps(unittest.TestCase, RunSubprocessMixin):
     def test_deps_succeeds(self):
         suffix = "-conditional-succeeds"
         try:
-            output, error = self.run_subprocess_assert_returncode([
+            output, _ = self.run_subprocess_assert_returncode([
                 podman_compose_path(),
                 "-f",
                 compose_yaml_path(suffix),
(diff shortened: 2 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -23,6 +23,20 @@ repos:
             "-sn", # Don't display the score
             "--rcfile=.pylintrc", # Link to your config file
           ]
+        files: ^[^/]+\.py$
+      # Hook for test code with different config
```
p12tic, line 27: What's the reason for the different config? If existing code is badly formatted we could reformat it.

tests/integration/deps/test_podman_compose_deps.py
```diff
@@ -183,3 +185,74 @@ def test_deps_fails(self):
...
+            web_cnt_started = ""
+            sleep_cnt_id = ""
+            sleep_cnt_name = ""
+            sleep_cnt_started = ""
+            for line in decoded_out.split("\n"):
+                if "web" in line:
+                    web_cnt_id, web_cnt_name, web_cnt_status, web_cnt_started = line.split("\t")
+                if "sleep" in line:
+                    sleep_cnt_id, sleep_cnt_name, _, sleep_cnt_started = line.split("\t")
+            self.assertNotEqual("", web_cnt_id)
```
p12tic, line 239: If asserts lived in the loop then the code would be way simplier.

tests/integration/deps/test_podman_compose_deps.py
```diff
@@ -183,3 +185,74 @@ def test_deps_fails(self):
...
+                compose_yaml_path(suffix),
+                "up",
+                "sleep",
+                "--detach",
+            ])
+
+            # The `podman wait --condition=healthy` is invalid prior to 4.6.0.
+            # Since the podman-compose project uses podman 4.3.1 in github actions, we
+            # use sleep as workaround to wait until the `sleep` container becomes running.
+            time.sleep(3)
```
p12tic, line 206: I think better idea would be to use polling. Right now all tests have been slowed down by 3 seconds. Additionally, it makes sense to check podman version and do proper test if the version is >= 4.6.0.

tests/integration/nets_test1/test_podman_compose_nets_test1.py
```diff
@@ -59,9 +59,13 @@ def test_nets_test1(self):
...
             # check if Host port is the same as provided by the service port
+            self.assertIsNotNone(container_info['NetworkSettings']["Ports"].get("8001/tcp", None))
+            self.assertGreater(len(container_info['NetworkSettings']["Ports"]["8001/tcp"]), 0)
+            self.assertIsNotNone(
+                container_info['NetworkSettings']["Ports"]["8001/tcp"][0].get("HostPort", None)
+            )
             self.assertEqual(
-                container_info['NetworkSettings']["Ports"],
-                {"8001/tcp": [{"HostIp": "", "HostPort": "8001"}]},
+                container_info['NetworkSettings']["Ports"]["8001/tcp"][0]["HostPort"], "8001"
```
p12tic, line 68: Why this change? Did the test start failing? The new version checks less things with more code.

tests/integration/nets_test1/test_podman_compose_nets_test1.py
```diff
@@ -77,9 +81,13 @@ def test_nets_test1(self):
...
 
+            self.assertIsNotNone(container_info['NetworkSettings']["Ports"].get("8001/tcp", None))
+            self.assertGreater(len(container_info['NetworkSettings']["Ports"]["8001/tcp"]), 0)
+            self.assertIsNotNone(
+                container_info['NetworkSettings']["Ports"]["8001/tcp"][0].get("HostPort", None)
+            )
             self.assertEqual(
-                container_info['NetworkSettings']["Ports"],
-                {"8001/tcp": [{"HostIp": "", "HostPort": "8002"}]},
+                container_info['NetworkSettings']["Ports"]["8001/tcp"][0]["HostPort"], "8002"
```
p12tic, line 90: Same here.

podman_compose.py
```diff
@@ -2784,28 +2796,25 @@ async def compose_up(compose: PodmanCompose, args):
...
             log.debug("** skipping: %s", cnt["name"])
             continue
-        podman_args = await container_to_args(
-            compose, cnt, detached=args.detach, no_deps=args.no_deps
-        )
-        subproc = await compose.podman.run([], podman_command, podman_args)
-        if podman_command == "run" and subproc is not None:
+        podman_args = await container_to_args(compose, cnt, detached=False, no_deps=args.no_deps)
+        subproc = await compose.podman.run([], "create", podman_args)
+        if not args.no_start and args.detach and subproc is not None:
```
p12tic, line 2813: This should be extracted to a separate PR, because if something breaks I will have easier time bisecting.

p12tic (changes requested):
Overall looks good, thank you very much. I had several comments.

p12tic:
Please add release note to the [newsfragments](https://github.com/containers/podman-compose/tree/main/newsfragments) directory (you can look [here](https://github.com/containers/podman-compose/commit/6de335beb295d8111f52053d75c4051f9f89b7be) for inspiration on how release note looks like).

.pre-commit-config.yaml
schnell18: The main idea of this change is to relax the lint conformance requirements for the unit and integration test programs so that it can pass the pre-commit check without having to fix a bunch of lint issues summarized as follows:
>   19 duplicate-code
     12 unused-variable
      9 redefined-builtin
      7 unused-argument
      7 missing-timeout
      4 invalid-name
      2 use-maxsplit-arg
      1 unspecified-encoding
      1 unnecessary-pass
      1 too-many-public-methods
      1 subprocess-run-check
      1 raising-bad-type
      1 pointless-string-statement

 While not everyone may agree, I think the test code might not require same lint rules as the production code. For example, having a few lines of duplicate code makes creating test easier without having to refactor existing test cases and extract common code into a shared method when you borrow some code from previous tests. 
Creating a separate pylint config for test code is also a programatic approach which is better than the current approach where no pylint check are performed in the CI pipeline. The pylint check is only optional triggered as a pre-commit hook that can be bypassed. So by adopting a relaxed config separately, we can enforce the pylint check in the CI pipeline and grandually tighten the lint rules for test code.

This change is not strictly necessary for this PR. I will cancel the pylint config change and raise a new issue for the discussion of the test code lint issues.

BTW, I attached a detailed pylint output for your reference. 
[podman-compose-ut-pylint-issues.txt](https://github.com/user-attachments/files/19841919/podman-compose-ut-pylint-issues.txt)

You should be able to reproduce it by running:

    pre-commit run pylint --all-files

tests/integration/nets_test1/test_podman_compose_nets_test1.py
schnell18: The original code's intention is just to validate the port as inidated by the comments:

> # check if Host port is the same as provided by the service port

The the test did fail on my laptop w/ podman 5.4.1. So I believe just focus on verifying service port is more appropriate. The additional code is to ensure the robustness of the test case and fail in an controlled way.

tests/integration/deps/test_podman_compose_deps.py
schnell18: Agree, this testcase has been branched by podman version.

tests/integration/nets_test1/test_podman_compose_nets_test1.py
schnell18: See previous comment

podman_compose.py
schnell18: Well, this block of change is crutial to the issue #1183. It is vital to depends_on condition check by using `create` rather than `run` to make a container in the first pass. If we `run` the container in the first place, the depends_on check is bypassed. This change also fixes the double-execution problem (#1176) in detached mode. So this change just happens to resolve multiple issues.  I hope this addresses your concern.

tests/integration/deps/test_podman_compose_deps.py
schnell18: Moving asserts into the loop isn't quite right. I simplify the code in other way by removing the loop and use list comprehension.

schnell18:
@p12tic Thanks for your insightful reviews. You comments have been addressed with the amended commit. Could you please take another look at your earliest convenience?

Tons of thanks!

p12tic (approved):
Amazing works, thanks!

p12tic:
I split the PR into smaller commits so that I myself would understand better what part causes what change of behavior and for that to be encoded into git history.

schnell18:
@p12tic 

> I split the PR into smaller commits so that I myself would understand better what part causes what change of behavior and for that to be encoded into git history.
> 
> Authorship has been preserved.

Thank you so much for the merge and the split!

PR #1190, opened 2025-04-30
Add `--abort-on-container-failure` option

### LINKED ISSUE #1164: Support for `--abort-on-container-failure` (to match `docker compose`)

Using compose to run containers that are expected to have a limited lifespan (e.g. running multiple test environments in parallel) doesn't currently have a mechanism to detect if one of those containers fails. The existing `--abort-on-container-exit` flag will abort everything even when a container finishes successfully - makes sense for server tasks that run indefinitely, but doesn't work when individual containers are expected to exit when their tests have finished. 

See https://github.com/docker/compose/issues/10225 for discussion around the same issue, with `--abort-on-container-failure` being added as the fix. 

Add `--abort-on-container-failure`, to stop all containers only if a container exits with a nonzero code.

gtebbutt:
Fixes #1164, with behaviour matching https://github.com/docker/compose/pull/11680 (see discussion in https://github.com/docker/compose/issues/10225)

### DIFF
--- a/tests/integration/abort/test_podman_compose_abort.py
+++ b/tests/integration/abort/test_podman_compose_abort.py
@@ -1,0 +1,46 @@
+# SPDX-License-Identifier: GPL-2.0
+
+import os
+import unittest
+
+from parameterized import parameterized
+
+from tests.integration.test_utils import RunSubprocessMixin
+from tests.integration.test_utils import podman_compose_path
+from tests.integration.test_utils import test_path
+
+
+def compose_yaml_path(failure_order):
+    return os.path.join(test_path(), "abort", f"docker-compose-fail-{failure_order}.yaml")
+
+
+class TestComposeAbort(unittest.TestCase, RunSubprocessMixin):
+    @parameterized.expand([
+        ("exit", "first", 0),
+        ("failure", "first", 1),
+        ("exit", "second", 0),
+        ("failure", "second", 1),
+        ("exit", "simultaneous", 0),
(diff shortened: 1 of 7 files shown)


tests/integration/abort/test_podman_compose_abort.py
```diff
@@ -0,0 +1,72 @@
...
+        "-f",
+        compose_yaml_path(failure_order),
+        "up",
+        f"--abort-on-container-{abort_type}",
+    ]
+
+
+class TestComposeAbort(unittest.TestCase, RunSubprocessMixin):
+    def test_abort_on_exit_fail_first(self):
+        self.run_subprocess_assert_returncode(
```
p12tic, line 27: You need to shutdown the containers with `compose down` after the test.

Also, consider using parameterized. Pass the scenario name as the first argument `name` and don't use it.

see e.g. tests/integration/lifetime/test_lifetime.py

p12tic (changes requested):
Overall looks good, had a comment about tests. Thanks a lot!

tests/integration/abort/test_podman_compose_abort.py
gtebbutt: Updated, thanks for the notes. I skipped the `name` parameter because the parameters seemed fairly self explanatory once they were written out, but happy to add that explicitly if you think it's helpful.

p12tic:
I combined commits into one because later commits fix the first commit.

p12tic (approved):
Looks great, thanks!

PR #1210, opened 2025-05-18
Add relabel option to secrets

jarovo:
On selinux enabled system, the secrets cannot be read without proper relabeling or correct policy being set.

This patch enables user to instruc podman-copose to use :z or :Z mount options to make podman relabel the file under bind-mount.

More info here:
https://unix.stackexchange.com/questions/728801/host-wide-consequences-of-setting-selinux-z-z-option-on-container-bind-mounts?rq=1

## Contributor Checklist:


p12tic:
Is this part of compose spec? Seems that no. In such case it should be named with `x-podman.` prefix and documentation added to docs/Extensions.md. I also don't think that `relabel` is clear name, though I don't have better suggestions at this time.

jarovo:
> Is this part of compose spec? Seems that no. In such case it should be named with `x-podman.` prefix and documentation added to docs/Extensions.md. I also don't think that `relabel` is clear name, though I don't have better suggestions at this time.

Thanks for the suggestion. It is not a part of the docker-compose spec and i don't know about any equivalent option there.

What about changig this to zzmount?

jarovo:
Similar issue -- with volumes in docker was discussed here with a conclusion that the best approach is setting the se-linux labels on host.

https://github.com/moby/moby/issues/32579#issuecomment-2142950183 

Should podman-compose or podman or something on the host relabel the files containing the secrets automagically?

p12tic:
Sorry for delay, I will get back to this.

docs/Extensions.md
```diff
@@ -27,6 +27,14 @@ services:
 
 For explanations of these extensions, please refer to the [Podman Documentation](https://docs.podman.io/).
 
+## Secrets
+```yml
+secrets:
+  custom-secret:
+    x-podman.relabel: z
```
p12tic, line 34: I think it makes sense to borrow values of the podman mount option (https://docs.podman.io/en/latest/markdown/podman-run.1.html#mount-type-type-type-specific-option). Bind and glob support `relabel` option with `shared` and `private` values possible.

p12tic:
I looked into how podman handles relabeling and it does have `relabel` suboption of some of its option which does similar things. So I think current naming is great. Let's rename `zZ` to `shared` and `private`. Other than that, looks great, thank you.

tests/unit/test_container_to_args_secrets.py
```diff
@@ -302,6 +302,17 @@ async def test_secret_target_matches_secret_name_secret_type_not_env(self):
             "file_secret",
             repo_root() + "/test_dirname/my_secret:/run/secrets/file_secret:ro,rprivate,rbind",
         ),
+        (
+            "relabel",
```
p12tic, line 306: This will need 2 cases: for private and shared relabel, because we'll be translating things to what podman understands in this context.

jarovo:
I am searching for proper doc for the relabel option. Ther are
[Red hat labs](https://redhatgov.io/workshops/security_container_intro/lab05-selinux/) speaking about them. There is a [blog posts from other geek](https://blog.christophersmart.com/2021/01/31/podman-volumes-and-selinux/) and [here](https://blog.ryanmartin.me/selinux-containers).

It should be noted that the [mount options related to shared subtrees](https://docs.kernel.org/filesystems/sharedsubtree.html) use the words shared and private but I believe they are different thing. This needs further clarification so I am marking this PR as a draft to make sure the we have the proper doc.

jarovo:
So if I understand, there is no blocker for merging this

jarovo:
@p12tic what can I do to get this merged? Thank you.

p12tic:
Sorry for delays, this is one of these more difficult PRs where I need to figure out more things than usual :-)

> I believe :zZ are options for podman --volume option and it actually has not much to do with private or shared mount option.

But `--volume` option documentation starts with "Create a bind mount". I interpret that `--volume` is just a subset of `--mount` option with different syntax (https://docs.podman.io/en/v4.4/markdown/options/mount.html).

If we look into podman sources, we see that `relabel=private` and `relabel=shared` are translated to `zZ` in the end (see here https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/pkg/specgenutil/volumes.go#L409) and then they are handled by podman itself (https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/libpod/container_internal_common.go#L415, https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/libpod/container_internal_common.go#L3124).

jarovo:
> Sorry for delays, this is one of these more difficult PRs where I need to figure out more things than usual :-)
> 
> > I believe :zZ are options for podman --volume option and it actually has not much to do with private or shared mount option.
The truth is they are not available as I would expect while the :z works:

``` shell
[jhenner@vydra tmp]$ rm /tmp/foo/ -rf
[jhenner@vydra tmp]$ mkdir /tmp/foo; echo Hello world > /tmp/foo/bar
[jhenner@vydra tmp]$ ll -Z foo/bar
-rw-r--r--. 1 jhenner jhenner unconfined_u:object_r:user_tmp_t:s0 12  4. čen 12.17 foo/bar
[jhenner@vydra tmp]$ podman run -v ./foo:/foo:relabel=shared alpine cat /foo/bar
Error: invalid option type "relabel=shared"
[jhenner@vydra tmp]$ ll -Z foo/bar
-rw-r--r--. 1 jhenner jhenner unconfined_u:object_r:user_tmp_t:s0 12  4. čen 12.17 foo/bar
```
The selinux label was not touched.
```
[jhenner@vydra tmp]$ podman run -v ./foo:/foo:z alpine cat /foo/bar
Hello world
[jhenner@vydra tmp]$ ll -Z foo/bar
-rw-r--r--. 1 jhenner jhenner system_u:object_r:container_file_t:s0 12  4. čen 12.17 foo/bar
```
Now we are talking! The files have been relabeled by something. Surely podman did this [in the file you mentioned](https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/libpod/container_internal_common.go#L3124
)
> 
> But `--volume` option documentation starts with "Create a bind mount". I interpret that `--volume` is just a subset of `--mount` option with different syntax (https://docs.podman.io/en/v4.4/markdown/options/mount.html).
> 
> If we look into podman sources, we see that `relabel=private` and `relabel=shared` are translated to `zZ` in the end (see here https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/pkg/specgenutil/volumes.go#L409) and then they are handled by podman itself (https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/libpod/container_internal_common.go#L415, https://github.com/containers/podman/blob/6a39f37845f5e812cb56c709694b01431f108a09/libpod/container_internal_common.go#L3124).

Thanks for the sources links! I searched the options declarations in the same repo. I see `:z` `:Z`  options are declared there, but nothing about `:relabel=private` or `:relabel=shared` or similar in [mount_opts.go](https://github.com/containers/podman/blob/4030ee62c8e29ae904c7f07c5621c5cf5d453576/pkg/util/mount_opts.go#L175)

Perhaps that is why only `:z` or `:Z` is working for me. I think the code for handling the `relabel=private` or `relabel=shared` options is an unreachable code due to lack of declaration of these options.

p12tic:
>  I think the code for handling the relabel=private or relabel=shared options is an unreachable code

Indeed, podman tries to conform to docker API and docker only has z and Z in volume option (https://docs.docker.com/engine/storage/bind-mounts/#options-for---volume).

So most likely podman documentation is wrong.

docs/Extensions.md
```diff
@@ -27,6 +27,14 @@ services:
 
 For explanations of these extensions, please refer to the [Podman Documentation](https://docs.podman.io/).
 
+## Secrets
+```yml
```
p12tic, line 31: Add:

`The following extension keys are available under secret configuration:

* `x-podman.relabel` - Configure SELinux relabeling

For example, the following configures `custom-secret` to use mount with private and unshared content.

podman_compose.py
```diff
@@ -618,7 +619,19 @@ def get_secret_args(compose, cnt, secret, podman_is_building=False):
...
-            volume_ref = ["--volume", f"{source_file}:{dest_file}:ro,rprivate,rbind"]
+
+            mount_options = 'ro,rprivate,rbind'
+
+            selinux_relabel_to_mount_option_map = {None: "", "z": ",z", "Z": ",Z"}
+            try:
+                selinux_mount_option = selinux_relabel_to_mount_option_map[x_podman_relabel]
+            except KeyError as exc:
+                raise ValueError(
+                    f'ERORR: Run secret "{secret_name} has invalid "relabel" option related '
```
p12tic, line 630: typo in ERORR

p12tic (changes requested):
LGTM overall, I had several comments. Please squash your commits into one as all subsequent commits are fixing the first commit.

jarovo:
> Thanks for the sources links! I searched the options declarations in the same repo. I see `:z` `:Z` options are declared there, but nothing about `:relabel=private` or `:relabel=shared` or similar in [mount_opts.go](https://github.com/containers/podman/blob/4030ee62c8e29ae904c7f07c5621c5cf5d453576/pkg/util/mount_opts.go#L175)
> 
> Perhaps that is why only `:z` or `:Z` is working for me. I think the code for handling the `relabel=private` or `relabel=shared` options is an unreachable code due to lack of declaration of these options.

The `relabel=shared` and `relabel=private` handling code is not dead code. It can be accessed trough the RUN command in Dockerfile like this:
```
RUN --mount=type=cache,target=/root/.cache/uv \
    --mount=type=bind,source=uv.lock,target=uv.lock,relabel=private \
    --mount=type=bind,source=pyproject.toml,target=pyproject.toml,relabel=private \
    uv sync --locked --no-install-project
```

FYI this should make possible to pre-install the python project deps using `uv` in a layer to speedup rebuild. I did get the base of this technique from [the astreal uv doc]( https://docs.astral.sh/uv/guides/integration/docker/#caching)

p12tic (approved):
Thanks!

Issue #1215, opened 2025-05-21
Paths in services/*/build/ssh don't work when current directory is not the one containing compose file


p12tic:
**Describe the bug**
Consider the following compose file:

```
version: "3"
services:
  myservice:
    build:
      context: ./context
      dockerfile: Dockerfile
      ssh:
        id1: "./id_ed25519"
```

The `id1` ssh key path sent to podman will be relative which means that it will work if and only if current directory matches the directory where the compose file resides (because all relative paths in compose file are relative to that directory).

**To Reproduce**
Setup an example using attached compose file.

**Expected behavior**
Works.

**Actual behavior**
Does not work.

**Additional context**

devurandom:
This appears to be a regression in version 1.4.0.  I am also affected since the update to `podman-compose;1.4.0-1.fc42;noarch;updates` yesterday.

teohhanhui:
This actually affects more than just the SSH path. All path resolution is wrong since 1.4.0 - it's a major bug.

teohhanhui:
```yaml
# .devcontainer/compose.yaml
services:
  server:
    build:
      context: ..
      target: dev
    volumes:
      - type: bind
        source: ..
        target: /srv/myapp
        bind:
          selinux: z
        consistency: cached

```

```json
// .devcontainer/devcontainer.json
{
	"name": "myapp",
	"dockerComposeFile": "compose.yaml",
	"service": "server",
	"workspaceFolder": "/srv/myapp"
}

```

`Dockerfile` is in the project dir (one level up from `.devcontainer` dir)

Being invoked like this from the project dir:

```bash
devcontainer up --workspace-folder .
```

(`devcontainer` is https://github.com/devcontainers/cli)

teohhanhui:
Digging deeper...

On host paths, the spec [says](https://github.com/compose-spec/compose-spec/blob/abff3682f7cc6483a765ac88359ce104b056565b/spec.md#volumes):

> Relative host paths are only supported by Compose that deploy to a local container runtime. This is because **the relative path is resolved from the Compose file’s parent directory** which is only applicable in the local case.

(emphasis mine)

But this seems to be an error in the spec's wording, as it is not the observed behaviour of `docker compose`, nor is it what `podman-compose` itself intends when parsing the [short syntax](https://github.com/compose-spec/compose-spec/blob/abff3682f7cc6483a765ac88359ce104b056565b/spec.md#short-syntax-5), as can be seen here:

[...]

PR #1218, opened 2025-05-22
Fix build ssh path to be relative to directory of compose file

mokibit:
Fixes https://github.com/containers/podman-compose/issues/1215.


podman_compose.py
```diff
@@ -2644,6 +2644,17 @@ def is_path_git_url(path):
     return r.scheme == 'git' or r.path.endswith('.git')
 
 
+def parse_agent_or_key_to_path(compose, agent_or_key):
```
p12tic, line 2647: adjust_build_ssh_key_paths

podman_compose.py
```diff
@@ -2644,6 +2644,17 @@ def is_path_git_url(path):
...
+    # when using a custom id for ssh property, path to a local SSH key is provided after "="
+    split_for_path = agent_or_key.split("=", 1)
```
p12tic, line 2649: split_for_path => parts

teohhanhui:
Please see https://github.com/containers/podman-compose/issues/1215#issuecomment-2902076544

The path resolution breakage affects much more than just the SSH key path.

I believe whatever broke the path resolution should be reverted. It's a major bug in podman-compose.

p12tic:
@teohhanhui Thanks. I think we still need this PR, because user homedir wasn't ever expanded properly.

podman_compose.py
```diff
@@ -2644,6 +2644,16 @@ def is_path_git_url(path):
     return r.scheme == 'git' or r.path.endswith('.git')
 
 
+def adjust_build_ssh_key_paths(compose, agent_or_key):
+    # when using a custom id for ssh property, path to a local SSH key is provided after "="
+    parts = agent_or_key.split("=", 1)
+    if len(parts) > 1:
```
p12tic, line 2650: You can return immediately in the simple case.

PR #1236, opened 2025-06-05
docker-compose compatibility enhancements

### LINKED ISSUE #1230: Update container name separator to hyphen to match Docker Compose v2
Quoting [Docker docs](https://docs.docker.com/compose/releases/migrate/#service-container-names):

> In Compose v1, an underscore (_) was used as the word separator. In Compose v2, a hyphen (-) is used as the word separator.

Mismatch causes issues like [this](https://github.com/devcontainers/cli/issues/1017).

uosis:
A number of enhancements around docker-compose compatibility:

1. Add support for using hyphens for name separation. Completes #1080 and fixes #1230. Controlled by new `name_separator_compat` x-podman setting.
2. Add support for providing x-podman settings using environment variables.
3. Add `docker_compose_compat` x-podman meta setting to make it easier to enable all docker compatibility settings. Note that I did not enable [existing `default_net_name_compat` setting](https://github.com/containers/podman-compose/blob/main/docs/Extensions.md#compatibility-of-default-network-names-between-docker-compose-and-podman-compose) as part of this because it is no longer relevant - `docker-compose` no longer removes hyphens from project name.

Let me know if you agree with these changes, and I will squash and update the docs.


podman_compose.py
```diff
@@ -2229,9 +2229,9 @@ def _parse_compose_file(self) -> None:
 
         self.x_podman = compose.get("x-podman", {})
         self.x_podman.update({
-            key.removeprefix("PODMAN_COMPOSE_").lower(): value  # type: ignore[misc]
+            key.removeprefix("PODMANCOMPOSE_").lower(): value  # type: ignore[misc]
```
p12tic, line 2236: Better use `PODMAN_COMPOSE` prefix. Not merging names is common across podman ecosystem, e.g. `PODMAN_COMPOSE_PROVIDER` and other variables affecting how podman works.

podman_compose.py
```diff
@@ -2219,6 +2219,13 @@ def _parse_compose_file(self) -> None:
             log.debug(" ** merged:\n%s", json.dumps(compose, indent=2))
         # ver = compose.get('version')
 
+        self.x_podman = compose.get("x-podman", {})
+        self.x_podman.update({
+            key.removeprefix("PODMAN_COMPOSE_").lower(): value  # type: ignore[misc]
```
p12tic: I think it makes sense to have an explicit list of allowed environment variables and throw a warning if unsupported one is passed. This solves the problem of people making types and wasting lots of time later.

Also, this commit needs a newsfragment.

podman_compose.py
```diff
@@ -2060,11 +2043,22 @@ async def run(self, argv: list[str] | None = None) -> None:
...
-        return self.global_args.in_pod
+
+    def resolve_pod_name(self) -> str | None:
+        # Priorities:
+        # - Command line --in-pod
+        # - docker-compose.yml x-podman.in_pod
+        # - Default value of true
+        in_pod_arg = self.global_args.in_pod or self.x_podman.get("in_pod", True)
+
+        match try_parse_bool(in_pod_arg):
```
p12tic, line 2058: we still support python 3.9

p12tic:
First commit is good, I will need more time to review the rest. Probably best if you extract that commit into separate PR. If my comments are addressed I would be able to merge it immediately.

p12tic:
Please fix tests before further review.

uosis:
I will split this into individual PRs as requested.

PR #1238, opened 2025-06-06
Add support for providing x-podman settings using environment variables

uosis:
Split from https://github.com/containers/podman-compose/pull/1236. Still needs final squash after review.

Environment variable prefix choice is open to discussion:
1. `PODMAN_COMPOSE_`: conflicts with `PODMAN_` taken by Podman
2. `PODMANCOMPOSE_`: does not follow naming convention by merging words
3. `COMPOSE_`: taken by Docker Compose
4. `X_PODMAN_COMPOSE_`: does not follow naming convention by having x prefix

@p12tic let me know which option you prefer. I left it with `#2` for now. My second choice would be `#4`.

### DIFF
--- a/docs/Extensions.md
+++ b/docs/Extensions.md
@@ -156,6 +156,9 @@ x-podman:
 By default `default_net_name_compat` is `false`. This will change to `true` at some point and the
 setting will be removed.
 
+This setting can also be changed by setting `PODMAN_COMPOSE_DEFAULT_NET_NAME_COMPAT` environment
+variable.
+
 ## Compatibility of default network behavior between docker-compose and podman-compose
 
 When there is no network defined (neither network-mode nor networks) in service,
@@ -176,6 +179,9 @@ x-podman:
     default_net_behavior_compat: true
 ```
 
+This setting can also be changed by setting `PODMAN_COMPOSE_DEFAULT_NET_BEHAVIOR_COMPAT` environment
+variable.
+
 ## Custom pods management
 
 Podman-compose can have containers in pods. This can be controlled by extension key x-podman in_pod.
@@ -195,6 +201,9 @@ x-podman:
     in_pod: false
 ```
 
+This setting can also be changed by setting `PODMAN_COMPOSE_IN_POD` environment
+variable.
+
 It is also possible to override the default arguments for pod creation that are
 used when --pod-args is not passed on the command line:
 ```yml
@@ -208,3 +217,6 @@ x-podman:
 ```
 When not set in docker-compose.yml or on the command line, the pod args default
 to `["--infra=false", "--share="]`.
+
+This setting can also be changed by setting `PODMAN_COMPOSE_POD_ARGS` environment
+variable.


p12tic:
I think it's fine to use `PODMAN_COMPOSE_` prefix. First, this name has least surprise factor, simpliest and easiest to remember because it follows existing conventions. The fact that podman already has `PODMAN_` prefix is not large problem, because podman and podman-compose are not completely unrelated projects, so I think we'll be able to coordinate to not reuse the same name. I think benefits of presenting good name to the user outweighs technical downsides and additional maintenance.

podman_compose.py
```diff
@@ -2070,7 +2078,42 @@ def resolve_pod_args(self) -> list[str]:
...
+            if known_key:
+                self.x_podman[known_key] = v
+            else:
+                log.warning(
+                    "unknown x-podman key [%s] in compose file, supported keys: %s",
+                    k,
+                    ", ".join(known_keys.keys()),
+                )
+
+        env = {
```
p12tic, line 2101: duplicate implementation

p12tic (changes requested):
Looks good, except let's use PODMAN_COMPOSE prefix.

podman_compose.py
uosis: What do you mean? Duplicate of what?

uosis:
Prefix changed.

podman_compose.py
p12tic: the for loop above already sets `self.x_podman[known_key] = v`. It seems that the for loop below does the same.

p12tic:
https://github.com/containers/podman-compose/pull/1238/files#r2134055512 is still relevant

uosis:
> https://github.com/containers/podman-compose/pull/1238/files#r2134055512 is still relevant

@p12tic sorry I did not see your reply. Responded now.

p12tic:
Rebased, squashed and added docs (forgot about this myself).

p12tic (approved):

p12tic:
@uosis One thing that is worth considering, but I don't want to block the PR anymore with it: should environment variables be overridden if the same setting is in compose file? I think it should, in generally the convention is that environment variables is the weakest form to adjust program settings. In which case the two loops where we set x_podman values should be swapped. What do you think?

uosis:
> @uosis One thing that is worth considering, but I don't want to block the PR anymore with it: should environment variables be overridden if the same setting is in compose file? I think it should, in generally the convention is that environment variables is the weakest form to adjust program settings. In which case the two loops where we set x_podman values should be swapped. What do you think?

@p12tic yeah I also debated the precedence, and while both ways have their use cases, the current approach is more consistent based on the following prior art:

1. In Docker Compose, for setting project name, [environment variable overrides compose file](https://docs.docker.com/compose/how-tos/project-name/#:~:text=The%20COMPOSE_PROJECT_NAME%20environment%20variable.).
2. In Git, environment variables of the form `GIT_CONFIG_*` [override all configuration files](https://git-scm.com/docs/git-config?utm_source=chatgpt.com#Documentation/git-config.txt-command).

p12tic:
Thanks, makes sense.

Issue #78, opened 2019-11-08
podman-compose run seems not to boot dependencies


Yajo:
STR:

1. Clone https://github.com/Tecnativa/doodba-scaffolding
1. `ln -s devel.yaml docker-compose.yml`
1. `podman-compose build`
1. `podman-compose run --rm odoo psql -l`

<details><summary>Logs</summary>

```
➤ podman-compose run --rm odoo psql -l
podman pod create --name=doodbadevel12 --share net -p 0.0.0.0:12069:8069 -p 127.0.0.1:6899:6899 -p 127.0.0.1:8025:8025 -p 127.0.0.1:1984:1984
Error: unable to create pod: error adding pod to state: name doodbadevel12 is in use: pod already exists
125
Namespace(T=False, cnt_command=['psql', '-l'], command='run', detach=False, dry_run=False, e=None, entrypoint=None, file=['docker-compose.yml'], label=None, name=None, no_ansi=False, no_cleanup=False, no_deps=False, podman_path='podman', project_name=None, publish=None, rm=True, service='odoo', service_ports=False, transform_policy='1podfw', user=None, volume=None, workdir=None)
podman volume inspect doodbadevel12_filestore || podman volume create doodbadevel12_filestore
[...]

PR #1254, opened 2025-06-28
wait for deps in compose_run

doc-sheet:
Hello.
This PR may be treated as an extra fix for #78
Currently podman compose run starts dependencies but doesn't check for condition requirements.

Could you suggest what test are needed?


doc-sheet:
Yeah. I'm still testing it.
I found some strange thing with clickhouse healthchecks and don't have much time to investigate.
I don't want to bring breaking changes.

Issue #708, opened 2023-06-11
No variable interpolation in label name


benzht:
**Describe the bug**
Variable interpolation from `.env` file is not happening everywhere where expected. In the following excerpt for labels for a traefik frontend, the env-variable `PROJECT_NAME` is interpolated _only at the right of the `=` character_, but not on the left

```
    labels:
      - "traefik.http.routers.${PROJECT_NAME}_VB_secure.rule=Host(`${HOST_NAME}`)"
      - testThis.${PROJECT_NAME}=${PROJECT_NAME}
```
With
```
PROJECT_NAME=testproject
HOST_NAME=www.example.com
```

The actually resulting labels look like this:
```
                "--label",
                "testThis.${PROJECT_NAME}=testproject",
                "--label",
                "traefik.http.routers.${PROJECT_NAME}_VB_secure.rule=Host(`www.example.com`)",
```
Expected outcome would be:
```
                "--label",
                "testThis.testproject=testproject",
                "--label",
                "traefik.http.routers.testproject_VB_secure.rule=Host(`www.example.com`)",
```

**To Reproduce**
Take any podman-compose file you have available, add the labels like show above, and define the variables in your .env file. Start the container and observe the generated podman command or `inspect` the created container.

**Environment:**
 - OS: `Ubuntu 22.04.2 LTS`
 - podman version:
```
$ podman version
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.17.3
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
``` 
 - podman compose version:
```
$ podman-compose version
podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 3.4.4
podman-compose version 1.0.6
podman --version
podman version 3.4.4
exit code: 0
```

ngolatka:
Can confirm this exact problem.

```
podman version
Client:       Podman Engine
Version:      4.9.4-rhel
API Version:  4.9.4-rhel
Go Version:   go1.21.11 (Red Hat 1.21.11-1.el9_4)
Built:        Mon Jul  1 12:27:14 2024
OS/Arch:      linux/amd64
```

awildturtok:
We are also experiencing this issue on v1.4.0, podman 4.9.4-rhel.

mokibit:
@benzht Thank you for opening an issue!
It was fixed in #1258.

PR #1240, opened 2025-06-10
Apply environment substitution to dictionary keys

awildturtok:
## Contributor Checklist:

If this PR adds a new feature that improves compatibility with docker-compose, please add a link
to the exact part of compose spec that the PR touches.

For any user-visible change please add a release note to newsfragments directory, e.g.
newsfragments/my_feature.feature. See newsfragments/README.md for more details.

All changes require additional unit tests.

PR #1258, opened 2025-06-30
Implement environment variable interpolation to YAML dictionary keys

mokibit:
`podman-compose` currently does not support interpolating environment variables in dictionary keys, despite the compose file specification indicating this capability. 
See the relevant compose-spec documentation:
https://github.com/compose-spec/compose-spec/blob/main/12-interpolation.md

This feature is useful in `labels` or `environment` sections, where keys can be user-defined strings. To enable interpolation, an alternate equal sign syntax must be used, e.g.:
```
services:
  foo:
    labels:
      - "$VAR_NAME=label_value"
```
After this PR `podman-compose` will align more closely with the compose file specification, allowing for the interpolation of environment variables in dictionary keys.

This PR completes PR https://github.com/containers/podman-compose/pull/1240 and fixes issue #708.

### DIFF
--- a/newsfragments/env-var-interpolation-for-keys.feature
+++ b/newsfragments/env-var-interpolation-for-keys.feature
@@ -1,0 +1,1 @@
+Add support for environment variable interpolation for YAML keys.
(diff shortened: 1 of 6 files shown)


p12tic (approved):
Looks great, thank you. Just one nit.

PR #1283, opened 2025-08-09
fix: podman-compose down should not stop the upstream dependencies.

### LINKED ISSUE #1260: podman compose down <service> should not stop dependencies
**Describe the bug**
podman compose down <service> should not stop dependencies

**To Reproduce**
```

services:
  db:
    image: localhost/t1:latest
  app1:
    image: localhost/t1:latest
    depends_on:
      - db
  app2:
    image: localhost/t1:latest
    depends_on:
      - db
```
```
# podman-compose up -d 
d0c9a5d152f197efc8eace8c79cc51d57091a9102e1131397e0db9867095258b
b0ab89d9390e37e4d0524779b5820d52f2ee1ff3a79c36c24c4cd1a293e120cf
t1_db_1
f543a8b9c9f0f0c47a0a5dcfb7e570d974e805b522d76642a8f45b7a084ee613
t1_app1_1
5c2971f58d238e08489b98146eded94d8c5c74c443e1d4a967fdbcdce9f8ab26
t1_app2_1
```
**Expected behavior**
podman-compose should not remove upstream services because it is possible that multiple services depends on that service. I want to be able to stop a single service to update it manually since podman-compose doesn't support --watch yet.
Similar to docker compose: https://stackoverflow.com/questions/77267494/docker-compose-do-not-exit-depends-on-containers-on-exit-why

**Actual behavior**
If i try to stop one single service, upstream service is stopped as well

```
# podman-compose down app1
WARN[0010] StopSignal SIGTERM failed to stop container t1_app1_1 in 10 seconds, resorting to SIGKILL 
WARN[0010] StopSignal SIGTERM failed to stop container t1_db_1 in 10 seconds, resorting to SIGKILL 
t1_app1_1
t1_db_1
t1_app1_1
Error: container b0ab89d9390e37e4d0524779b5820d52f2ee1ff3a79c36c24c4cd1a293e120cf has dependent containers which must be removed before it: 5c2971f58d238e08489b98146eded94d8c5c74c443e1d4a967fdbcdce9f8ab26: container already exists


# podman ps -a | grep t1_
b0ab89d9390e  localhost/t1:latest                                                                         7 minutes ago  Exited (137) 7 minutes ago                                                                                    t1_db_1
5c2971f58d23  localhost/t1:latest                                                                         7 minutes ago  Up 7 minutes
```
**Output**

```
# podman-compose version
podman-compose version 1.4.0
podman version 5.4.0
...

```

**Environment:**
 - OS: Linux / WSL / Mac
 - podman version: 
 - podman compose version: (git hex)

**Additional context**

lisongmin:
Say there is an app service depends on a db service, when we run `podman-compose down app`, the db service should not be stopped.

This will fix #1260

### DIFF
--- a/tests/integration/compose_down_behavior/docker-compose_default.yaml
+++ b/tests/integration/compose_down_behavior/docker-compose_default.yaml
@@ -1,0 +1,12 @@
+services:
+  app:
+    image: nopush/podman-compose-test
+    command: ["dumb-init", "/bin/busybox", "httpd", "-f", "-p", "8080"]
+    depends_on:
+      - db
+  db:
+    image: nopush/podman-compose-test
+    command: ["dumb-init", "/bin/busybox", "httpd", "-f", "-p", "8080"]
+  no_deps:
+    image: nopush/podman-compose-test
+    command: ["dumb-init", "/bin/busybox", "httpd", "-f", "-p", "8080"]
--- a/tests/integration/compose_down_behavior/test_compose_down_behavior.py
+++ b/tests/integration/compose_down_behavior/test_compose_down_behavior.py
@@ -1,0 +1,81 @@
+# SPDX-License-Identifier: GPL-2.0
+
+import os
+import unittest
+
+from parameterized import parameterized
+
+from tests.integration.test_utils import RunSubprocessMixin
+from tests.integration.test_utils import podman_compose_path
+from tests.integration.test_utils import test_path
+
+
+def compose_yaml_path(scenario: str) -> str:
+    return os.path.join(
+        os.path.join(test_path(), "compose_down_behavior"), f"docker-compose_{scenario}.yaml"
+    )
+
+
+class TestComposeDownBehavior(unittest.TestCase, RunSubprocessMixin):
+    @parameterized.expand([
+        ("default", ["down"], set()),
+        (
+            "default",
+            ["down", "app"],
+            {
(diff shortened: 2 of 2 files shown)


lisongmin:
Say there is an app service depends on db service, like:

```yaml
services:
    app:
        image: myapp
        restart: always
        depends_on:
            - db
    db:
        image: mydb
```

The behavior of docker-compose and podman-compose is different as follows:


| scenario | docker-compose behavior | current podman-compose behavior | new podman-compose behavior |
| -------- | ----------------------- | ------------------------------- | --------------------------- |
| down all | stop and remove db, app | same as docker-compose          | same as docker-compose      |
| down app | stop and remove app     | stop and remove db, app         | same as docker-compose      |
| down db  | stop and remove db      | stop but failed to remove db    | stop and remove db, app   |


podman-compose relies on podman's `--requires` to determine the dependency between containers.
When we try to remove the db container, podman will refuse to do so because it is required by the
app container. And we can not create a new db container without removing the old one.

Moreover, if we rename the db container, and create a new one, the dependencies will lost,
and there may be two db containers running at the same time, which is not what we want.
So we should remove both db and app containers when we want to remove the db container.

this behavior can only changed when podman support dynamic dependencies, which is not supported now.

p12tic:
Great, thanks!

PR #1294, opened 2025-08-21
Support service level config hash

lisongmin:
Currently, podman-compose does not support service level configuration changes.
This means that any configuration changes will lead to a full recreation of services.

By setting `service_level_change_detect: true` under global `x-podman` key, podman-compose
will calculate a hash of the service level configuration and will only recreate
that service if the hash has changed.

### DIFF
--- a/newsfragments/service_level_change_detect.bugfix
+++ b/newsfragments/service_level_change_detect.bugfix
@@ -1,0 +1,1 @@
+Support service level configuration change detection on up command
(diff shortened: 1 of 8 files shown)


p12tic:
I think this could be considered a bug and thus we don't need to introduce `service_level_change_detect`.

lisongmin:
The `service_level_change_detect` flag is removed.

p12tic (approved):

PR #1297, opened 2025-08-27
Activate profiles of a given targeted service

### LINKED ISSUE #930: Building or Running a Service Fails Without Explicit Profile
**Describe the bug**
In our `compose.yml` files, we make active use of profiles to group services together. It it my understanding, that proper profiles support was just recently added to `podman-compose`, and the newest release (1.1.0) seems to break a bunch of things because of that.

More precisely, if a service has an associated profile, building or running it doesn't seem to work without explicitly specifying a profile that service belongs to. For example, if a service `test` has `profiles: [test_profile]`, then calling `podman-compose build test` fails and `podman-compose --profile test_profile build test` seems to be required.

On the other hand, using `docker compose`, one doesn't need to specify a profile, and I believe that that's the correct behaviour. Quoting from the spec concerning profiles:

> A service is ignored by Compose when none of the listed profiles match the active ones, unless the service is explicitly 
targeted by a command. In that case its profile is added to the set of active profiles.

The way I interpret this, calling `podman-compose build test` does explicitly target the `test` service and specifying the profile should not be necessary.

This happens on both the released version 1.1.0, as well as the latest version in the `main` branch.

**To Reproduce**
1. Create a `Dockerfile` and a `compose.yml` file setting up a single, minimal service that has an assigned profile.
    - `Dockerfile`:
        ```dockerfile
        FROM docker.io/library/alpine:latest
        CMD [ "true" ]
        ```
    - `compose.yml`:
        ```yml
        services:
          test:
            image: test:latest
            profiles: [test_profile]
            build:
              context: .
              dockerfile: Dockerfile
        ```

2. Try to build the service:
    ```bash
    $ podman-compose build test
    WARNING:podman_compose:missing services [test]
    ```
    However, when explicitly specifying a profile as in `podman-compose --profile test_profile build test`, it works.

3. Similarly, running the service without an explicitly set profile (i.e. `podman-compose run test`) also fails. If the profile is specified, everything works as expected.

**Expected behavior**
Building or running a service should be possible without explicitly specifying a profile that service is associated with.

**Actual behavior**
The service is not correctly identified, presumably because none of its listed profiles is active.

**Output**
```
$ podman-compose version
podman-compose version 1.1.0
podman version 4.9.4
```

**Environment:**
 - OS: Linux

Kache:
> A service is ignored by Compose when none of the listed `profiles` match the active
> ones, unless the service is explicitly targeted by a command. In that case its profile is
> added to the set of active profiles.

WIP - have not run tests nor looked at CONTRIBUTING.md yet

See: https://docs.docker.com/reference/compose-file/profiles/
Fixes: #930

## Contributor Checklist:

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2254,6 +2254,8 @@ class PodmanCompose:
         }
         requested_profiles = set(args.profile).union(profiles_from_env)
 
+        target_service = (args and 'service' in args and args.service) or None
+
         compose: dict[str, Any] = {}
         # Iterate over files primitively to allow appending to files in-loop
         files_iter = iter(files)
@@ -2319,7 +2321,8 @@ class PodmanCompose:
                 # Solution is to remove 'include' key from compose obj. This doesn't break
                 # having `include` present and correctly processed in included files
                 del compose["include"]
-        resolved_services = self._resolve_profiles(compose.get("services", {}), requested_profiles)
+        resolved_services = self._resolve_profiles(compose.get("services", {}), requested_profiles,
+                                                   target_service)
         compose["services"] = resolved_services
         if not getattr(args, "no_normalize", None):
             compose = normalize_final(compose, self.dirname)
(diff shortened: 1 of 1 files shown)


mokibit:
@p12tic  this draft PR can be closed, at it was resolved by https://github.com/containers/podman-compose/pull/1472.

PR #1314, opened 2025-09-25
Support podman specific 'glob' mount type

rjeffman:
Add support for podman specific mount type 'glob', which allows a set of files to be made available under a mount point in the container.

### DIFF
--- a/tests/unit/test_container_to_args.py
+++ b/tests/unit/test_container_to_args.py
@@ -626,6 +626,50 @@ class TestContainerToArgs(unittest.IsolatedAsyncioTestCase):
 
     @parameterized.expand([
         (
+            "glob_mount",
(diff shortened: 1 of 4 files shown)


tests/unit/test_container_to_args.py
```diff
@@ -624,6 +624,50 @@ async def test_selinux_volume(
...
+                "--mount",
+                (
+                    "type=glob,source="
+                    f"{get_test_file_path('test_dirname/foo') + '/*.ext'},"
+                    "destination=/mnt"
+                ),
+            ],
+        ),
+    ])
+    async def test_volumes_glob_mount_soure(  # pylint: disable=unused-argument
```
p12tic: typo

rjeffman: Thanks, fixed.

PR #1325, opened 2025-10-19
Rich UI for the restart command

FrostyX:
Trying to mimic the `docker-compose` output as much as possible

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -36,6 +36,14 @@ from typing import Sequence
 from typing import Union
 from typing import overload
 
+import time
+from rich.console import Console, Group
+from rich.live import Live
+from rich.spinner import Spinner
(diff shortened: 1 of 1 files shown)


FrostyX:
Screenshot:

Screenshot_2025-10-19_14-47-31

It behaves the same way that `docker-compose` does, i.e. it prints the text and then updates the values and colors as the action progresses. So the screenshot only tells you a half of it, to see everything, you'll have to run it.

The PR is not finished, I spent a couple of hours on this minimal working example that I wanted to show you. If you like it, I'd like to continue working on it and finish the PR.

p12tic:
We need to think if it makes sense to add this as a dependency to the project. Debian includes rich in its repositories, so probably it's not a problem.

Issue #1202, opened 2025-05-16
Support for volume.type=image


lrowe:
Docker compose recently added support for volume.type=image in https://github.com/docker/compose/pull/12725.

Example from https://github.com/docker/compose/blob/23fc76a5403541eeb8dbdd638f88424bbb799d97/pkg/e2e/fixtures/volumes/compose.yaml#L1-L10
```yml
services:
  with_image:
    image: alpine
    command: "ls -al /mnt/image"
    volumes:
      - type: image
        source: nginx:alpine
        target: /mnt/image
        image:
          subpath: usr/share/nginx/html/
```

PR #1336, opened 2025-11-18
Implement 'image' volume type for services

lautat:
This adds support for volume.type=image in services, as specified in [Compose spec](https://github.com/compose-spec/compose-spec/blob/76d4a3d08f9d4eb251092746394a64327031a6c6/05-services.md#long-syntax-5).

Fixes https://github.com/containers/podman-compose/issues/1202

### DIFF
--- a/newsfragments/mount-image-as-volume.feature
+++ b/newsfragments/mount-image-as-volume.feature
@@ -1,0 +1,1 @@
+Support volume.type=image in services, for mounting files from container images.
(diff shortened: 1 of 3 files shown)


lautat:
I also realised that `pull` probably should pull images that are used in volumes. That's what the initial `docker-compose` implementation seemed to do by creating a service for the image that is used in a volume.

p12tic (approved):

PR #1337, opened 2025-11-19
Fix service_completed_successfully to require exit code 0

mrIncompetent:
Previously accepted any container stop, now correctly validates exit code 0 matching Docker Compose behavior.

Docker Compose implementation reference:
https://github.com/docker/compose/blob/v2.29.7/pkg/compose/convergence.go#L433

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -3123,6 +3123,21 @@ def get_excluded(
     return excluded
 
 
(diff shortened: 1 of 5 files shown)


p12tic (changes requested):
Looks good, thanks. Please add release note.

p12tic:
Merged via https://github.com/containers/podman-compose/pull/1481.

Issue #1305, opened 2025-09-16
podman compose pull does not return an exit code != 0 in case of error


Qwarctick:
**Describe the bug**

Using the “podman compose pull” command and encountering an error during extraction (network or missing image) does not return an exit code != 0.

**To Reproduce**

/tmp/fake_image.yaml
```yaml
services:
  foobar:
    image: docker.io/abcedf/zyx:321
```

```sh
$ podman compose -f /tmp/fake_image.yaml pull
Trying to pull docker.io/abcedf/zyx:321
Error: initializing source docker.io/abcedf/zyx:321: reading manifest latest in docker.io/abcedf/zyx:321: requested access to the resource is denied
$ echo $?
0
```

Else, remove the connection. Adapt to your connection.

```sh
sudo nmcli connection down eth0
```

```sh
$ podman compose -f /tmp/fake_image.yaml pull
Error: initializing source docker.io/abcedf/zyx:321: pinging container registry docker.io: Get "https://harbor.cyberwatch.fr/v2/": dial tcp: lookup harbor.cyberwatch.fr on [::1]:53: read udp [::1]:44306->[::1]:53: read: connection refused
$ echo $?
0
```

**Expected behavior**

When pulling ecouting an error, the error code should at least be 1.

**Actual behavior**

Error is 0 not matter the result.

**Environment:**

```
$ podman-compose version
podman-compose version 1.5.0
podman version 5.4.0

$ hostnamectl
 Static hostname: sipa-9
       Icon name: computer-vm
         Chassis: vm 🖴
  Virtualization: kvm
Operating System: AlmaLinux 9.6 (Sage Margay)
     CPE OS Name: cpe:/o:almalinux:almalinux:9::baseos
          Kernel: Linux 5.14.0-570.39.1.el9_6.x86_64
    Architecture: x86-64
 Hardware Vendor: QEMU
  Hardware Model: Ubuntu 24.04 PC _i440FX + PIIX, 1996_
Firmware Version: 1.16.3-debian-1.16.3-2
```

mokibit:
Fixed by https://github.com/containers/podman-compose/pull/1341.

PR #1341, opened 2025-11-28
Properly surface error code from `pull` command

mokibit:
This PR implements returning error codes for podman-compose `pull` command.

Fixes https://github.com/containers/podman-compose/issues/1305.

### DIFF
--- a/tests/integration/commands_fail_exit_code/docker-compose.yml
+++ b/tests/integration/commands_fail_exit_code/docker-compose.yml
@@ -6,3 +6,7 @@ services:
   good:
     build:
       context: good
+  test_pull_good:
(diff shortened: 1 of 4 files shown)


tests/integration/commands_fail_exit_code/docker-compose.yml
```diff
@@ -6,3 +6,7 @@ services:
   good:
     build:
       context: good
+  test_pull_good:
+    image: busybox
```
p12tic: Use local image created by the testsuite.

p12tic (changes requested):
Looks good overall, thanks. Some comments to address.

PR #1342, opened 2025-12-03
Document Python 3.8 requirement

ciis0:
the following changes introduced a requirement on python 3.8+

https://github.com/containers/podman-compose/commit/e03d675b9bcd92864fd1aedc23d92f72e410a54d

https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions


p12tic:
It's actually python 3.9, we would run 3.8 in CI if it's supported.

p12tic:
Adjusted PR according to my comment.

PR #1345, opened 2025-12-05
Fix Podman binary path

### LINKED ISSUE #1306: `podman-compose systemd` generates unit files assuming podman executable at `/usr/bin/podman`, instead of the actual path `/usr/local/bin/podman`
**Describe the bug**
I have a docker-compose project migrated to podman-compose, trying to generate systemd unit files, but the podman path is wrong in all the unit files.

**To Reproduce**
1. the current working directory have docker-compose.yml which defines a single service.
2. run `podman-compose up -d` to bring up the service container
3. run `podman-compose systemd -a create-unit` and `podman-compose systemd -a register` to generate the systemd unit files
4. the generated unit file have a line `ExecStartPre=/usr/bin/podman pod start pod_%i`, but my podman executable is actually installed at `/usr/local/bin/podman`, so the unit file would fail to start.

please use [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for example give me a small busybox-based compose yaml

**Expected behavior**
The systemd service should start without problem.

**Actual behavior**
The podman executable path is wrong so the systemd service fail to start.

**Output**

```
$ podman-compose version
podman-compose version 1.5.0
podman version 5.6.1

$ podman-compose systemd -a create-unit

while in your project type `podman-compose systemd -a register`

$ podman-compose systemd -a register

you can use systemd commands like enable, start, stop, status, cat all without `sudo` like this:

    systemctl --user enable --now 'podman-compose@drone-runner'
    systemctl --user status 'podman-compose@drone-runner'
    journalctl --user -xeu 'podman-compose@drone-runner'

and for that to work outside a session
you might need to run the following command *once*

    sudo loginctl enable-linger 'root'

you can use podman commands like:

    podman pod ps
    podman pod stats 'pod_drone-runner'
    podman pod logs --tail=10 -f 'pod_drone-runner'
```

**Environment:**

```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

**Additional context**

TheRealBecks:
Fixes #1306

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2860,9 +2860,9 @@ Description=%i rootless pod (podman-compose)
 Type=simple
 EnvironmentFile=%h/{stacks_dir}/%i.env
 ExecStartPre=-{script} up --no-start
-ExecStartPre=/usr/bin/podman pod start pod_%i
+ExecStartPre=/usr/bin/env podman pod start pod_%i
 ExecStart={script} wait
-ExecStop=/usr/bin/podman pod stop pod_%i
+ExecStop=/usr/bin/env podman pod stop pod_%i
 
 [Install]
 WantedBy=default.target


TheRealBecks:
The pipeline fails with internal server errors 🥲

p12tic:
> The pipeline fails with internal server errors 🥲

Cloudflare outage takes down a lot of internet it seems :)

Issue #1308, opened 2025-09-17
Logging freeze with multi-byte characters in long output lines


sakamoto-sxd:
**Describe the bug**
It seems that logging gets stuck when long lines contain multi-byte characters.

If you print a very long string that contain multi-byte characters without a new line in it in a container, podman-compose gets stuck and becomes unresponsive. 
It will need keyboard interrupt or a kill.

**To Reproduce**

1. Prepare this compose.yaml in a new directory:
```
services:
  s1:
    image: docker.io/python:3.11-slim
    command: python -c'print("あ" * 1000000 + " end")'
```

2. Run podman-compose up.

**Expected behavior**
It prints a very long line of "あ あ あ ..." and exits.

**Actual behavior**
It gets stuck before printing anything. Ctrl-c will get you the output below.

**Output**
```
# podman-compose version
podman-compose version 1.5.0
podman version 4.9.4-rhel
```

podman-compose up and Ctrl-c after 1 second:

```
# podman-compose up
bc47bad90d0248df9454a04c102b89cc5fbcc929714d9c59eaabb23441a71d56
Trying to pull docker.io/library/python:3.11-slim...
Getting image source signatures
Copying blob 1961ca026b04 done   |
Copying blob fcec5a125fd8 done   |
Copying blob 396b1da7636e done   |
Copying blob a27cb4be7017 done   |
Copying config 2ba2e944f9 done   |
Writing manifest to image destination
6785b3b6c1e7311c67e83a3f610b7b0c30b90b9c329cada25a4c1aa3949d137a
^CError: timed out waiting for file /run/libpod/exits/6785b3b6c1e7311c67e83a3f610b7b0c30b90b9c329cada25a4c1aa3949d137a
test_s1_1
^C[s1] | time="2025-09-08T06:06:36Z" level=error msg="forwarding signal 2 to container 6785b3b6c1e7311c67e83a3f610b7b0c30b90b9c329cada25a4c1aa3949d137a: no container with ID 6785b3b6c1e7311c67e83a3f610b7b0c30b90b9c329cada25a4c1aa3949d137a found in database: no such container"
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-28' coro=<Podman._format_stream() done, defined at /usr/lib/python3.9/site-packages/podman_compose.py:1581> exception=UnicodeDecodeError('utf-8', b'\xe3\x81\x82\xe3\x81\x82\

: snip

[...]

PR #1349, opened 2025-12-11
Use incremental decoder for log output

whym:
Long log lines can be split internally when it is very long. The default decoder (str.decode) raises UnicodeDecodeError against partial fragments of multi-byte characters.  Incremental decoder doesn't have the same problem.

Fixes https://github.com/containers/podman-compose/issues/1308

About tests:

I have done manual testing with the yaml file below:
```
services:
  s1:
    image: docker.io/python:3.11-slim
    command: env LANG=ja_JP.utf8 python -c'print("あ" * 10000000 + " end")'
```

I realize I'm expected to add unit tests, but I don't know what test to add here. Any advice?

- I think it should interact with the decoder's internal buffer size for reproducibility. I don't see that part exposed in the decoder API.
- I think it should handle potential freezes (maybe with a hard time limit of N seconds), because the potential failure mode is not wrong result, but freeze here. I don't know how to imprement that in a unit test in a clean way.

### DIFF
--- a/tests/integration/longlog/docker-compose-multibyte.yml
+++ b/tests/integration/longlog/docker-compose-multibyte.yml
@@ -1,0 +1,4 @@
+services:
+  s1:
+    image: docker.io/python:3.11-slim
+    command: env LANG=ja_JP.utf8 python -c'print("あ" * 10000000 + " end", end="")'
--- a/tests/integration/longlog/test_long_log.py
+++ b/tests/integration/longlog/test_long_log.py
@@ -1,0 +1,47 @@
+# SPDX-License-Identifier: GPL-2.0
+
+import os
+import unittest
+import uuid
+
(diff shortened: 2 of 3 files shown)


whym:
It looks like [proc.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate)'s timeout argument works for this.

whym:
When a service prints an incomplete multibyte character and ends there (as opposed to incomplete characters generated within podman-compose internally), what should we do?

Skip it, emit an error and continue? Raise an exception and immediately exit (like other exceptions)?

whym:
I added commits to do the "emit an error" version. Does it look good?

p12tic (approved):
LGTM, thanks. Last commit should have been 2.

PR #1291, opened 2025-08-18
fix link resolution for mounted volume on macos

aholthagerty:
updated call to os.path.realpath inside assert_volumes to be os.path.abspath

Fixes https://github.com/containers/podman-compose/issues/1290

## Contributor Checklist:

If this PR adds a new feature that improves compatibility with docker-compose, please add a link
to the exact part of compose spec that the PR touches.

For any user-visible change please add a release note to newsfragments directory, e.g.
newsfragments/my_feature.feature. See newsfragments/README.md for more details.

All changes require additional unit tests.

PR #1361, opened 2025-12-23
Support forwarding of '/var/run/docker.sock' on macOS

Cyberax:
Podman Desktop doesn't support forwarding of sockets on macOS, but with this hack we can work around it. This also brings podman-compose to parity with the regular command line.

This PR does not improve compatibility with the spec, but it improves the compatibility with raw Podman on macOS. The current behavior of Podman results in this invocation working fine:

```
podman run -it -v /var/run/docker.sock:/var/run/docker.sock --security-opt label=disable base sh
```

However, podman-compose normalizes the socket paths, so that the equivalent Compose file doesn't work.

The change has an integration test for the macOS desktop.


Cyberax:
Is it possible to merge this patch? It's limited in scope and should not affect other OSes.

p12tic:
There's another PR fixing the same problem: https://github.com/containers/podman-compose/pull/1291. Why wouldn't that approach work?

Cyberax:
It's essentially the same approach. My PR just adds a more detailed comment about this behavior and makes it conditional for macOS. 

So merging either PR is fine.

p12tic:
Merged via https://github.com/containers/podman-compose/pull/1465

Issue #1332, opened 2025-11-07
dockerfile_inline ignored


Fydon:
**Describe the bug**
Using the compose file below with `podman compose --file docker-compose.yml up` results in the Dockerfile in the folder being used for the build instead of `dockerfile_inline` or an error if not Dockerfile is in the folder (`Error: no Containerfile or Dockerfile specified or found in context directory, C:\foo: The system cannot find the file specified.`).

docker-compose.yml
```
services:
  caddy:
    image: foo/caddy:latest
    build:
      dockerfile_inline: |
        FROM caddy:builder AS builder

        RUN \
          xcaddy build \
            --with github.com/caddyserver/transform-encoder

        FROM caddy

        COPY --from=builder /usr/bin/caddy /usr/bin/caddy
    pull_policy: build
```

**To Reproduce**
1. `podman compose --file docker-compose.yml up`

**Expected behavior**
foo/caddy:latest is built with the inline dockerfile

**Actual behavior**
Dockerfile in folder where the command is run is used or an error if no Dockerfile present.


**Output**

```
$ podman-compose version
podman-compose version 1.5.0
podman version 5.6.2

$ podman --version 
Client:       Podman Engine
Version:      5.6.2
API Version:  5.6.2
Go Version:   go1.25.1
Git Commit:   9dd5e1ed33830612bc200d7a13db00af6ab865a4
Built:        Tue Sep 30 20:36:54 2025
OS/Arch:      windows/amd64

Server:       Podman Engine
Version:      5.4.2
API Version:  5.4.2
Go Version:   go1.23.7
Git Commit:   be85287fcf4590961614ee37be65eeb315e5d9ff
Built:        Wed Apr  2 01:00:00 2025
OS/Arch:      linux/amd64

$ podman-compose up
Error: no Containerfile or Dockerfile specified or found in context directory, C:\foo: The system cannot find the file specified.
ERROR:__main__:Build command failed
```

**Environment:**
 - OS: Windows
 - podman version: 5.6.2
 - podman compose version: f7eeda1a3db10952424af6a5b0501c269ebe3f0d

**Additional context**

I reported this in [podman](https://github.com/containers/podman/issues/27392), but they say that the issue is with this project instead. I mentioned this in [this related issue](https://github.com/containers/podman-compose/issues/864#issuecomment-3461396424), but opened a new one given that one is close.

podman compose has the same result as podman-compose. Using docker compose with podman seems to work.

```
[...]

Issue #1312, opened 2025-09-22
compose spec build spec "dockerfile" attribute being ignored


mr-andres-carvajal:
**Describe the bug**
The container spec build spec attribute "dockerfile" is being ignored: https://github.com/compose-spec/compose-spec/blob/main/build.md#dockerfile

Only "Dockerfile" and "Containerfile" are supported

**To Reproduce**

```yaml
services:
  postgres:
    build:
      context: postgres
      dockerfile: Containerfile.custom
```

The above should work, but only works if the dockerfile is named "Dockerfile" or "Containerfile"

When running `podman-compose up`

The error is:

`Error: no Containerfile or Dockerfile specified or found in context directory, {CONTEXT}: The system cannot find the file specified.
ERROR:podman_compose:Build command failed`

**Expected behavior**
Any custom Dockerfile name should be allowed

**Actual behavior**
Will only work if the Dockerfile is named "Dockerfile" or "Containerfile"

**Environment:**
 - OS: WSL
 - podman version: 5.6.1
 - podman compose version: 1.5.0

henjam-bodc:
```yaml
version: '3'

services:
  orm-definitions-api:
    build:
      context: .
      dockerfile: local.dockerfile
    env_file:
      - .env
    ports:
      - 127.0.0.1:8000:8000
```

Would be very interested in seeing the results of this. My company uses many *docker-compose.yml* Compose files with specified dockerfiles like the above. At the moment, none of these work with ` podman-compose `

mr-andres-carvajal:
> version: '3'
> 
> services:
>   orm-definitions-api:
>     build:
>       context: .
>       dockerfile: local.dockerfile
>     env_file:
>       - .env
>     ports:
>       - 127.0.0.1:8000:8000
> Would be very interested in seeing the results of this. My company uses many _docker-compose.yml_ Compose files with specified dockerfiles like the above. At the moment, none of these work with `podman-compose`

We have the same issue. For now we work around it by moving the containerfile to a separate directory and referencing that in the context.

So in your case it would be something like:


```yaml
services:
  orm-definitions-api:
    build: local # directory where you'd move and rename "local.dockerfile" to "Containerfile"
    env_file:
      - .env # I haven't tested this, but I assume this would have to move into the "local" dir too.
    ports:
      - 127.0.0.1:8000:8000
```

hhmmjjnn:
I'm unable to use the Dev Container feature from VSCode because of this.

[...]

PR #1372, opened 2026-01-10
Fix dockerfile and dockerfile_inline path ignored

woidwerse:
Fixes https://github.com/containers/podman-compose/issues/1332 and https://github.com/containers/podman-compose/issues/1312

edit: 
fix issue

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -3016,11 +3016,15 @@ def container_to_build_args(
     path_exists: Callable[[str], bool],
     cleanup_callbacks: list[Callable] | None = None,
 ) -> list[str]:
+    build_args = []
     build_desc = cnt["build"]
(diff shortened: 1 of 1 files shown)


Fydon:
Thank you. This doesn't seem to result in `dockerfile_inline` being used to fix #1332, but seems to result in the correct docker-compose file being used.

Is this attempting to fix more than #1332 (#1332 is listed twice)

jedKnighthook:
I've been following this one as it affects a project I'm working on as well. I believe it was intended to fix #1332 and #1312.

p12tic (changes requested):
Breaks existing tests.

agent-voidverse-xyz:
Fixed. Thanks

Nathan-Kelo:
When will this be released ?

PR #1382, opened 2026-01-25
support cgroup-conf in compose yaml file

JimmyMa:
## Contributor Checklist:

Support extention `x-podman.cgroup_conf` in services, for configuring cgroup v2 parameters of container. 

Currently the `cgroup-conf` can be configured on `podman-compose` command line, like `podman-compose --podman-run-args="--cgroup-conf=memory.high=512M"`, but it's not declarative. One requirement to configure cgroup is the `memory.high`, the value of it is `max` by default, Linux PSI (Pressure Stall Information) needs it to work properly.

### DIFF
--- a/docs/Extensions.md
+++ b/docs/Extensions.md
@@ -25,6 +25,19 @@ services:
         x-podman.rootfs: "/path/to/rootfs"
 ```
 
+* `x-podman.cgroup_conf` - Run the container with list of cgroup v2 configuration entries passed to Podman via `--cgroup-conf`.
+
+Example:
+```yaml
+version: "3"
+services:
+    my_service:
+        command: ["/bin/busybox"]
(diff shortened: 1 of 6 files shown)


p12tic (changes requested):
Marking as "requested changes" for myself.

JimmyMa:
@p12tic could you help to review the change? I think I address your comment. Thank you!!!

p12tic (approved):

p12tic:
@JimmyMa Could you squash your commits into one? The two subsequent commits currently just fix the first commit - no need to put mistakes into history.

p12tic:
@JimmyMa Seems like tests fail in rootless configuration. I think the tests should detect rootless and adjust expectations.

PR #1383, opened 2026-01-26
feat: support nested defaults

GerkinDev:
Add support for nested env var interpolation

### DIFF
--- a/tests/unit/test_rec_subs.py
+++ b/tests/unit/test_rec_subs.py
@@ -9,6 +9,15 @@ from parameterized import parameterized
 from podman_compose import rec_subs
 
 
+def from_pairs(pairs: list[list[str]]) -> tuple[list[str], list[str]]:
+    keys = []
+    values = []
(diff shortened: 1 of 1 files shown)


GerkinDev:
While testing, I saw that some case were incorrect (for example: `${foo:-default} is ${bar-none}`). It seems that a better regex is needed

podman_compose.py
```diff
@@ -260,8 +260,8 @@ def fix_mount_dict(
         (?:{
             (?P<braced>[_a-zA-Z][_a-zA-Z0-9]*)
             (?:(?P<empty>:)?(?:
-                (?:-(?P<default>[^}]*)) |
```
p12tic: I think it's bad idea to allow this to match closing '}'.

We should do something like running the replacement algorithm multiple times.

GerkinDev: Yeah I went this direction afterwards, this is still very much a work in progress.

PR #1387, opened 2026-02-02
Fix orphan removal: use getattr

benrogmans:
When exiting podman compose by repeating ctrl+C rapidly, its state may get corrupt and no longer have the remove_orphans attribute. This broke podman compose for me.

## Contributor Checklist:


p12tic:
Without exact reproduction steps I cannot merge this PR. It's not possible to verify that the issue actually happens and that this PR fixes it.

benrogmans:
I can't reproduce it either as the interrupt needs to be timed precisely... but it is a state that can occur.

Issue #1384, opened 2026-01-28
Feature Request: Full Variable Interpolation Support in .env Files (Recursive and Default Values)


FreeSynergy:
**Is your feature request related to a problem? Please describe.**  
I'm always frustrated when I cannot use recursive variable substitution in `.env` files with `podman compose`. Unlike Docker Compose, `podman compose` does not support resolving variables defined in terms of other variables within the `.env` file. For example, defining `A=Hello` and `B=${A} World` results in `B` being interpreted literally as `${A} World` instead of `Hello World`. This forces redundant definitions and violates the DRY principle, making environment configuration error-prone and harder to maintain.

**Describe the solution you'd like**  
`podman compose` should fully support variable interpolation in `.env` files, including recursive and default-value syntax (e.g., `${VAR}` and `${VAR:-default}`), consistent with Docker Compose behavior. This would allow clean, modular configuration where variables can be defined once and reused, improving readability and reducing duplication.

**Describe alternatives you've considered**  
Currently, I work around this by:
- Using external tools like `envsubst` or custom shell scripts to pre-process `.env` files.
- Duplicating values across the `.env` file, which increases maintenance effort and risk of inconsistency.
- Explicitly using interpolation in the `compose.yaml` (e.g., `ENV_VAR: ${ENV_VAR}`), which forces changes in the compose file and reduces portability. 

These workarounds are non-standard, add complexity, and break compatibility with Docker Compose workflows.

**Additional context**  
This issue is well-documented in the community (e.g., [Issue #718](https://github.com/containers/podman-compose/issues/718), [Issue #1064](https://github.com/containers/podman-compose/issues/1064)) and represents a deviation from Docker Compose's expected behavior. Supporting full `.env` interpolation would improve compatibility, usability, and adherence to established container orchestration practices.

antonpetrov145:
@Lord-KalEl I created a PR for this, the functionality is there, just wasn't applied to environment variables.

mokibit:
I cannot reproduce this issue anymore.
Environment:
```
Debian GNU/Linux 13 (trixie)
Python 3.13.5
podman-compose version 1.6.0
podman version 5.4.2
```
`.env` file:
```
A=Hello
B=${A} World
```
`compose.yml` file:
```
services:
  test:
    image: busybox
    environment:
      A: ${A}
      B: ${B}
      C: ${MISSING:-default}
```
Running `podman-compose config` produces the following output:
```
services:
  test:
    environment:
      A: Hello
[...]

PR #1390, opened 2026-02-10
Add environment variables interpolation

antonpetrov145:
Added the existing functionality to interpolate the environment set in compose.

Refers to [1384](https://github.com/containers/podman-compose/issues/1384)

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2312,15 +2312,16 @@ class PodmanCompose:
             dotenv_path = os.path.realpath(args.env_file)
             dotenv_dict.update(dotenv_to_dict(dotenv_path))
 
-        # Match Regex for env_file directive
-        dotenv_dict = rec_subs(dotenv_dict, dict(os.environ))
         os.environ.update({
             key: value  # type: ignore[misc]
             for key, value in dotenv_dict.items()
-            # if key.startswith("PODMAN_")  # type: ignore[misc]
+            if key.startswith("PODMAN_")  # type: ignore[misc]
         })
         self.environ = dotenv_dict  # type: ignore[assignment]
         self.environ.update(dict(os.environ))
+
+        # Match Regex for env_file directive
+        self.environ = rec_subs(dotenv_dict, dict(os.environ))
         # see: https://docs.docker.com/compose/reference/envvars/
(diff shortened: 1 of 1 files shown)


p12tic:
Tests are broken.

antonpetrov145:
@p12tic seems the tests are passing now, what got me confused are these lines
```
os.environ.update({
    key: value  # type: ignore[misc]
    for key, value in dotenv_dict.items()
    if key.startswith("PODMAN_")  # type: ignore[misc]
})
self.environ = dotenv_dict  # type: ignore[assignment]
self.environ.update(dict(os.environ))
````
when we have the `dotenv_dict` from file we set `self.environ` to it and then update with whatever we have in `os.environ` which means that everything gets replaced back, so my mistake was that I tried to set the `dotenv_dict` with the interpolated values, then update the whole `os.environ` object before it gets assigned back to `self.environ` which seems that broke the test for env files

now I set the interpolation to be done directly on the `self.environ` object, tell me if the logic is not correct, but now all values get interpolated and set correctly

**Comment by p12tic** [comment]
Seems that the PR is still incomplete because 2 lines above we have `self.environ = <...>`.

Also, it would be great to have integration tests that show in a reproducible way what exactly didn't work before and is now fixed.

**Review by p12tic — CHANGES_REQUESTED** [review]
Marking as "request changes" for myself

**Comment by antonpetrov145** [comment]
@p12tic you mean the lines above what I added right, I left those as they are because 
```
self.environ = dotenv_dict  # here we are creating the self.environ obect
```
then here
```
self.environ.update(dict(os.environ))  # we are updating that with what we have in os.environ
```
and my change just interpolates the variables in the final object so there are no variables like `$Abar`
```
self.environ = rec_subs(dotenv_dict, dict(os.environ))
```
tell me if this logic is not correct, if I move the line above where we have the first assign, we will have the same result as now - variables are not interpolated because they are stored like that in `os.environ`

marc2016:
This features looks great and would be super helpful! Thanks for working on this. 

Is there anything I can do to help move this forward? I'd be happy to test this branch locally in my environment if that helps validate the changes.

mokibit:
Closing in favor of #1419, which resolves the same issue.

PR #1399, opened 2026-02-18
[ADD] Support tls verify flag for compose

Chethan30:
## Summary

Added support for disabling TLS verification when podman-compose runs `podman pull`, `podman push`, and `podman build`. This allows use behind VPNs or with self-signed/invalid registry certificates where podman would otherwise fail with x509 TLS errors.

## Motivation

- **Problem:** Under VPNs or with self-signed registry certs, `podman pull` (and push/build that contact registries) can fail with x509 certificate verification errors.
- **Podman** already supports `--tls-verify=false` for `podman pull`, `podman push`, and `podman login`, but podman-compose did not expose this when invoking these commands internally.
- **Workaround today:** Users can pass `--podman-pull-args "--tls-verify=false"` (and similarly for push/build), but that is hard to discover and easy to forget for commands like `up` that trigger pulls. This change provides a single, explicit option.

## Changes

- **CLI:** New global flag `--tls-verify` with values `true` or `false` (default `true`). When set to `false`, podman-compose passes `--tls-verify=false` to all registry-related podman subcommands.
- **Injection:** In `get_podman_args(cmd)`, when `--tls-verify=false` is in effect and `cmd` is `pull`, `push`, or `build`, the returned arguments include `--tls-verify=false`. This automatically covers:
  - **pull:** `podman-compose pull` and the pull phase in `podman-compose up`
  - **push:** `podman-compose push`
  - **build:** `podman-compose build` and the build phase in `up` (e.g. when using `--pull=always`)
- **Tests:** New unit tests in `tests/unit/test_get_podman_args_tls_verify.py` for `get_podman_args()` with `tls_verify` true/false and for registry vs non-registry commands.
- **Backward compatibility:** Default remains TLS verification **on**. Behavior is unchanged when the new flag is not used.

### DIFF
--- a/tests/unit/test_get_podman_args_tls_verify.py
+++ b/tests/unit/test_get_podman_args_tls_verify.py
@@ -1,0 +1,69 @@
+# SPDX-License-Identifier: GPL-2.0
+
+import argparse
+import unittest
+
+from parameterized import parameterized
+
+from podman_compose import PodmanCompose
+
+
+def compose_with_tls_verify(tls_verify: str) -> PodmanCompose:
+    compose = PodmanCompose()
+    compose.global_args = argparse.Namespace(
+        podman_args=[],
+        tls_verify=tls_verify,
+        podman_pull_args=[],
+        podman_push_args=[],
+        podman_build_args=[],
+        podman_run_args=[],
+        podman_inspect_args=[],
+        podman_start_args=[],
+        podman_stop_args=[],
+        podman_rm_args=[],
+        podman_volume_args=[],
+    )
+    return compose
+
+
+class TestGetPodmanArgsTlsVerify(unittest.TestCase):
+    @parameterized.expand([
+        ("pull",),
+        ("push",),
+        ("build",),
+    ])
+    def test_tls_verify_true_registry_commands_no_tls_flag(self, cmd: str) -> None:
+        compose = compose_with_tls_verify("true")
+        xargs = compose.get_podman_args(cmd)
+        self.assertNotIn("--tls-verify=false", xargs)
+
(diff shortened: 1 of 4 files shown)


tests/unit/test_get_podman_args_tls_verify.py
```diff
@@ -0,0 +1,69 @@
...
+        podman_rm_args=[],
+        podman_volume_args=[],
+    )
+    return compose
+
+
+class TestGetPodmanArgsTlsVerify(unittest.TestCase):
+    def test_tls_verify_true_pull_push_build_no_flag(self) -> None:
+        compose = compose_with_tls_verify("true")
+        for cmd in ("pull", "push", "build"):
```
p12tic: use parameterized.expand

tests/unit/test_get_podman_args_tls_verify.py
```diff
@@ -0,0 +1,69 @@
...
+            xargs = compose.get_podman_args(cmd)
+            self.assertNotIn(
+                "--tls-verify=false",
```
p12tic: Brittle test, it should verify that there are no `--tls-verify*` flags. Similar problem with other tests.

Chethan30:
> tls-verify still needs documentation. Also look into git history and structure your commits like in other PRs.
> 
> Please read project requirements carefully because I don't have time to review PRs where I need to explicitly point out every little thing it misses. Such PRs will be ignored.

Apologies. It's my first open PR in a while. Some of our internal practices had followed through. I've made effort to fix the style and requirements. Please let me know if any changes. 

Thank you for your time.

Issue #1319, opened 2025-10-07
Setting top level `name:` in compose file to an environment variable fails


jameeters:
**Describe the bug**
Attempting to set the top level `name:` attribute in a compose file to an environment variable results in an error, even when the same variable is properly applied elsewhere in the file. 

I encountered this on macos, and could not reproduce it on my linux machine, so it **may be specific to macos**.
I'll include details for both machines below.

**To Reproduce**
1. Create `docker-compose.yml`:
```
version: "3.8"
name: ${INSTANCE_NAME}
services:
  postgres:
    container_name: "${INSTANCE_NAME}_pg"
    image: "postgres:16-bullseye"
```
2. create `.env`:
```
export INSTANCE_NAME='ALICE'
```
3. `source .env`
4. `podman compose up`

**Expected behavior**
On linux, `podman compose up` gives:
 ```
>>>> Executing external compose provider "/usr/bin/podman-compose". Please refer to the documentation for details. <<<<

podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.9.3
** excluding:  set()
['podman', 'ps', '--filter', 'label=io.podman.compose.project=ALICE', '-a', '--format', '{{ index .Labels "io.podman.compose.config-hash"}}']
['podman', 'network', 'exists', 'ALICE_default']
podman create --name=ALICE_pg --label io.podman.compose.config-hash=3492c3437da7e8a60e214621247c689574fa3069d57a4fb80c1e7c7a9a3c8831 --label io.podman.compose.project=ALICE --label io.podman.compose.version=1.0.6 --label PODMAN_SYSTEMD_UNIT=podman-compose@ALICE.service --label com.docker.compose.project=ALICE --label com.docker.compose.project.working_dir=/home/james/Documents/podman_test --label com.docker.compose.project.config_files=docker-compose.yml --label com.docker.compose.container-number=1 --label com.docker.compose.service=postgres --net ALICE_default --network-alias postgres postgres:16-bullseye
cd846e8441016be887c9d6896bcc135f82d5f85219a585296b77d72e190e78a3
exit code: 0
podman start -a ALICE_pg

(container output omitted)
```

**Actual behavior**
On macos, `podman compose up` produces:
```
>>>> Executing external compose provider "/opt/homebrew/bin/podman-compose". Please see podman-compose(1) for how to disable this message. <<<<

Traceback (most recent call last):
[...]

PR #1416, opened 2026-03-13
Fix top-level `name` interpolation

Ramdek:
This fixes the top-level name attribute interpolation not happening.

See https://github.com/containers/podman-compose/issues/1319.

### DIFF
--- a/tests/integration/env_tests/container-compose.yml
+++ b/tests/integration/env_tests/container-compose.yml
@@ -1,6 +1,6 @@
 version: "3"
 
-name: my-project-name
+name: my-project-name${ZZVAR1:-}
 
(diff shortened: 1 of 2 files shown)


Ramdek:
Sorry for the noise.  
Would you prefer gpg signed commits with my pseudo as sign-off or non gpg commits with real name as sign-off ?

p12tic:
>  Would you prefer gpg signed commits with my pseudo as sign-off or non gpg commits with real name as sign-off ?

Best would be real name sign-off . GPG signing is not necessary

Ramdek:
One check fails because of docker pull rate limit.

PR #1419, opened 2026-03-17
support nested variable interpolation

blechschmidt-ldc:
This commit adds support for nested variable interpolation. Instead of relying on regular expressions with which adhering to the Compose spec will be impossible, lookaheads for matching closing brackets are performed. In addition, support for alternative values (:+ and + operators) is added and the full interpolation feature set described at https://github.com/docker/docs/blob/050f15d67d5956e8a48f8b92c9449c21bc3207cb/content/reference/compose-file/interpolation.md is implemented.

### DIFF
--- a/newsfragments/nested-variable-interpolation.feature
+++ b/newsfragments/nested-variable-interpolation.feature
@@ -1,0 +1,1 @@
+Support nested variable interpolation
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -23,6 +23,7 @@ import random
 import re
 import shlex
 import signal
+import string
 import subprocess
 import sys
 import tempfile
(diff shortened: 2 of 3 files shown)


blechschmidt-ldc:
Hi,

thanks a lot for your review. I have adapted the PR (and rebased it onto main, which unfortunately makes it a little harder to track the changes introduced through my force push through the GitHub interface).

To summarize, the main changes I have made are the following:

1. Inside `${<content>}`, if `<content>` starts with a char that is not a letter or `_`, an error is returned:
https://github.com/containers/podman-compose/blob/4aa55126882b7dcddfa8ba58843703d373f44aad/podman_compose.py#L375-L381
2. A sequence of `$` followed by any character that is not alphabetic and not `_` is rendered as is. This matches Docker Compose behavior, i.e. `price $5` is kept as is, while differing from POSIX shell behavior, where `$5` would correspond to the fifth argument:
https://github.com/containers/podman-compose/blob/4aa55126882b7dcddfa8ba58843703d373f44aad/podman_compose.py#L430-L433
3. Tests for this behavior have been added. https://github.com/containers/podman-compose/blob/4aa55126882b7dcddfa8ba58843703d373f44aad/tests/unit/test_var_interpolate.py#L143-L174

Unrelated to this, I have also beautified the value error raised in case of an invalid operator to include the surrounding `${...}`:
https://github.com/containers/podman-compose/blob/4aa55126882b7dcddfa8ba58843703d373f44aad/podman_compose.py#L394

p12tic (approved):

Issue #1423, opened 2026-03-21
Support secrets.*.environment


issei-m:
**Is your feature request related to a problem? Please describe.**

When using `environment` as a secret source in `compose.yaml`, podman-compose fails with `ERROR: unparsable secret` and refuses to start.

```yaml
secrets:
aws-access-key-id:
  environment: 'AWS_ACCESS_KEY_ID'
```

This works fine with Docker Compose but is not supported in podman-compose.

**Describe the solution you'd like**

Implement the environment source type for top-level secrets, as defined in the
[Compose Specification](https://github.com/compose-spec/compose-spec/blob/3911ddb6a41f50aea124f033877e873ce1c96f8c/09-secrets.md):

▎ environment: The secret is set to the value of an environment variable.

The expected behavior is that `get_secret_args()` in podman_compose.py reads the
environment key and passes it to podman build as `--secret id=<id>,env=<VAR>`.
Buildah has supported `--secret id=X,env=Y` since v1.24.0
(containers/buildah#3598),
so no changes outside of podman-compose are required.

**Describe alternatives you've considered**

- file source: Requires writing secret values to host files, which is undesirable
for credentials that are already available as environment variables (e.g. in CI).
- external source: Requires pre-creating podman secrets manually before running
podman-compose, which adds friction and breaks compatibility with Docker Compose.

**Additional context**

Minimal reproduction:

```yaml
# compose.yaml

secrets:
aws-access-key-id:
  environment: 'AWS_ACCESS_KEY_ID'

services:
app:
  build:
	context: .
	secrets:
	  - aws-access-key-id
```

```dockerfile
# Dockerfile

# syntax=docker/dockerfile:1
FROM alpine
RUN --mount=type=secret,id=aws-access-key-id \
cat /run/secrets/aws-access-key-id
```

Running podman-compose build produces:
ERROR: unparsable secret: "aws-access-key-id", service: "app"

The environment source type is part of the official Compose Specification
([compose-spec/compose-spec#09-secrets.md](https://github.com/compose-spec/compose-spec/blob/3911ddb6a41f50aea124f033877e873ce1c96f8c/09-secrets.md#example-2))
and is implemented in Docker Compose. podman-compose should support it for compatibility.

PR #1424, opened 2026-03-21
Support secrets.*.environment

issei-m:
Fixes https://github.com/containers/podman-compose/issues/1423

Compose Spec: https://github.com/compose-spec/compose-spec/blob/main/09-secrets.md

### Note

Runtime support for `environment` secrets is not included in this PR. Docker Compose handles this by injecting the secret into the container via the Docker API (`CopyToContainer`) after creation but before start. Implementing the
  equivalent in podman-compose would require changes beyond `get_secret_args`, which only generates CLI arguments, so I considered it out of scope for this PR. Happy to implement that part as well if desired.

### DIFF
--- a/tests/integration/build_secrets/docker-compose.yaml
+++ b/tests/integration/build_secrets/docker-compose.yaml
@@ -11,12 +11,17 @@ services:
     build:
       context: .
       secrets:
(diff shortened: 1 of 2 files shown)


p12tic (approved):

issei-m:
@p12tic Thank you for your review. I've added an integration test for it.

Issue #1426, opened 2026-03-25
Missing integration tests for --policy flag


p12tic:
PR #1298 was merged without integration tests. It broke pull functionality on older podman and due to missing integration tests this wasn't noticed.

Issue #1417, opened 2026-03-15
compose file with both build and image results in errors in console


p12tic:
**Describe the bug**

Running `podman-compose up -d` on the following compose file:

```
services:
  test:
    build:
      context: .
      dockerfile: Dockerfile
    image: my-test-image
```

results in:
```
Error: unknown flag: --policy
See 'podman pull --help'
```

**To Reproduce**
Run the compose file above.

**Expected behavior**
No error.

**Actual behavior**
Error output.

This is with latest podman-compose from master branch as of 2026-03-15 and podman 5.5.

$ podman-compose up
...

```

**Environment:**
 - OS: Linux / WSL / Mac
 - podman version: 
 - podman compose version: (git hex)

**Additional context**

Add any other context about the problem here.

PR #1298, opened 2025-08-27
Feat pull images before teardown containers on up command

lisongmin:
This PR reduces services downtime by pulling images before teardown containers on up command.

PR #1427, opened 2026-03-26
Only pull image before containers teardown if podman version >= 5.6.0

mokibit:
This PR adds missing integration tests (https://github.com/containers/podman-compose/issues/1426) and fixes the `Error: unknown flag: --policy` errors (https://github.com/containers/podman-compose/issues/1417).

After container teardown, podman `create` internally pulls missing images using the default `--pull=missing` policy.
By explicitly calling podman `pull` before teardown, we reduce container downtime. This requires to pass the `--policy=missing` option to podman `pull` to maintain podman-compose's defaults.
Since this option was introduced in Podman 5.6.0, the optimization is only available from that version onwards.

Integration tests for #1298 (Feat pull images before teardown containers on up command, ef9785a) were missing, so compatibility issues with older Podman versions were not caught.

podman `pull --policy` documentation:
https://docs.podman.io/en/stable/markdown/podman-pull.1.html
    
compose specification for `pull_policy`:
https://github.com/compose-spec/compose-spec/blob/main/05-services.md#pull_policy

### DIFF
--- a/newsfragments/fix_policy_flag_error.bugfix
+++ b/newsfragments/fix_policy_flag_error.bugfix
@@ -1,0 +1,1 @@
+Fix shell error: `unknown flag: --policy`.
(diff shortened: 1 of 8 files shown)


whatsupbros:
Any update on this PR review? It kinda blocks the new release :)

PR #1434, opened 2026-04-09
Fix override tag to work with `depends_on` parameter

Malex14:
This pr fixes a bug where `!override` doesn't work with `depends_on` in the service configuration. Currently the following example does work in docker-compose, but throws an error with podman-compose:

compose.yml:
```yaml
services:
  app:
    image: busybox
    command: ["/bin/busybox", "echo", "Hello"]
    depends_on:
      - db
```

compose.override.yml:
```yaml
services:
  app:
    depends_on: !override
      - other_db
  other_db:
    image: busybox
    command: ["/bin/busybox", "sleep", "10"]
```
Here is the corresponding Docker-Documentation: https://docs.docker.com/reference/compose-file/merge/#replace-value

This issue is also referenced in #1208

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2059,10 +2059,12 @@ def normalize_service(service: dict[str, Any], sub_dir: str = "") -> dict[str, A
             deps = {deps: {}}
         elif is_list(deps):
             deps = {x: {} for x in deps}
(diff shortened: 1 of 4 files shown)


p12tic (approved):
Thanks a lot!

PR #1438, opened 2026-04-15
Honor `target` option for mount-type external secrets

Qwarctick:
Aligns behavior with docker-compose, which always honors `target`.

Previously, `target` on an external secret was only forwarded to `podman run --secret` when `type=env`. For `type=mount` (the default), the target was silently dropped and a warning was emitted, even though podman (>= 4.0) supports `--secret name,type=mount,target=/path`.

Changes:
- Always append `target=X` to the `--secret` options when set, not only for env-type secrets.
- Drop the "un-supported and a no-op and is ignored" warning.
- Drop the ValueError raised when `target != secret_name` and `type != env` — this was rejecting configurations that podman now accepts natively.

The check for a mismatched `name:` on the declared secret is kept: external secrets with a custom name still raise ValueError, as podman cannot rename external secrets at mount time.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -891,28 +891,13 @@ def get_secret_args(
         secret_opts += f",gid={secret_gid}" if secret_gid else ""
         secret_opts += f",mode={secret_mode}" if secret_mode else ""
         secret_opts += f",type={secret_type}" if secret_type else ""
-        secret_opts += f",target={secret_target}" if secret_target and secret_type == "env" else ""
-        # The target option is only valid for type=env,
-        # which in an ideal world would work
-        # for type=mount as well.
-        # having a custom name for the external secret
-        # has the same problem as well
+        secret_opts += f",target={secret_target}" if secret_target else ""
+        # having a custom name for the external secret is not supported
         ext_name = declared_secret.get("name")
-        err_str = (
-            'ERROR: Custom name/target reference "{}" '
-            'for mounted external secret "{}" is not supported'
-        )
         if ext_name and ext_name != secret_name:
-            raise ValueError(err_str.format(secret_name, ext_name))
-        if secret_target and secret_target != secret_name and secret_type != 'env':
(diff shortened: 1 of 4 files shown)


podman_compose.py
```diff
@@ -891,28 +891,13 @@ def get_secret_args(
...
+        secret_opts += f",target={secret_target}" if secret_target else ""
+        # having a custom name for the external secret is not supported
         ext_name = declared_secret.get("name")
-        err_str = (
-            'ERROR: Custom name/target reference "{}" '
-            'for mounted external secret "{}" is not supported'
-        )
         if ext_name and ext_name != secret_name:
-            raise ValueError(err_str.format(secret_name, ext_name))
-        if secret_target and secret_target != secret_name and secret_type != 'env':
```
p12tic, line 907: wondering why these checks were there in the first place. Do you know it this was limitation of older podman?

Qwarctick: Yes, this was a limitation of older podman.

When the original secrets support was added in July 2021 ([`62aa337`](https://github.com/containers/podman-compose/commit/62aa337)), podman 3.x rejected `--secret name,type=mount,target=/path` with `target option is invalid for
  mounted secrets`. The in-code comment at the time literally said *"The target option is only valid for type=env, which in an ideal world would work for type=mount as well."* The `ValueError` and the "un-supported and a no-op" warning
  were a workaround to fail fast before calling podman.

  This was fixed upstream in [containers/podman#12294](https://github.com/containers/podman/pull/12294) (merged 2021-11-16, closing [#12287](https://github.com/containers/podman/issues/12287)) and shipped in **podman v4.0.0** (February
  2022). Since then, `--secret name,type=mount,target=/path` works natively, so the checks have been obsolete for a while, hence their removal here.

  The only guard kept is the one on a mismatched `name:` for external secrets, since podman still cannot rename an external secret at mount time.

p12tic (approved):

PR #1441, opened 2026-04-19
Adds the correct behavior for the dockerfile attribute when using git context.

KKonaOG:
The current implementation of Git remote contexts fails to respect/honor the Dockerfile attribute in the compose file. This fix results in "Dockerfile" being used if the attribute is unset and the value of the dockerfile attribute otherwise. This brings it in-line with docker compose and the intended compose-spec behavior.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -3488,6 +3488,9 @@ def container_to_build_args(
                 raise OSError(f"Dockerfile not found in {dockerfile}")
             raise OSError(f"Dockerfile not found in {ctx}")
 
+    if is_context_git_url(ctx) and dockerfile:
+        build_args.extend(["-f", dockerfile])
+
     build_args.extend(["-t", cnt["image"]])
 
     if "platform" in cnt:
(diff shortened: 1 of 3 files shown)


podman_compose.py
```diff
@@ -3317,6 +3317,9 @@ def cleanup_temp_dockfile() -> None:
                 raise OSError(f"Dockerfile not found in {dockerfile}")
             raise OSError(f"Dockerfile not found in {ctx}")
 
+    if is_context_git_url(ctx) and dockerfile:
```
p12tic: I think it makes sense to reuse `if not is_context_git_url(ctx):` check above. Remove `not` to simplify the conditional.

p12tic (changes requested):
Looks great. Just a small nit and it's ready to merge.

KKonaOG:
I believe that is what you were suggesting! I'll remember to do a sign-off commit eventually haha!

p12tic (approved):

PR #1446, opened 2026-05-03
fix: prefer built image over registry when build: is set (#1445)

### LINKED ISSUE #1445: build: image is not used if image: is also set
If both `build:` and `image:` are present, the image builds but is not used, and it's impossible to make use of it even with `--build` option:
```yml
  web:
    image: caddy:alpine
    build: .
```
Run the service:
```console
$ podman-compose --verbose run --rm --service-ports --build web
...
INFO:podman_compose:podman build -f Containerfile -t caddy:alpine <compose dir>
...
# <compose dir>_web:latest is built..
Successfully tagged localhost/caddy:alpine
Successfully tagged localhost/<compose dir>_web:latest
...
# ..but is not used then - downloaded caddy:alpine is run instead
INFO:podman_compose:podman run ... caddy:alpine
```
If we remove `image:`, the image built is used then properly:
```yml
  web:
    build: .
```
```console
$ podman-compose --verbose run --rm --service-ports --build web
...
# <compose dir>_web image is run properly
INFO:podman_compose:podman run ... <compose dir>_web
```
This behavior doesn't correspond to docker-compose, where `build:` has a preference.

```
$ podman-compose version
podman-compose version 1.5.0
podman version 5.8.2
```
**Environment:**
 - OS: Linux

gavin913-lss:
## Problem

When both `build:` and `image:` keys are present in a service definition, `podman-compose` builds the image but then runs the registry image instead of the locally built one.

```yaml
web:
  image: caddy:alpine
  build: .
```

`podman build -t caddy:alpine` creates `localhost/caddy:alpine`, but `podman run caddy:alpine` resolves to `docker.io/library/caddy:alpine` (the registry image). This does not match docker-compose behaviour, where the built image is preferred.

Fixes #1445

## Root Cause

In `build_one()`, after a successful build, the `cnt["image"]` reference still holds the bare image name (e.g., `caddy:alpine`). When podman resolves this at run time, it looks up the registry instead of the locally built image (tagged as `localhost/caddy:alpine`).

## Fix

After a successful build, when the image name does not already have a `localhost/` prefix or a registry domain (no `/`), update `cnt["image"]` to the `localhost/` form. This ensures `podman run` resolves to the locally built image.

## Testing

Added 4 unit tests covering:
- **Bare image name** (e.g., `caddy:alpine`): gets prefixed with `localhost/`
- **Already localhost-prefixed**: no double-prefixing
- **Registry-prefixed** (e.g., `ghcr.io/user/caddy:alpine`): left unchanged
- **Build failure**: image name not modified

All 54 existing unit tests continue to pass.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -3368,6 +3368,17 @@ async def build_one(compose: PodmanCompose, args: argparse.Namespace, cnt: dict)
     status = await compose.podman.run([], "build", build_args)
     for c in cleanup_callbacks:
         c()
+
+    # When both build: and image: are specified, podman build tags the image
+    # as localhost/<image>, but podman run resolves the bare image name to
+    # the registry. Update the image reference to the localhost/ form so the
+    # locally built image is used instead of the registry one (matching
+    # docker-compose behaviour). See: #1445
+    if status == 0 and "image" in cnt:
+        image = cnt["image"]
+        if not image.startswith("localhost/") and "/" not in image:
+            cnt["image"] = f"localhost/{image}"
+
     return status
 
 
--- a/tests/unit/test_prefer_built_image.py
+++ b/tests/unit/test_prefer_built_image.py
@@ -1,0 +1,76 @@
+"""Test that when both build: and image: are specified, the built image
+is preferred over the registry image. See: #1445"""
+
+import unittest
+from argparse import Namespace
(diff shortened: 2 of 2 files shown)


bam80:
Thanks, why is closed?

PR #1449, opened 2026-05-06
fix: recursively pull explicitly depended services regardless of active profile

vkaylee:
## Contributor Checklist:

these guidelines.

- Compatibility improved: [Docker Compose Profiles Documentation](https://docs.docker.com/compose/profiles/#resolving-dependencies)

- [x] Added `newsfragments/resolve_profiles_dependencies.bugfix`

- [x] Added unit tests in `tests/unit/test_resolve_profiles.py` (all passing)

---

### Description

**Bug:**
When using `podman-compose --profile X up`, if a service in profile `X` contains a `depends_on` block pointing to another service `Y` that is NOT part of profile `X`, `podman-compose` crashes with a `KeyError: 'Y'`. The filtering of profiles happened before dependency trees were validated, causing targeted dependencies to be stripped entirely from parsed compose data.

**Fix:**
This PR updates the `_resolve_profiles` method to recursively walk through explicitly defined dependencies (via `depends_on` and `extends`). If an active service targets an excluded service, that target service is now securely pulled back into the active execution graph regardless of its underlying profile structure.

This ensures proper fallback parity with `docker-compose`'s Go implementation where "explicitly depended services are always started."

**Testing:**
Added comprehensive unit tests inside `test_resolve_profiles.py` which validate:
- Explicit dict/list structure inclusions for `depends_on`.
- Explicit dict/str structure inclusions for `extends`.
- Multi-tier transitive dependencies (e.g. `A` -> `B` -> `C`).
- Circular and non-existent edge-cases.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2810,7 +2810,8 @@ class PodmanCompose:
         the requested_profiles list.
 
         The returned service dictionary contains all services which do not include/reference a
-        profile in addition to services that match the requested_profiles.
+        profile in addition to services that match the requested_profiles, plus any services
+        they explicitly depend on.
 
         :param defined_services: The service dictionary
         :param requested_profiles: The profiles requested using the --profile arg.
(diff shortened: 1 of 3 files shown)


podman_compose.py
```diff
@@ -2819,11 +2820,39 @@ def _resolve_profiles(
...
+            dep_names = []
+
+            # Check depends_on
+            depends_on = config.get("depends_on", [])
+            if isinstance(depends_on, dict):
+                dep_names.extend(depends_on.keys())
+            elif isinstance(depends_on, list):
+                dep_names.extend(depends_on)
+
+            # Check extends
```
p12tic, line 2844: No need for this commend and similar one above. Seems obvious from the next line.

tests/unit/test_resolve_profiles.py
```diff
@@ -0,0 +1,117 @@
...
+    def setUp(self):
+        self.compose = PodmanCompose()
+
+    def test_no_profiles(self):
+        defined_services = {
+            "web": {"image": "nginx"},
+            "db": {"image": "postgres"},
+        }
+        services = self.compose._resolve_profiles(defined_services, {"test"})
+        self.assertIn("web", services)
```
p12tic, line 16: Better use something like `self.assertEqual(sorted(list(services.keys())), ["db", "web"])`. This is both clearer and harder to get wrong, especially in cases when you'd otherwise use assertNotIn.

p12tic (changes requested):
Sorry for delay in review. I had several small comments. Otherwise PR looks good, thanks!

PR #1458, opened 2026-05-17
feat(tui): Improve logging and output when pulling new images

febrezo:
Improve UX with a more colourful approach inspired by docker-compose for a fresher experience

imagen

Also, interruptions are grabbed accordingly.

imagen

Verbose outputs have also been beautified.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -22,11 +22,13 @@ import os
 import random
 import re
 import shlex
+import shutil
 import signal
 import string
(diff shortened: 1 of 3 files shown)


p12tic:
This PR does at least 3 things:
 - adds colored output
 - refactors things like PullResult
 - suppresses output from subprocesses
 
 These should be at least separate commits if not PRs. At the current state the PR is too hard to review.

PR #1462, opened 2026-05-20
Fix/include relative path resolution

### LINKED ISSUE #1301: Wrong relative paths in included files
**Describe the bug**

If you have relative paths in included compose files `podman-compose` does not adjust the paths to the root compose file.

**To Reproduce**

Assume the following structure

```yaml
# /dev.yml
include:
  - ./dev/db.yml
```

```yaml
# /dev/db.yml
services:
  db:
    image: ubuntu
    env_file:
      - ./db/.env
```

```env
# /dev/db/.env
FOO="bar"
```

1. just call `podman-compose -f dev.yml config

**Expected behavior**
The config output should look something like

```yaml
services:
  db:
    image:ubuntu
  env_file:
    ./dev/db/.env
```

or better

```yaml
services:
  db:
    image:ubuntu
  env_file:
    /path/to/my/work/directory/dev/db/.env
```

**Actual behavior**
The config looks like

```yaml
services:
  db:
    image:ubuntu
  env_file:
    ./db/.env
```

**Additional context**

It seems like the compose files are just merged. But the spec states, that:

```
Each path listed in the include section is loaded as an individual Compose application model, with its own project directory, in order to resolve relative paths.
```

So just merging the files and then processing it (as it seems to be implemented currently) will not work.

lytedev:
> **Disclaimer**: I authored this PR with LLM assistance. This description is written entirely by an LLM.

Fixes #1301.

## Compose Spec reference

[`include` — path resolution](https://github.com/compose-spec/compose-spec/blob/main/14-include.md):

> Each path listed in the `include` section is loaded as an individual Compose
> application model, with its own project directory, in order to resolve
> relative paths.

podman-compose previously resolved relative paths inside `include`d files
against the top-level project root instead of the included file's directory.

## The bug

Given:

```
project/
├── compose.yml         (include: [compose/sub.yml])
├── compose/sub.yml     (volumes: [../fixtures/seed.sql:/seed.sql:ro])
└── fixtures/seed.sql
```

The bind mount source resolved to `project/../fixtures/seed.sql` rather than
`project/fixtures/seed.sql`, and podman silently created an empty directory
at the wrong path. Same problem affected `env_file` and `build.context`.

## The fix

- Track which files in the parse loop arrived via `include:` (as opposed to
  via `-f` on the command line).
- For included files only, compute the file's directory relative to the
  project root and pass it as `sub_dir` to `normalize()`.
- `normalize_service` already had `sub_dir` handling for `volumes` and
  `build.context`; this PR extends it to `env_file` (string and
  `{path, required}` dict forms).

Files passed directly with `-f` keep their existing relative-path
behavior, so the legacy `-f`-merge semantics covered by
`tests/integration/multicompose` are preserved.

## Commits

1. **Extend `normalize_service` to rewrite env_file paths under `sub_dir`** —
   also threads `sub_dir` through `normalize()` (backward-compatible
   default). I judged the signature change too trivial to split into its
   own commit; happy to split if reviewers prefer.
2. **Resolve relative paths in included compose files against the included
   file** — the actual bug fix, plus newsfragment and an integration test.

## Verification

- All 473 unit tests pass at each commit (bisectable).
- New `tests/integration/include_relative_paths` integration test passes on
  this branch, **fails on `main`** with the exact bug signature
  (`./local.env` left unrewritten as `./local.env` instead of
  `sub/./local.env`).
- `tests/integration/multicompose` (which exercises legacy `-f` merge
  behavior) continues to pass — the fix is scoped to files added via
  `include:`.
- `ruff format`, `ruff check`, `pylint podman_compose.py` (10/10), `mypy .`
  all clean (the 4 mypy errors that exist are pre-existing on `main` —
  missing type stubs in unrelated test files).

## Known limitation (out of scope)

`${VAR:-default}` env-var interpolation runs *after* `normalize()`, so a
volume like `${X:-../config/foo}:/...` won't get the prefix applied even
with this fix. Worth a separate issue if anyone hits it.

### DIFF
--- a/tests/integration/include_relative_paths/test_podman_compose_include_relative_paths.py
+++ b/tests/integration/include_relative_paths/test_podman_compose_include_relative_paths.py
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: GPL-2.0
+
+"""
+Tests that relative paths (volumes, env_file) inside an included compose
+file are resolved against the included file's directory, not the project
+root. See https://github.com/containers/podman-compose/issues/1301 and the
+Compose Spec include resolution rules.
+"""
+
+import os
+import textwrap
+import unittest
+
+from tests.integration.test_utils import RunSubprocessMixin
+from tests.integration.test_utils import podman_compose_path
+from tests.integration.test_utils import test_path
+
+
+def compose_file() -> str:
+    return os.path.join(test_path(), "include_relative_paths", "docker-compose.yaml")
+
+
+class TestIncludeRelativePaths(unittest.TestCase, RunSubprocessMixin):
+    def test_relative_paths_in_included_file(self) -> None:
+        """
+        The included file at ``sub/included.yaml`` references
+        ``./local.env``, ``../shared.env``, ``./data:/data:ro`` and
+        ``../assets:/assets:ro``. After include resolution these must point
+        at paths under ``sub/`` for the ``./`` forms and at the project
+        root for the ``../`` forms.
+        """
+        out, _ = self.run_subprocess_assert_returncode([
+            "coverage",
+            "run",
+            podman_compose_path(),
+            "-f",
+            compose_file(),
+            "config",
+        ])
+
+        expected = textwrap.dedent("""\
+            name: include-relative-paths
+            services:
+              web:
+                env_file:
+                - ./sub/./local.env
+                - ./sub/../shared.env
+                image: nopush/podman-compose-test
+                volumes:
+                - ./sub/./data:/data:ro
+                - ./sub/../assets:/assets:ro
+            version: '3.6'
+
+            """)
+        self.assertEqual(out.decode("utf-8"), expected)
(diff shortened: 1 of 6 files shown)


tests/integration/include_relative_paths/test_podman_compose_include_relative_paths.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        out, _ = self.run_subprocess_assert_returncode([
+            "coverage",
+            "run",
+            podman_compose_path(),
+            "-f",
+            compose_file(),
+            "config",
+        ])
+        rendered = out.decode("utf-8")
```
p12tic: best to compare the returned string in full. This is especially important with things like assertNotIn as such assertions are very brittle and may not throw error if result is wrong in unexpected way.

p12tic (changes requested):
Small nitpick, but looks good otherwise. Thanks! Please rebase as there are merge conflicts.

tests/integration/include_relative_paths/test_podman_compose_include_relative_paths.py
lytedev: ah good call - updated!

lytedev:
> Small nitpick, but looks good otherwise. Thanks! Please rebase as there are merge conflicts.

Done! Thank you for reviewing!

p12tic (approved):

p12tic:
Looks great, thank you! Please resolve merge conflicts by rebasing on latest `main` and I'll be able to merge the PR.

lytedev:
Should be good to go!

p12tic (approved):

PR #1464, opened 2026-05-25
logs: Allow custom `container_name` as a log prefix

mokibit:
Previously, the `logs` command by default used the service name as the prefix, ignoring any custom `container_name` specified in `compose.yml`. Container number was also not provided with service name prefix. 
Docker-compose, by contrast, uses `container_name` as the prefix when available, and falls back to the service name concatenated with the container number.
 
This change brings `podman-compose logs` command behavior closer to `docker-compose`.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2854,6 +2854,11 @@ class PodmanCompose:
                     name = service_desc.get("container_name", name0)
                 else:
                     name = name0
+
+                if service_desc.get("container_name", False):
+                    log_prefix = name
+                else:
+                    log_prefix = f"{service_name}_{num}"
                 container_names_by_service[service_name].append(name)
                 # log(service_name,service_desc)
                 cnt = {
@@ -2861,6 +2866,7 @@ class PodmanCompose:
                     "name": name,
                     "num": num,
                     "service_name": service_name,
+                    "log_prefix": log_prefix,
                     **service_desc,
                 }
                 x_podman = service_desc.get("x-podman")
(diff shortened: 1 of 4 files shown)


podman_compose.py
```diff
@@ -4247,25 +4253,33 @@ async def compose_logs(
     max_service_length = 0
     tasks = []
     max_service_length = max(len(service) for service in args.services)
-    for i, service in enumerate(args.services):
-        # Add colored service prefix to output by piping output through sed
-        if args.no_log_prefix:
```
p12tic, line 4469: This code should not be put into additional loop because it would make code way more confusing.

podman_compose.py
```diff
@@ -4256,27 +4296,11 @@ async def compose_logs(
...
-            log_formatter = f"{color}[{service}]{space_suffix}|\x1b[0m"
 
-        podman_args_with_target = podman_args + container_names_by_service[service]
-        tasks.append(
-            asyncio.create_task(
-                compose.podman.run([], "logs", podman_args_with_target, log_formatter=log_formatter)
-            )
-        )
+    for _, service in enumerate(args.services):
+        tasks = collect_service_logs(compose, args, service, podman_args, tasks, max_service_length)
```
p12tic: You should not pass a list to collect_service_logs, but just return additional tasks and then join to single list at the caller.

podman_compose.py
```diff
@@ -3564,6 +3570,42 @@ def deps_from_container(args: argparse.Namespace, cnt: dict) -> set:
...
 
+def collect_service_logs(
+    compose: PodmanCompose,
+    args: argparse.Namespace,
+    service: str,
+    podman_args: list,
+    tasks: list,
+    max_service_length: int,
+) -> list:
+    for i, cnt in enumerate(compose.containers):
```
p12tic: You should have a function get_service_by_name or something like that

p12tic (approved):

PR #1435, opened 2026-04-10
fix symlink volumes for remote container

bam80:
For now, if volume source is a symlink, it's dereferenced, resulting to a change of the volume source path. 
If remote container host doesn't have the same path as the local symlink target, such volume can't mount and breaks the compose.

This scenario is commonly used in flatpaks, where $XDG_RUNTIME_DIR is full of flatpaks-specific symlinks:
```console
$ find $XDG_RUNTIME_DIR/ -type l -ls
      797      0 lrwxrwxrwx   1 bam      bam            17 апр  8 01:27 /run/user/1001/bus -> ../../flatpak/bus
      796      0 lrwxrwxrwx   1 bam      bam            19 апр  8 01:27 /run/user/1001/pulse -> ../../flatpak/pulse
      795      0 lrwxrwxrwx   1 bam      bam            23 апр  8 01:27 /run/user/1001/wayland-0 -> ../../flatpak/wayland-0
      794      0 lrwxrwxrwx   1 bam      bam            17 апр  8 01:27 /run/user/1001/app -> ../../flatpak/app
      793      0 lrwxrwxrwx   1 bam      bam            17 апр  8 01:27 /run/user/1001/doc -> ../../flatpak/doc
      792      0 lrwxrwxrwx   1 bam      bam            22 апр  8 01:27 /run/user/1001/.flatpak -> ../../flatpak/.flatpak
      791      0 lrwxrwxrwx   1 bam      bam            21 апр  8 01:27 /run/user/1001/p11-kit -> ../../flatpak/p11-kit
      790      0 lrwxrwxrwx   1 bam      bam            22 апр  8 01:27 /run/user/1001/flatpak-info -> ../../../.flatpak-info
```
Consider following typical volume declaration for Wayland socket pass-through:
```yaml
    volumes:
      - $XDG_RUNTIME_DIR/$WAYLAND_DISPLAY:$XDG_RUNTIME_DIR/$WAYLAND_DISPLAY:ro
```
If compose happens in flatpak, and `podman` is in the remote mode (`podman-compose --podman-path podman-remote` option or `remote = true` in [`containers.conf`](https://github.com/containers/container-libs/blob/main/common/docs/containers.conf.5.md)), this breaks `podman create` call as the remote system doesn't have local `/run/flatpak/wayland-0` path, without any feasible workarounds:
```console
$ podman-compose --verbose up
...
podman create ... -v /run/flatpak/wayland-0:/run/user/1001/wayland-0:ro ...
[...]

PR #1465, opened 2026-05-25
Fix link resolution for mounted volume

### LINKED ISSUE #1290: volume utilizing /var/run/docker.sock is improperly resolved using realpath, resulting in errors on MacOS
**Describe the bug**

OS: MacOS
CPU Arch: M1
podman-compose version: 1.5.0

When running a docker-compose that includes the following

```
volumes:
  - /var/run/docker.sock:/var/run/docker.sock
```

running `podman compose up` results in the following error:

```
Error: statfs /private/var/folders/0f/y6yk5ngd4r1_3wy0_5qf46xr0000gp/T/podman/podman-machine-default-api.sock: operation not supported
```

I believe this to be the result of calling `os.path.realpath` on the `mount_src` inside `assert_volume` rather than calling `os.path.abspath`.  Switching the `os.path.realpath` call out in favor of `os.path.abspath` appears to fix the issue for me locally.

Please make sure it's not a bug in podman (in that case report it to podman)
or your understanding of docker-compose or how rootless containers work (for example, it's normal for rootless container not to be able to listen for port less than 1024 like 80)

**To Reproduce**
1.  on MacOS M1
2. any docker-compose that includes a container with the following volume mount:
    ```
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    ```
3. podman compose up

**Expected behavior**
Container is created and volume is mounted to the container

**Actual behavior**
```
Error: statfs /private/var/folders/0f/y6yk5ngd4r1_3wy0_5qf46xr0000gp/T/podman/podman-machine-default-api.sock: operation not supported
```

**Output**

```
$ podman-compose version
podman-compose version 1.5.0
podman version 5.6.0
podman --version 
podman version 5.6.0

$ podman-compose up
...

```

**Environment:**
 - OS: Mac
 - podman version: 5.6.0

**Additional context**

mokibit:
This PR builds on https://github.com/containers/podman-compose/pull/1291 and https://github.com/containers/podman-compose/pull/1435.
I cleaned up the integration test and added a newsfragment file.

Fixes https://github.com/containers/podman-compose/issues/1290

### DIFF
--- a/tests/integration/vol/short_syntax/test_podman_compose_vol_short_syntax.py
+++ b/tests/integration/vol/short_syntax/test_podman_compose_vol_short_syntax.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-2.0
 
+import json
 import os
 import shutil
 import unittest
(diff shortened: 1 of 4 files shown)


tests/integration/vol/short_syntax/test_podman_compose_vol_short_syntax.py
```diff
@@ -53,3 +54,50 @@ def test_source_host_dir(self, source_host_dir_exists: bool, service_name: str)
...
+
+    def test_volume_source_is_symlink(self) -> None:
+        try:
+            base = "/tmp/symlink-test"
+            real_dir = os.path.join(base, "real")
+            link_dir = os.path.join(base, "link")
+
+            os.makedirs(real_dir, exist_ok=True)
+
+            # Create the symlink: link_dir -> real_dir
```
p12tic: Unnecessary comment.

tests/integration/vol/short_syntax/test_podman_compose_vol_short_syntax.py
```diff
@@ -53,3 +54,50 @@ def test_source_host_dir(self, source_host_dir_exists: bool, service_name: str)
...
+            out, _ = self.run_subprocess_assert_returncode(
+                [
+                    "podman",
+                    "inspect",
+                    "short_syntax_test3_1",
+                ],
+                0,
+            )
+
+            inspect_data = json.loads(out)
```
p12tic: no point in having inspect_data and actual_source variables

p12tic (approved):

PR #1469, opened 2026-06-01
Fixed Windows file paths in 'context' attribute being treated as URLs

jedKnighthook:
This PR fixes normalized, absolute windows file paths being detected and treated as URLs and thus skipping the part of the code that uses the path locally to specify custom `dockerfile` and `dockerfile_inline` temp file path names when building via compose.

There was some confusion when looking at #1332, #1312, and the other PR I was following that was abandoned [PR 1372](https://github.com/containers/podman-compose/pull/1372) as it seemed to work for some and not others. From my own testing both the `dockerfile` and `dockerfile_inline` attributes work perfectly when running podman-compose on a linux based system but then have these issues on Windows due to the `is_context_git_url()` function check misinterpreting windows file paths as URLs because of the `:` being present.


podman_compose.py
```diff
@@ -3232,7 +3232,9 @@ def is_context_git_url(path: str) -> bool:
     if r.scheme in ('git', 'http', 'https', 'ssh', 'file', 'rsync'):
         return True
     # URL contains a ":" character, a hint of a valid URL
-    if r.scheme != "" and r.netloc == "" and r.path != "":
+    # But also detects windows file paths (e.g. "C:\path\to\contextdir") as urls
+    localdir = os.path.isdir(path)
```
p12tic: Needs unittest. Best to patch os.path with ntpath.

p12tic (changes requested):
Needs unittest, but otherwise looks good. Thanks!

jedKnighthook:
When adding the unit test I noticed what appeared to be duplicated unittest files related to this commit: https://github.com/containers/podman-compose/commit/dd471c891831833c97dce2b051e479817727630a so I removed the duplicate and renamed the original file as it looks like it was intended to be updated which required a second commit to gracefully rename the unittest as the same name as the removed file while preserving git history.

podman_compose.py
```diff
@@ -3243,7 +3249,14 @@ def is_context_git_url(path: str) -> bool:
     if r.scheme in ('git', 'http', 'https', 'ssh', 'file', 'rsync'):
         return True
     # URL contains a ":" character, a hint of a valid URL
-    if r.scheme != "" and r.netloc == "" and r.path != "":
+    # But also detects windows file paths (e.g. "C:\path\to\contextdir") as urls
+    is_path_with_drive_letter = (
+        (primarypathisabs(path) or secondarypathisabs(path))
```
p12tic: Better just call os.path.isabs for primarypathisabs

p12tic (changes requested):
Just a small nit. After it the PR will be merged.

podman_compose.py
jedKnighthook: that's what I was initially doing and that works when running the tool but it only correctly loads and detects the type of file paths for the OS it's running on. In the case of the unit test it will fail the windows tests when running on linux and vice versa which is why I forcibly check using both `posixpath.isabs` and `ntpath.isabs` here.

jedKnighthook: oh, do you mean rather than importing the lib a second time as `primarypathisabs`?

p12tic: > oh, do you mean rather than importing the lib a second time as `primarypathisabs`?

Yes, exactly

p12tic (approved):

PR #1472, opened 2026-06-06
Activate profiles of a given targeted service or services

### LINKED ISSUE #930: Building or Running a Service Fails Without Explicit Profile

mokibit:
This PR is based on https://github.com/containers/podman-compose/pull/1297, thanks to https://github.com/Kache.

Previously, `podman-compose` did not automatically activate the profiles of explicitly targeted services. This change ensures that targeting a service adds its profile to the active set, aligning behavior with `docker-compose`.

Per the Docker Compose documentation:
    > A service is ignored by Compose when none of the listed profiles
    > match the active ones, unless the service is explicitly targeted by
    > a command. In that case its profile is added to the set of active
    > profiles.
    
https://docs.docker.com/reference/compose-file/profiles/

Fixes https://github.com/containers/podman-compose/issues/930.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2575,6 +2575,15 @@ class PodmanCompose:
         }
         requested_profiles = set(args.profile).union(profiles_from_env)
 
+        target_service = getattr(
+            args, "service", None
+        )  # example: command `run` can only have one service
+        target_services = getattr(
+            args, "services", None
+        )  # example: command `build` can have several services
+
+        target = [target_service] if target_service else target_services or []
+
         compose: dict[str, Any] = {}
         # Iterate over files primitively to allow appending to files in-loop
         files_iter = iter(files)
@@ -2661,7 +2670,7 @@ class PodmanCompose:
                 # having `include` present and correctly processed in included files
                 del compose["include"]
         resolved_services = self._resolve_profiles(
-            compose.get("services") or {}, requested_profiles
+            compose.get("services") or {}, target, requested_profiles
         )
         compose["services"] = resolved_services
         if not getattr(args, "no_normalize", None):
@@ -2685,7 +2694,7 @@ class PodmanCompose:
         if not services:
             log.warning("WARNING: No services defined")
         # include services with no profile defined or the selected profiles
-        services = self._resolve_profiles(services, requested_profiles)
+        services = self._resolve_profiles(services, target, requested_profiles)
 
         # NOTE: maybe add "extends.service" to _deps at this stage
         flat_deps(services, with_extends=True)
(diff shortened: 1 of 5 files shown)


tests/integration/profile/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM busybox
```
p12tic: Use nopush/podman-compose-test

podman_compose.py
```diff
@@ -2814,21 +2823,31 @@ def _parse_compose_file(self) -> None:
...
+        or services.
 
         :param defined_services: The service dictionary
         :param requested_profiles: The profiles requested using the --profile arg.
+        :param target: Name of service or services targeted by the current command
         """
         if requested_profiles is None:
             requested_profiles = set()
 
+        if target:
```
p12tic: It would make sense for `target` to always be a list. We already do decision on what target services is, all different scenarios can be regularized into a list with no, 1 or many elements.

p12tic (approved):

PR #1473, opened 2026-06-06
Support `configs:` from Compose spec

### LINKED ISSUE #835: Support `configs:`
I tried to run https://git.sr.ht/~emersion/sr.ht-container-compose/ which requires `configs:` support. However, I always got "No such file or directory" errors being displayed for all defined configs.

Full support for `configs:` would be the preferred solution. :smiley: 

I don't really think there is an alternative to supporting configs, because, if a compose file uses (and thus requires) them, you cannot use `podman-compose` right now.

**Additional context**
https://docs.docker.com/compose/compose-file/05-services/#configs
https://docs.docker.com/compose/compose-file/08-configs/

**Minimal example**
```bash
$ cat compose.yaml 
configs:
  my_config:
    file: ./my_config.ini
services:
  busybox:
    image: busybox:latest
    command: /bin/sh -c "id; ls -lash /; cat /my_config.ini"
    configs:
      - my_config
$ cat my_config.ini 
[config]
key=value
$ podman compose up
>>>> Executing external compose provider "/usr/bin/podman-compose". Please refer to the documentation for details. <<<<

podman-compose version: 1.0.6
['podman', '--version', '']
using podman version: 4.9.0
...
podman start -a podman-compose-test_busybox_1
[busybox] | uid=0(root) gid=0(root) groups=10(wheel)
[busybox] | total 4K     
[busybox] |       0 dr-xr-xr-x   14 root     root          26 Feb  7 21:19 .
[busybox] |       0 drwxr-xr-x    2 root     root        4.6K Jan 17 21:58 bin
[busybox] |       0 drwxr-xr-x    5 root     root         340 Feb  7 21:19 dev
[busybox] |       0 drwxr-xr-x    3 root     root          56 Feb  7 21:19 etc
[busybox] |       0 drwxr-xr-x    2 nobody   nobody         0 Jan 17 21:59 home
[busybox] |       0 drwxr-xr-x    2 root     root         270 Jan 17 21:58 lib
[busybox] |    4.0K lrwxrwxrwx    1 root     root           3 Jan 17 21:58 lib64 -> lib
[busybox] |       0 dr-xr-xr-x  353 nobody   nobody         0 Feb  7 21:19 proc
[busybox] |       0 drwx------    2 root     root           0 Jan 17 21:59 root
[busybox] |       0 drwxr-xr-x    3 root     root          80 Feb  7 21:19 run
[busybox] |       0 dr-xr-xr-x   13 nobody   nobody         0 Feb  7 08:09 sys
[busybox] |       0 drwxrwxrwt    2 root     root           0 Jan 17 21:59 tmp
[busybox] |       0 drwxr-xr-x    4 root     root          14 Jan 17 21:59 usr
[busybox] |       0 drwxr-xr-x    4 root     root          16 Jan 17 21:59 var
cat: can't open '/my_config.ini': No such file or directory
exit code: 1

### PR DESCRIPTION
**Pull request by overhacked**
`configs:` and `services.*.configs:` allow content specified in the compose file, in environment variables, in local files, and in OCI artifacts to be mounted in the container. They are similar to volume-mounting a single file but with more options for where the content is stored. See https://github.com/compose-spec/compose-spec/blob/main/08-configs.md and https://github.com/compose-spec/compose-spec/blob/main/05-services.md#configs for details.

Resolves #835.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2910,6 +2910,16 @@ class PodmanCompose:
         ]
         # other top-levels:
         self.declared_configs = compose.get("configs", {})
+        if (
+            self.declared_configs
+            and self.podman_version is not None
+            and strverscmp_lt(self.podman_version, "5.6.0")
+        ):
+            raise PodmanComposeError(
+                "Compose configs are only supported when running podman >= 5.6.0 "
+                f"(current version {self.podman_version})"
+            )
+
         self.declared_secrets = compose.get("secrets", {})
         given_containers = []
         container_names_by_service: dict[str, list[str]] = {}
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -585,7 +585,7 @@ async def assert_config_artifact(
     source_env = config.get("environment")
     source_content = config.get("content")
     source_external = config.get("external", False)
-    if not bool(source_file) ^ bool(source_env) ^ bool(source_content) ^ source_external:
+    if bool(source_file) + bool(source_env) + bool(source_content) + source_external != 1:
         raise ValueError(
             f'ERROR: Config "{config_name}" must specify one and only one of file, '
             'environment, content, or external.'
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -613,7 +613,7 @@ async def assert_config_artifact(
                 f"External config [{artifact_name}] does not exist. "
                 f"Create it first with: podman artifact add '{artifact_name}'"
             ) from e
-    except (json.JSONDecodeError, KeyError) as e:
+    except (json.JSONDecodeError, KeyError, TypeError) as e:
         raise PodmanComposeError(
             f'Artifact {artifact_name} exists, but the Manifest.layers[0].digest '
             f'field could not be parsed. Check the output of: '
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: podman_compose.py [hunk 1]
```diff
@@ -567,6 +567,113 @@ def try_parse_bool(value: Any) -> bool | None:
...
+    create artifact if needed
+
+    Returns:
+        Artifact name and tag, or None if file source
+    """
+    source_file = config.get("file")
+    source_env = config.get("environment")
+    source_content = config.get("content")
+    source_external = config.get("external", False)
+    if not bool(source_file) ^ bool(source_env) ^ bool(source_content) ^ source_external:
```
p12tic: This is incorrect if more than 2 values are specified.

Better do something like `if int(bool(...)) + int(bool(...) + ... != 1:`

p12tic:
Rather large PR, will take some time to review.

podman_compose.py
overhacked: Yeah, don't know what I was thinking there. Have fixed and pushed a commit with your suggestion.

overhacked:
Understandable. Most of the line count is tests, though. Changes to `podman_compose.py` core adds <250 lines.

ketozhang:
@p12tic Could the open community contribute a limited-scope review or something else that helps your workload here?

overhacked:
Hello, checking in if this could still be in consideration, understanding that everyone's time is limited and there are competing priorities. Thanks!

PR #1476, opened 2026-06-16
Fix transitive `depends_on` with oneshot/healthy conditions and podma…

allen-munsch:
…n 5.x/6.x dependency graph bugs

Changes:
Only add `--requires` for `service_started` (RUNNING) dependencies. Oneshot (`service_completed_successfully`) and health-checked (`service_healthy`) dependencies are handled by `check_dep_conditions` at start time, avoiding broken transitive `--requires` chains when intermediate containers have already exited (podman 5.x dependency graph bug).

Parse `required` and `restart` properties from `depends_on` entries (default `true`). `required: false` allows a dependency to be absent without error. `restart: false` prevents a dependency change from restarting dependents.

Batch-start all containers in a single `podman start` call so podman can resolve the full `--requires` graph atomically, fixing a podman 6.x regression where one-at-a-time starts fail when a transitive dependency is already running.

After batch start, retry containers stuck in "Created" state up to 3 times — podman's graph resolver can fail on them when transitive dependencies are already running, but a retry after the batch succeeds.

`check_dep_conditions` waits per-container with `asyncio.gather` instead of a single blocking loop for all deps, so a missing non-required dependency doesn't block others.

Only restart dependents when `dep.restart` is `True`.

Tests: 423/423 unit tests pass. Manual integration verified against podman 4.9.3 linux/amd64, 5.4.2 linux/arm64 engine.

several representative matrices were tested via nekto act

---

repro compose yaml:

```
version: "3"
# Minimum reproduction of transitive depends_on bug with oneshot containers.
#
# Podman 5.x: fails when an intermediate container (init) exits after
# completing — podman's dependency graph resolver can't handle a --requires
# chain where a middle node has already finished.
#
# Podman 6.x: regression podman#25633 — starting one-at-a-time fails when
# a transitive dependency is already running.
#
# This compose works with the patch applied because:
# 1. Only service_started (RUNNING) deps get --requires; oneshot/healthy
#    deps are handled at start time by check_dep_conditions.
# 2. Containers are batch-started so podman sees the full dependency graph.
# 3. Containers stuck in "Created" state (after batch-start fails on a
#    transitive dep) are retried individually.

services:
  server:
    image: docker.io/alpine:latest
    command: ["sleep", "infinity"]

  init:
    image: docker.io/alpine:latest
    command: ["echo", "init done"]
    depends_on:
      server:
        condition: service_started

  app:
    image: docker.io/alpine:latest
    command: ["sleep", "infinity"]
    depends_on:
      init:
        condition: service_completed_successfully

```


allen-munsch:
sorry for burning up build minutes

its just a rough sketch, but was a big friction point, seemed to fix the issue, doing a drop in replacement of docker

ran into some issues with some cert gen sidecars when switching over

podman_compose.py
```diff
@@ -1590,9 +1597,13 @@ def from_value(cls, value: str) -> ServiceDependencyCondition:
 
 
 class ServiceDependency:
-    def __init__(self, name: str, condition: str) -> None:
```
p12tic, line 1654: Seems like we should use dataclass for this.

tests/unit/test_depends_on.py
```diff
@@ -51,3 +52,24 @@ def test_flat_deps(
...
+                        "condition": "service_healthy",
+                        "required": False,
+                        "restart": False,
+                    }
+                }
+            },
+        }
+        flat_deps(services)
+        b_deps = services["service_b"]["_deps"]
+        self.assertEqual(len(b_deps), 1)
```
p12tic: Better just do like this:

`assertEqual(services["service_b"]["_deps"], [ServiceDependency(...)])`

podman_compose.py
```diff
@@ -3610,32 +3642,42 @@ async def check_dep_conditions(compose: PodmanCompose, deps: set) -> None:
...
+                        )
+                        log.debug(
+                            'Podman wait returned an error (%d) when executing "%s": %s',
+                            _exc.returncode,
+                            _exc.cmd,
+                            output,
+                        )
+                    await asyncio.sleep(1)
+
+            await asyncio.gather(*(wait_one(cnt) for cnt in deps_cd))
```
p12tic: This is unrelated change to the fix. Should be at least separate commit with separate release note.

allen-munsch: :+1:  splitting this out

https://github.com/containers/podman-compose/pull/1479

Thank you for the review :pray:

p12tic (approved):

p12tic:
Looks great, thank you! Please resorve merge conflict and I can then merge the PR.

allen-munsch:
resolved the conflict, is this flake, related, or a separate thing?

PR #1477, opened 2026-06-18
fix: normalize after resolving extends

theSuess:
This fixes an issue with the order of resolving compose file extensions. Normalization expands relative paths and fills in defaults which should be taken from the base file as well

### DIFF
--- a/newsfragments/normalize_after_resolving_extends.bugfix
+++ b/newsfragments/normalize_after_resolving_extends.bugfix
@@ -1,0 +1,1 @@
+Fixed order of operations when populating defaults and using `extends` making sure the context points at the right directory
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2126,11 +2126,10 @@ def normalize_service_final(service: dict[str, Any], project_dir: str) -> dict[s
     return service
 
 
-def normalize_final(compose: dict[str, Any], project_dir: str) -> dict[str, Any]:
-    services = compose.get("services", {})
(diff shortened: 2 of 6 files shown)


p12tic:
Looks good. Do you have a reproduction case where the current code breaks and your fixed version works?

Could you also add a release note in newsfragments/ directory?

theSuess:
Sorry for the late reply, this is the specific error case:
```
.
├── base
│   ├── docker-compose.yml
│   └── Dockerfile
└── docker-compose.yml
```

```
# docker-compose.yml in root folder
services:
  example:
    extends:
      file: base/docker-compose.yml
      service: example
    build:
      args:
        alpine_version: 3.23
```

```
# docker-compose.yml in ./base
services:
  example:
    build:
      context: .
    command: ["cat", "/etc/alpine-release"]
```

```
ARG alpine_version=3.24
FROM alpine:${alpine_version}
```

Running `podman-compose up --build` at the root level fails as the presence of `services.example.build.args` causes the entire `build` section to be filled with missing defaults. `build.context` defaults to the current working directory which causes the command to fail with this error:

```
OSError: Dockerfile not found in <path-to-root-dir>
````

By making sure the values from `extends` are applied first, `context` points to the right dir and the build succeeds

(also added a release note)

p12tic:
@theSuess Thanks. Sorry for a late comment - could you add an integration test with the reproduction data in your comment.

theSuess:
Also sorry for the late reply :sweat_smile: I've modified the existing subdir integration test for extends to include build args as that test is covering the same subject but didn't catch this edge case

Issue #575, opened 2022-10-19
Support depends_on type service_completed_successfully


micheljung:
**Is your feature request related to a problem? Please describe.**

The problem is that podman-compose is incompatible with current docker-compose.yml because it does not support `service_completed_successfully`.

According to the [compose spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#depends_on)

- `condition`: condition under which dependency is considered satisfied
  - `service_started`: is an equivalent of the short syntax described above
  - `service_healthy`: specifies that a dependency is expected to be "healthy"
    (as indicated by [healthcheck](#healthcheck)) before starting a dependent
    service.
  - `service_completed_successfully`: specifies that a dependency is expected to run
    to successful completion before starting a dependent service.

**Describe the solution you'd like**
podman-compose implements `service_completed_successfully` as per the spec.

**Describe alternatives you've considered**
None

**Additional context**
I didn't check if the other conditions are supported yet, but they should also be.

mbergen:
It seems like `service_healthy` also does not work.
`service_started` is the equivalent to no condition (default) and therefore works.
Would be great to get these working!

mdellweg:
Actually, i think podman waits for the healthiness anyway, so anything one specifies behaves like `service_healthy`.

mbergen:
> Actually, i think podman waits for the healthiness anyway, so anything one specifies behaves like `service_healthy`.

Last time i checked it podman was not executing any healthchecks, it was only waiting for services to be started. That's a huge difference.

Greenall:
I'm on the following versions and can confirm that service_healthy condition seems to be ignored and treated as service_started. 
podman-compose version 1.0.6
podman version 4.4.1

kontsaki:
any status update on this feature?

any way to workaround this?

w568w:
@kontsaki Have a look at #453, which has a preliminary implementation of `depends_on`.

**TL;DR (maybe inaccurate!):** `podman-compose` is calling `podman` commands directly to do all stuffs (e.g. executing `podman start foo --bar`), which do not provide a way of "depending on" another container, because it needs a daemon to do that (as Docker does). There are some workarounds:

- `podman-compose` waits for all dependencies to start (dirty, but has been done in #453)
- rely on `podman` socket and use its API to start instead of calling it directly (see discussion in <https://github.com/containers/podman-compose/discussions/456>; community has no motivation to do this)
[...]

PR #1481, opened 2026-06-22
Fix `service_completed_successfully` to require exit code 0

### LINKED ISSUE #1330: `depends_on` service_completed_successfully isn't always followed
**Describe the bug**
The `depends_on:` attribute with value "service_completed_successfully" isn't always honored, and containers may start out of their desired order.

**To Reproduce**

Use this compose file:
```
---
services:  
  init:
    container_name: init
    image: busybox  
    command: sh -c "echo 'Init started' && sleep 10 && echo 'Init complete'"
      
  app:
    container_name: app
    image: busybox  
    command: sh -c "echo 'App started' && echo 'App complete'"  
    depends_on:
      init:  
        condition: service_completed_successfully
```

Run `podman-compose up` 5-6 times.

**Expected behavior**

The `init` container will complete and print all of its text before `app` starts.

**Actual behavior**

Intermittently, the `app` container will start just after the `init` container starts and finish before `init` finishes.

**Output**

Undesired behavior:
```
$ podman-compose up
4d780508231db7903bfab54e90a1cbe0924d39aa9fcb174a8df35eff51bd572b
ec25a6b589bebd281076cf029ed0ce6ef764a9e4fbb2e976b33cf67afa932365
e2c5d07059a2b3eb52fe6ca6482269c57b8f152c14cbcdd045a39f5f8310b954
[init] | Init started
[app]  | App started
[app]  | App complete
[init] | Init complete
```

Desired behavior:
```
$ podman-compose up
468b853baa118eaa05468e3521d62f4f92170c9c520fe3ae01d5e5eef808d28b
5f57810587c918410f453c4835c6179108370149ff028d42d9c87ea4d632976b
e8148764fb3d0c2a1fbc5a2b8d23b084d2aac94bb9480ed5c231cacec47b5cee
[init] | Init started
[init] | Init complete
[app]  | App started
[app]  | App complete
```

**Environment:**
 - OS: Linux 6.16.8-200.fc42.x86_64 GNU/Linux
 - podman-compose version 1.5.0
 - podman version 5.6.2

**Additional context**

This behavior can be reproduced using `podman` alone. If `podman wait` is run immediately after `podman start` then it might not be enough time for the container to have actually started (i.e. "stopped" is the initial state of a container). A delay of only 0.01 seconds (i.e. `sleep 0.01`) is usually enough to overcome this race condition.

I'll also probably file a bug in Podman itself. Seems reasonable that a created container might have a different status (e.g. "created") compared to another container that actually ran and exited successfully.

mokibit:
Fix `service_completed_successfully` to verify the dependency exits with code 0, per the Docker Compose spec. Previously, it only checked that the container had stopped, ignoring non-zero exit codes and allowing dependent services to start even if the dependency failed.

Docker Compose implementation reference:
https://github.com/docker/compose/blob/v2.29.7/pkg/compose/convergence.go#L433

This PR builds on https://github.com/containers/podman-compose/pull/1337 thanks to [mrIncompetent](https://github.com/mrIncompetent).

Fixes https://github.com/containers/podman-compose/issues/1330, https://github.com/containers/podman-compose/issues/575.

### DIFF
--- a/tests/integration/deps/test_podman_compose_deps.py
+++ b/tests/integration/deps/test_podman_compose_deps.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: GPL-2.0
+import json
 import os
 import unittest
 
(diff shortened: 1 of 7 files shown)


tests/integration/deps/test_podman_compose_deps.py
```diff
@@ -192,6 +192,69 @@ def test_deps_fails(self) -> None:
...
+            ])
+
+            output, _ = self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_yaml_path(suffix),
+                "ps",
+            ])
+
+            self.assertIn(b"oneshot", output)
```
p12tic: let's use less assertIn if possible

p12tic (approved):

Issue #466, opened 2022-03-31
Recreate container if image is updated


elya5:
**Describe the bug**
I'm not sure whether this is actually considered a bug but I expected podman-compose to destroy and create a new container when the image is updated and the `up` command is called. Might be related to #56.

**To Reproduce**
1. `podman-compose up -d`
2. `podman-compose build`
3. `podman-compose up -d`

**Expected behavior**
Recreation of containers with outdated image.

**Actual behavior**
Old containers keep running.

**Environment:**
 - OS: Fedora 35
 - podman version: 3.4.4
 - podman compose version: 1.0.3

mokibit:
I cannot reproduce this issue anymore with this configuration:
Environment:
```
Debian GNU/Linux 13 (trixie)
Python 3.13.5
podman-compose version 1.6.0
podman version 5.4.2
```
`compose.yml` file:
```
services:
  app:
    build:
      context: .
      dockerfile: Dockerfile.image_change
    image: nopush/podman-compose-test:build-image-change
    command: ["sleep", "infinity"]
```
`Dockerfile.image_change`:
```
FROM busybox
```
Run command `podman-compose up -d --build`
Check container ID through the command `podman-compose ps --format json`:
```
"Id": "24196cedabdda14320a<...>"
```
Now modify the Dockerfile to produce a different image:
```
FROM busybox
RUN touch /marker
```
Rebuild and restart: run the command `podman-compose up -d --build` again and check the container ID again through the command `podman-compose ps --format json`:
```
<...>
"Exited": false,
<...>
"Id": "74e38558b968799f3<...>"
<...>
"State": "running",
```
We can see a different container ID and running container; therefore, `app` is recreated.
This means `podman-compose up -d` now recreates any container whose image has changed, even when the compose file itself is unchanged.

Fixed by https://github.com/containers/podman-compose/pull/1484.

PR #1484, opened 2026-06-27
detect image changes and recreate affected containers

### LINKED ISSUE #1453: up -d does not recreate container after image rebuild (stale image)
**Describe the bug**

After rebuilding an image (`podman-compose build <service>`), running `podman-compose up -d --no-deps <service>` does NOT recreate the container with the new image. The existing container keeps running with the old image, even though `podman images` confirms the image tag now points to a new image ID.

This differs from Docker Compose, where `up -d` detects the image change and auto-recreates the container.

**To Reproduce**

1. Start a service:
   ```bash
   podman-compose --in-pod=false up -d api
   ```

2. Verify the image ID:
   ```bash
   podman inspect api --format '{{.Image}}'
   # e.g. 5c64194536d1...
   ```

3. Rebuild with a changed source file (producing a new image):
   ```bash
   podman-compose --in-pod=false build --build-arg BUILD_DATE=$(date -u +%Y-%m-%dT%H:%M:%S) api
   # New image: 9a10b264aac6...
   ```

4. Verify the tag points to the new image:
   ```bash
   podman images localhost/portfolio-manager_api --format '{{.ID}} {{.Tag}}'
   # 9a10b264aac6 dev
   ```

5. Try to recreate:
   ```bash
   podman-compose --in-pod=false up -d --no-deps api
   # No output, no recreation
   ```

6. Check — container still uses old image:
   ```bash
   podman inspect api --format '{{.Image}}'
   # Still 5c64194536d1... (old image)
   ```

**Expected behavior**

Container should be recreated with the new image, matching Docker Compose behavior.

**Workaround**

Manually stop + remove + up:
```bash
podman stop api
podman rm -f api
podman-compose --in-pod=false up -d --no-deps api
```

This is fragile because `podman rm` fails when dependent containers exist (see #1177), forcing reverse-dependency-order removal.

**Environment**
- podman-compose version: 1.5.0
- podman version: 5.8.2
- OS: Linux (Hostinger VPS, Ubuntu 24.04)
- `--in-pod=false` used throughout

**Root cause hypothesis**

podman-compose may be comparing the container's current image ID to the compose config's image tag, but resolving the tag at config-parse time (before the build). After a rebuild, the tag resolves to a new ID, but compose's cached config still references the old one. Alternatively, compose may not be checking image ID at all and only recreating when the compose YAML changes.

thrawn-sh:
podman-compose up currently only compares the YAML config hash to decide whether a container needs to be recreated.  This misses the case where an image has been updated (e.g. after a pull) while the compose file itself stays the same — the container keeps running the old image.

Add an image_id field to ExistingContainer, populated from the ImageID returned by podman ps.  Before the recreation loop, resolve the current local image ID for each service via podman inspect and compare it against the running container's image ID.  When they differ, mark the service for recreation.

Fixes: #1453, #466

### DIFF
--- a/tests/integration/base_image/Dockerfile.test2
+++ b/tests/integration/base_image/Dockerfile.test2
@@ -1,0 +1,2 @@
+FROM nopush/podman-compose-test
+RUN touch /nopush-podman-compose-test2
--- a/tests/integration/compose_up_behavior/test_compose_up_behavior.py
+++ b/tests/integration/compose_up_behavior/test_compose_up_behavior.py
@@ -488,3 +488,162 @@ class TestComposeUpBehavior(unittest.TestCase, RunSubprocessMixin):
             self.run_subprocess_assert_returncode(
                 ["podman", "rmi", "-f", image],
             )
+
+    def test_recreate_on_image_changed(self) -> None:
+        """Verify that containers are recreated when their image changes.
+
+        Simulates pulling a new version of an image by re-tagging a different
+        base image under the same tag. Only the service whose image changed
+        should be recreated; the unchanged service should keep its container.
+        """
+        compose_file = compose_yaml_path("image_change")
+        tag = "nopush/podman-compose-test:image-change-app"
+
+        try:
+            # create the tagged image from the existing base image
+            self.run_subprocess_assert_returncode([
+                "podman",
+                "tag",
+                "nopush/podman-compose-test",
+                tag,
+            ])
+
+            # start containers
+            self.run_subprocess_assert_returncode(
+                [podman_compose_path(), "-f", compose_file, "up", "-d"],
+            )
+
+            original_containers = self.get_existing_containers("image_change")
+
+            # simulate a new image version: re-tag a different image under the same tag
+            self.run_subprocess_assert_returncode([
+                "podman",
+                "tag",
+                "nopush/podman-compose-test2",
+                tag,
+            ])
+
+            # verify the app container exists
+            self.assertIn(
+                "compose_up_behavior_app_1",
+                original_containers,
+                "app container should exist after initial up",
+            )
+
+            # run up again — should detect the changed image and recreate app
+            self.run_subprocess_assert_returncode(
+                [podman_compose_path(), "-f", compose_file, "up", "-d"],
+            )
+
+            new_containers = self.get_existing_containers("image_change")
+
+            # app should be recreated (different container ID)
+            self.assertNotEqual(
+                original_containers["compose_up_behavior_app_1"]["id"],
+                new_containers["compose_up_behavior_app_1"]["id"],
+                "app container should be recreated when its image changes",
+            )
+
+            # db should NOT be recreated (same container ID)
+            self.assertEqual(
+                original_containers["compose_up_behavior_db_1"]["id"],
+                new_containers["compose_up_behavior_db_1"]["id"],
+                "db container should not be recreated when its image did not change",
+            )
+
+            # all containers should be running
+            self.assertTrue(
+                all(c["exited"] is False for c in new_containers.values()),
+                "Not all containers are running after up command",
+            )
+
+        finally:
+            self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_file,
+                "down",
+                "-t",
+                "0",
+            ])
+            self.run_subprocess_assert_returncode(
+                ["podman", "rmi", "-f", tag],
+            )
+
+    def test_recreate_on_build_image_changed(self) -> None:
+        """Verify that containers are recreated when a locally built image changes.
+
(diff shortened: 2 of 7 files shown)


tests/integration/compose_up_behavior/test_compose_up_behavior.py
```diff
@@ -488,3 +488,167 @@ def test_localhost_image_built_if_does_not_exist(self) -> None:
...
+                [podman_compose_path(), "-f", compose_file, "up", "-d"],
+            )
+
+            original_containers = self.get_existing_containers("image_change")
+
+            # simulate a new image version: re-tag a different image under the same tag
+            self.run_subprocess_assert_returncode([
+                "podman",
+                "pull",
+                "docker.io/library/alpine:latest",
```
p12tic: Better create `nopush/podman-compose-test2`. Each pull uses docker free user quota. When running tests many times it's possible to exhaust it.

tests/integration/compose_up_behavior/test_compose_up_behavior.py
```diff
@@ -488,3 +488,167 @@ def test_localhost_image_built_if_does_not_exist(self) -> None:
...
+            # run up again — should detect the changed image and recreate app
+            self.run_subprocess_assert_returncode(
+                [podman_compose_path(), "-f", compose_file, "up", "-d"],
+            )
+
+            new_containers = self.get_existing_containers("image_change")
+
+            # app should be recreated (different container ID)
+            self.assertNotEqual(
+                original_containers.get("compose_up_behavior_app_1", {}).get("id"),
```
p12tic: Don't use `.get(...)` because we want test to fail if there's a typo in `compose_up_behavior_app_1`.

p12tic (changes requested):
Thanks. Several small nitpicks, but otherwise looks good.

tests/integration/compose_up_behavior/test_compose_up_behavior.py
thrawn-sh: done, also in tests

tests/integration/compose_up_behavior/test_compose_up_behavior.py
thrawn-sh: @p12tic: I added another nopush image to the tests, next to the current one, did you want it like this, or as part of the unittest itself?

tests/integration/compose_up_behavior/Dockerfile.image_change
```diff
@@ -0,0 +1 @@
+FROM busybox:latest
```
p12tic: Would `FROM nopush/podman-compose-test` work here?

tests/integration/base_image/Dockerfile.test2
```diff
@@ -0,0 +1,7 @@
+FROM docker.io/library/debian:bookworm-slim
```
p12tic: Would `FROM nopush/podman-compose-test` work here?

tests/integration/compose_up_behavior/test_compose_up_behavior.py
```diff
@@ -488,3 +488,162 @@ def test_localhost_image_built_if_does_not_exist(self) -> None:
...
+
+            self.assertIn(
+                "compose_up_behavior_app_1",
+                original_containers,
+                "app container should exist after initial up",
+            )
+
+            # modify the Dockerfile to produce a different image
+            with open(dockerfile, "w") as f:
+                f.write("FROM busybox:latest\nRUN touch /marker\n")
```
p12tic: FROM nopush/podman-compose-test

tests/integration/compose_up_behavior/test_compose_up_behavior.py
```diff
@@ -488,3 +488,162 @@ def test_localhost_image_built_if_does_not_exist(self) -> None:
...
+
+            # all containers should be running
+            self.assertTrue(
+                all(c["exited"] is False for c in new_containers.values()),
+                "Not all containers are running after up command",
+            )
+
+        finally:
+            # restore original Dockerfile
+            with open(dockerfile, "w") as f:
```
p12tic: I would rather not change files committed to repository, because in this case the source of truth for the file is `original_content` variable, not what's committed to repository.

Better create the file at the beginning of the test, change however you want and then remove at the end.

p12tic (changes requested):
Looks good overall, thank you very much and sorry for delayed review. I had only nitpicks regarding test images, otherwise everything is good.

thrawn-sh:
I've updated to code according to your suggestions, or at least how I understood them, if still goes in a wrong direction can you please elaborate?

p12tic (approved):

thrawn-sh:
@p12tic anything required from my side still?

PR #1485, opened 2026-06-29
pass stop_grace_period as --stop-timeout to podman create

thrawn-sh:
podman-compose passes stop_signal to podman create via --stop-signal but never passes stop_grace_period as --stop-timeout.  This means containers are created with podman's default stop timeout of 10 seconds regardless of the compose file setting.

When podman stops a container (e.g. during recreation on up), it uses the container's embedded stop timeout, causing a premature SIGKILL after 10 seconds even when stop_grace_period is set to a higher value.

Parse stop_grace_period with str_to_seconds and pass it as --stop-timeout to podman create, matching how stop_signal is already handled.


p12tic (approved):
Thanks!

PR #1488, opened 2026-07-07
add secret environment loading

### LINKED ISSUE #655: Secrets aren't getting created properly when specified in the compose file
**Summary**
`podman-compose` fails to create secrets defined in the compose file, resulting in permission errors.

**Describe the bug**
When defining secrets in a compose file for `podman-compose` for public/private ssh keys, the secrets are not created, and permission errors occur for the ssh keys in the container.

**Expected Behavior**
The secrets should be created and ssh keys should be present in the container with proper permissions.

**Actual Behavior**
The secrets are not created, and permission errors occur for the ssh keys in the container. Contents of the `.ssh` folder in the container:
```
[root@f89333bda8fa .ssh]# ls -l
ls: cannot access 'id_ed25519': Permission denied
ls: cannot access 'id_ed25519.pub': Permission denied
total 0
-????????? ? ? ? ?            ? id_ed25519
-????????? ? ? ? ?            ? id_ed25519.pub
```
Using `podman secret ls` confirms that no secrets are created.

**To Reproduce**
Run `podman-compose up` using the following `compose.yaml` file:
```
services:
  secrets-test:
    image: registry.fedoraproject.org/fedora:latest
    secrets:
      - source: ssh-key
        target: /root/.ssh/id_ed25519
        mode: 0600
      - source: ssh-key-pub
        target: /root/.ssh/id_ed25519.pub
        mode: 0644
    tty: true
    stdin_open: true
secrets:
  ssh-key:
    file: $HOME/.ssh/id_ed25519
  ssh-key-pub:
    file: $HOME/.ssh/id_ed25519.pub
```

Inspecting the `/root/.ssh` directory of the resulting container:
```
[root@f89333bda8fa .ssh]# ls -l
ls: cannot access 'id_ed25519': Permission denied
ls: cannot access 'id_ed25519.pub': Permission denied
total 0
-????????? ? ? ? ?            ? id_ed25519
-????????? ? ? ? ?            ? id_ed25519.pub
```

Output of  `podman secret ls`:
```
ID          NAME        DRIVER      CREATED     UPDATED
```

**Additional Context**
I can successfully accomplish this using the following command in `bash`:
```
podman secret create ssh-key $HOME/.ssh/id_ed25519
podman secret create ssh-key-pub $HOME/.ssh/id_ed25519.pub

podman create -t --secret ssh-key,target=/root/.ssh/id_ed25519,mode=0600 --secret ssh-key-pub,target=/root/.ssh/id_ed25519.pub,mode=0644 registry.fedoraproject.org/fedora:latest
```
Runnig the created container and inspecting the `/root/.ssh` directory confirms that the secrets are working as expected.

**Version info**
```
$ podman-compose version
using podman version: 4.4.2
podman --version 
podman version 4.4.2
```

**Environment:**
 - OS: Linux
 - podman version: 4.4.2
 - podman compose version: 1.0.3

ManInDark:
# Actions

I implemented the loading of secrets from environment variables as defined in the specification.

Every time `podman-compoese up` is executed, it creates the secret environment variable that don't already exist with the currently set one and refuses to start if it is not present.

This partially resolves #655, but not fully.

## Specification

https://github.com/compose-spec/compose-spec/blob/main/09-secrets.md

Especially the first example of `Example 2`

## Contributor Checklist:

I've read through the whole document and run all commands except `python -m unittest discover -v tests/integration  # takes a while to run`, since it is run by GitHub CI on pull request anyway.

The commands returned no errors for which I believe my code changes are responsible, though there are some others, possibly because of my incomplete setup.

Since I'm new here, please tell me if there is anything I should change.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -794,6 +794,42 @@ async def get_mount_args(
     return ["--mount", args]
 
 
+async def create_secrets_from_environment(compose: PodmanCompose) -> None:
+    if not compose.declared_secrets:
+        return
+    for secret_name in compose.declared_secrets.keys():
+        secret_environment = compose.declared_secrets[secret_name].get("environment")
+        if secret_environment:
+            secret_environment_value = os.getenv(secret_environment)
+
+            if secret_environment_value is None:
+                raise ValueError(
+                    f"Environment variable '{secret_environment}' required"
+                    + " by secret '{secret_name}' is not set in the process environment."
+                )
+
+            log.debug(
+                "attempting creation of secret '%s' set to '%s'",
+                secret_name,
+                secret_environment_value,
+            )
+
+            assert compose.project_name is not None
+
+            await compose.podman.run(
+                [],
+                "secret",
+                [
+                    "create",
+                    "--label",
+                    "io.podman.compose.project=" + compose.project_name,
+                    "--env",
+                    f"{compose.project_name}_{secret_name}",
+                    secret_environment,
+                ],
+            )
+
+
 def get_secret_args(
     compose: PodmanCompose,
     cnt: dict[str, Any],
@@ -833,13 +869,10 @@ def get_secret_args(
         if podman_is_building:
             secret_id = secret_target if secret_target else secret_name
             return ["--secret", f"id={secret_id},env={source_env}"]
-        # TODO: Runtime support for environment secrets requires injecting the
-        # value into the container (e.g. via podman cp or podman secret create),
-        # which is beyond the scope of this arg-generating function.
-        raise ValueError(
-            f'ERROR: Secret "{secret_name}" uses environment source, '
-            "which is not supported for runtime secrets in podman-compose."
-        )
+
+        assert compose.project_name is not None
+        log.debug("mounting secret '%s'", secret_name)
+        return ["--secret", f"{compose.project_name}_{secret_name}"]
 
     if source_file:
         # assemble path for source file first, because we need it for all cases
@@ -3959,6 +3992,8 @@ async def compose_up(compose: PodmanCompose, args: argparse.Namespace) -> int | 
     existing_containers = await compose.podman.existing_containers(compose.project_name)
     recreate_services: set[str] = set()
     running_services = {c.service_name for c in existing_containers.values() if not c.exited}
+
+    await create_secrets_from_environment(compose)
 
     if existing_containers:
         if args.force_recreate and args.no_recreate:
--- a/newsfragments/environment_secrets.feature
+++ b/newsfragments/environment_secrets.feature
@@ -1,0 +1,1 @@
+Added support for loading secrets from the environment
--- a/tests/integration/secrets/docker-compose.yaml
+++ b/tests/integration/secrets/docker-compose.yaml
@@ -33,6 +33,7 @@ services:
         - source: podman_compose_test_secret
           target: ENV_SECRET
           type: env
+        - environment_secret
 
 secrets:
   podman_compose_test_secret:
@@ -46,3 +47,5 @@ secrets:
     name: podman_compose_test_secret_3
   file_secret:
     file: ./my_secret
+  environment_secret:
+    environment: "TEST_ENVIRONMENT"
--- a/tests/integration/secrets/test_podman_compose_secrets.py
+++ b/tests/integration/secrets/test_podman_compose_secrets.py
@@ -47,6 +47,7 @@ class TestComposeNoSecrets(unittest.TestCase, RunSubprocessMixin):
                     "up",
                     "test",
                 ],
+                env={"TEST_ENVIRONMENT": "important-secret-is-important"},
             )
 
             self.assertNotIn(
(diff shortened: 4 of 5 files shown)


ManInDark:
I suspect based on the fact that I cannot find any pattern in the failed to not-failed tests and that they fail because a container was not found that the reason for them failing is just some timing or something, not with the code I changed.

I'd appreciate if someone could manually trigger them again to see if they are then "fixed".

ManInDark:
I've looked at the logs a bit more and it seems like every podman version before 5.4.2 fails for some reason, I'll investigate that.

ManInDark:
I believe I've found the issue. Tested with podman `4.3.1`, as packaged in Debian Bookworm.

```bash
$ podman secret create --env euid SHELL
Error: cannot create store secret data: environment variable SHELL is not set
$ echo $SHELL
/bin/bash
```

The variable is clearly set, but podman does not recognize that. How am I supposed to handle this case? I see two possible options:
1. Check for installed Podman Version and handle them differently
2. Rewrite the secret creation in a way that both versions are supported

Which option is the preferred in this project?

podman_compose.py
```diff
@@ -833,13 +870,10 @@ def get_secret_args(
...
             return ["--secret", f"id={secret_id},env={source_env}"]
-        # TODO: Runtime support for environment secrets requires injecting the
-        # value into the container (e.g. via podman cp or podman secret create),
-        # which is beyond the scope of this arg-generating function.
-        raise ValueError(
-            f'ERROR: Secret "{secret_name}" uses environment source, '
-            "which is not supported for runtime secrets in podman-compose."
-        )
+
+        assert compose.project_name is not None
```
p12tic, line 873: Check podman version and throw error if it's too old to support environment secrets.

podman_compose.py
```diff
@@ -794,6 +794,43 @@ async def get_mount_args(
...
+        if secret_environment:
+            secret_environment_value = os.getenv(secret_environment)
+
+            if secret_environment_value is None:
+                raise ValueError(
+                    f"Environment variable '{secret_environment}' required"
+                    + " by secret '{secret_name}' is not set in the process environment."
+                )
+
+            log.debug(
```
p12tic, line 811: Check podman version and throw error if podman is too old.

p12tic:
@ManInDark 

Thanks for working on this.

> Check for installed Podman Version and handle them differently

It's fine if we don't support older podman to implement features that we want. So in this case, please figure out the oldest podman which still supports the feature and add an error if the feature is used but podman is too old.

In tests, the same, skip test if podman is too old. In this particular case I suppose you'll need a separate docker-compose file so that other test cases aren't skipped unnecessarily.

p12tic (changes requested):
Looks good, but needs to not break older podman.

ManInDark:
Thanks for the feedback!

I've now started building podman according to [this guide](https://podman.io/docs/installation#building-from-source) and discovered something strange:

```bash
# podman -v
podman version 6.1.0-dev
# echo $SHELL
/bin/bash
# podman secret create --env --replace test SHELL
Error: cannot create store secret data: environment variable SHELL is not set
```
Interestingly:
```bash
# podman secret create --env --replace test PATH
ad1dc38a2513148925c1de1b6
```

This confuses me a bit, but I belive the issue lies somewhere in `bash` and how it handels environment variables, since `SHELL` also doesn't appear in the `/usr/bin/env` dump, but `PATH` does.

I'll go investigate that a bit more to see if I can figure out what is going on.

ManInDark:
Ok so this "rabbit hole" I've gone down isn't actually that deep, it might as well just be a small ditch.

The result: bash just simply doesn't always export the environment variables, so my manual tests all failed since the `SHELL` environment variable was - as mentioned by `podman` - not actually set.

So the problem is somewhere else.

A rerun of the test I've done above, this time with an environment variable that is set:
```bash
# podman -v
podman version 4.3.1
# podman secret create --env test PWD
95a02cd4c64976bf3e7269b39
# podman secret ls
ID                         NAME        DRIVER      CREATED        UPDATED
95a02cd4c64976bf3e7269b39  test        file        2 seconds ago  2 seconds ago
```

Which makes the whole thing even more confusing, since this would equal pretty much the setup of `bookworm, 3.11, 4.3.1, rootless` (which doesn't exist apparently?)

ManInDark:
Well sometimes it might help to read the log more thoroughly:

> Error: unknown flag: --replace

ManInDark:
The option `--replace` exists since [v4.7.0](https://github.com/podman-container-tools/podman/releases/tag/v4.7.0) of the podman cli.

ManInDark:
I've now simply switched to only creating the secret, not replacing it, since it works that way for all versions.

p12tic (approved):
Great, thank you!

ManInDark:
Thank you for the support too!

Should I look into implementing the deletion of the secrets on `podman-compose down`? I'm kinda torn on whether that is something I'd expect or dislike, though I am leaning towards the latter, since preserving them is IMO the thing I'd like more.

@p12tic (not sure if GH notifies people if the PR is merged)

PR #1491, opened 2026-07-08
Support up --no-attach

NVolcz:
Add Docker Compose-compatible parsing and handling for repeated up --no-attach SERVICE options.

Services selected with --no-attach are still started and managed by the attached up lifecycle, but their output is suppressed.

### DIFF
--- a/tests/integration/logs/test_podman_compose_logs.py
+++ b/tests/integration/logs/test_podman_compose_logs.py
@@ -136,3 +136,38 @@ class TestLogs(unittest.TestCase, RunSubprocessMixin):
                 compose_path,
                 "down",
             ])
+
+    def test_up_no_attach_suppresses_selected_service_output(self) -> None:
+        compose_path = os.path.join(test_path(), "logs/docker-compose.yml")
+
+        try:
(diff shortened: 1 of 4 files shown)


tests/integration/logs/test_podman_compose_logs.py
```diff
@@ -136,3 +136,28 @@ def test_logs(
...
+            out, _ = self.run_subprocess_assert_returncode([
+                podman_compose_path(),
+                "-f",
+                compose_path,
+                "up",
+                "--no-attach",
+                "test2",
+            ])
+
+            self.assertIn(b"test1", out)
```
p12tic: Please compare whole output - decode, split lines, maybe sort and compare to a list.

p12tic (changes requested):
LGTM, just small nit regarding test.

tests/integration/logs/test_podman_compose_logs.py
NVolcz: Something like this?

p12tic (approved):

PR #1502, opened 2026-07-15
Support COMPOSE_ENV_FILES

### LINKED ISSUE #1501: `COMPOSE_ENV_FILES`  environment variable is ignored
**Describe the bug**
podman-compose ignores the `COMPOSE_ENV_FILES` environment variable. When set, the specified env files should be loaded in order, but currently they are not parsed at all.

**To Reproduce**

```
#.env.default
SUBJECT=World
GREETING=Hello
```
```
#.env.override
SUBJECT=Podman
```
```
# docker-compose.yml
services:
  dummy:
    image: busybox
    command: ["sh", "-c", "echo $GREETING $SUBJECT"]
```
Run the command:
```COMPOSE_ENV_FILES=.env.default,.env.override podman-compose up```

**Expected behavior**
The container should print:
```Hello Podman```

**Actual behavior**
The environment variables are not loaded, so the output is empty.

**Environment:**
OS: Debian GNU/Linux 13 (trixie)
Python 3.13.5
podman-compose version: 1.6.0
podman version: 5.4.2

**Additional context**
docker-compose correctly handles the same command and produces:
```Hello Podman```

The following command works as expected, confirming the env files themselves are valid:
```podman-compose --env-file .env.default --env-file .env.override up```
This should be equivalent to the COMPOSE_ENV_FILES approach above.

This issue was partially discussed in #813. That report contained two separate problems; one was resolved, but support for `COMPOSE_ENV_FILES` remains broken.

iamdhrv:
Closes #1501

Load the comma-separated files from `COMPOSE_ENV_FILES` when the CLI does not provide `--env-file`. Explicit `--env-file` arguments retain higher precedence, and the existing project `.env` fallback remains unchanged when neither source is set.

This matches Docker Compose's documented behavior: https://docs.docker.com/compose/how-tos/environment-variables/envvars/#compose_env_files

## Verification

- `python -m unittest discover tests/unit` (448 tests)
- `ruff format --check`
- `ruff check`
- `mypy .`
- `pylint podman_compose.py`

AI disclosure: Codex assisted with issue research, implementation, and validation. I reviewed the Compose precedence requirements, final diff, and test results before submission.

## Contributor Checklist

- [x] Read and followed `CONTRIBUTING.md`
- [x] Added focused unit tests
- [x] Added a `newsfragments` bug-fix entry
- [x] Signed off the commit

### DIFF
--- a/newsfragments/compose_env_files.bugfix
+++ b/newsfragments/compose_env_files.bugfix
@@ -1,0 +1,2 @@
+Load the comma-separated environment files from ``COMPOSE_ENV_FILES`` when no
+``--env-file`` option is provided.
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -3032,6 +3032,10 @@ class PodmanCompose:
                 cmd_parser(subparser)
         self.global_args = parser.parse_args(argv)
 
+        compose_env_files = os.environ.get("COMPOSE_ENV_FILES")
+        if not self.global_args.env_file and compose_env_files:
+            self.global_args.env_file = compose_env_files.split(",")
+
         if self.global_args.version:
             self.global_args.command = "version"
         if not self.global_args.command or self.global_args.command == "help":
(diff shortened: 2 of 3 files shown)


p12tic (approved):
Thanks a lot!

Issue #939, opened 2024-05-13
Required variable syntax ("${BAR:?BAR variable missing}") not resolved when not in quotation marks since v1.1.0 (in command)


rugk:
**Describe the bug**
The required variable syntax (`${BAR:?BAR variable missing}`) is not resolved when not in quotation marks since v1.1.0.

**To Reproduce**
Here is an full example/MWE:
https://gist.github.com/rugk/ac09b28bc9de833e772cb1509364b421 (You can just git clone the gist.)

The relevant part is:
```yaml
version: "3"

services:
  example:
    image: busybox
    env_file: .foo
    # here, variable replacing does not work
    command: /app/script.sh ${BAR:?BAR variable missing}
    # This works:
    # command: /app/script.sh "${BAR:?BAR variable missing}"
    # this fails with "FOO variable missing":
    # command: /app/script.sh "${FOO:?FOO variable missing}"
    volumes:
      # :z just needed because of SELinux
      - .:/app:z
```

`script.sh`:
```sh
#!/bin/sh
# chmod +x this script, if needed

set | grep -E 'FOO|BAR'
echo FOO=$FOO and BAR=$BAR
echo CMDARGS: "$@"
```

(copied/adjusted from https://github.com/containers/podman-compose/issues/848, which may or may not be related)

`BAR` can be there or not, in any case it is not handled at all.

**Expected behavior**
AFAIK this has worked in versions before:
* when `BAR` is not set, show an error `BAR variable missing`
* when `BAR` is set, expand it with it's value

AFAIK [this is the correct syntax/usage for required variables](https://docs.docker.com/compose/environment-variables/env-file/#interpolation).

**Actual behavior**
It is not expanded, and just passed/treated as is.

In `podman-compose-test`:

```
$ podman-compose down&&podman-compose up
podman-compose-test_example_1
podman-compose-test_example_1
a7974b799ede9b6f4ae6f3ef28163a23bd3a4ecd72b0ca74088690d95f857c85
22728b5a7933c90293f5260549317ef083be0ef346e557f680b6338bbbe27c1f
e77f8f7dbefe1a8c78fc63f929f3dc8cb1adc201b247b029ccd0e26f641a037f
[example] | FOO='test'
[example] | FOO=test and BAR=
[example] | CMDARGS: ${BAR:?BAR variable missing}
```

**Output**

```
$ podman-compose version
podman-compose version 1.1.0
podman version 5.0.2

$ podman --version
podman version 5.0.2
```

[...]

PR #1504, opened 2026-07-17
Fix variable interpolation in unquoted values (command and entrypoint)

mokibit:
Fix variable interpolation in unquoted ``command`` and ``entrypoint`` values. Previously, ``shlex.split()`` was applied before interpolation, which broke expressions such as ``${VAR:?error message}`` because spaces inside the variable operand were split into separate tokens. String values are now kept intact during parsing and only split after interpolation, matching docker-compose behavior.

Relevant docker-compose documentation: https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation

Fixes issue https://github.com/containers/podman-compose/issues/939.


tests/integration/interpolation/docker-compose-command-interpolation.yml
```diff
@@ -0,0 +1,5 @@
+version: "3.7"
+services:
+    command_test:
+        image: busybox
```
p12tic: Please use local test image.

p12tic (changes requested):

PR #1506, opened 2026-07-18
Support COMPOSE_FILE in `.env` for compose file discovery

### LINKED ISSUE #475: some .env values are ignored
**Describe the bug**
COMPOSE_FILE value from .env is always overwritten with the default value.

**To Reproduce**
```
$ cat .env 
COMPOSE_FILE=docker-compose.yaml
COMPOSE_PROJECT_NAME=blahblah
$ cat docker-compose.yaml 
version: "2.4"
services:
  srv:
    image: alpine
sagemcom@prod-test:~/test$ cat docker-compose.override.yaml 
services:
  srv2:
    image: alpine

$ ~/podman-compose/podman_compose.py ps
podman-compose version: 1.0.4
['podman', '--version', '']
using podman version: 3.0.1
 ** merged:
 {
  "version": "2.4",
  "services": {
    "srv": {
      "image": "alpine"
    },
    "srv2": {
      "image": "alpine"
    }
  },
  "_dirname": "/home/sagemcom/test"
}
podman ps -a --filter label=io.podman.compose.project=test
CONTAINER ID  IMAGE   COMMAND  CREATED  STATUS  PORTS   NAMES
exit code: 0
$ export COMPOSE_FILE=docker-compose.yaml # defined the same value, 
$ ~/podman-compose/podman_compose.py ps
podman-compose version: 1.0.4
['podman', '--version', '']
using podman version: 3.0.1
podman ps -a --filter label=io.podman.compose.project=test
CONTAINER ID  IMAGE   COMMAND  CREATED  STATUS  PORTS   NAMES
exit code: 0
```
**Expected behavior**
.env values should be used, if neither command line arg, nor environment variable of the same name is present

**Actual behavior**
some variables are always overwritten (compared to the value in .env file), even if documented logic says those shouldn't 

**Output**

```
$ podman-compose --version
['podman', '--version', '']
using podman version: 3.0.1
podman --version 
podman version 3.0.1
exit code: 0
```

**Environment:**
 - OS: Linux / WSL / Mac
 - podman version: 3.0.1
 - podman compose version: (1.0.3)

**Additional context**

The problem is the applied logic in _parse_compose_file(self):
It should be something like this:
- .env file should be used as a set of default values for all variables in the file
- then comes the os.environ, which possibly overrides the defaults, if the same var is present
- then comes the command line argument, if provided, which has the highest precedence

mokibit:
Previously, podman-compose only checked the shell environment for `COMPOSE_FILE` when resolving the default compose file. It did not consider `.env` or `--env-file` during this phase, so project-level defaults were silently ignored.

This changes loads env file(s) early (before compose file discovery) so that `COMPOSE_FILE` defined in `.env` or via `--env-file` is honored when no `-f` flag is provided.

Relevant Docker Compose documentation:
https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/

Fixes https://github.com/containers/podman-compose/issues/475.

### DIFF
--- a/tests/integration/compose_file_from_env/test_podman_compose_compose_file_from_env.py
+++ b/tests/integration/compose_file_from_env/test_podman_compose_compose_file_from_env.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: GPL-2.0
+
+import unittest
+from pathlib import Path
+
+from tests.integration.test_utils import RunSubprocessMixin
+from tests.integration.test_utils import podman_compose_path
+
+
+def compose_base_path() -> Path:
+    return Path(__file__).parent
+
+
+class TestComposeFileFromEnv(unittest.TestCase, RunSubprocessMixin):
+    """Test that COMPOSE_FILE from .env is honored when no -f is given."""
+
+    def test_compose_file_from_dotenv(self) -> None:
+        base_path = compose_base_path()
+        out, _ = self.run_subprocess_assert_returncode(
+            [podman_compose_path(), "config"],
+            0,
+            cwd=base_path,
+        )
+        self.assertEqual(
+            out.decode("utf-8"),
+            'services:\n'
+            '  dotenv-service:\n'
+            '    command: echo "from-dotenv"\n'
+            '    image: nopush/podman-compose-test\n'
+            '\n',
+        )
+
+    def test_compose_file_from_explicit_env_file(self) -> None:
+        base_path = compose_base_path()
+
+        out, _ = self.run_subprocess_assert_returncode(
+            [
+                podman_compose_path(),
+                "--env-file",
+                str(base_path / "custom.env"),
+                "config",
+            ],
+            0,
+            cwd=base_path,
+        )
+        self.assertEqual(
+            out.decode("utf-8"),
+            'services:\n'
+            '  explicit-env-service:\n'
+            '    command: echo "from-explicit-env"\n'
+            '    image: nopush/podman-compose-test\n'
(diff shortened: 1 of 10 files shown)


tests/integration/compose_file_from_env/test_podman_compose_compose_file_from_env.py
```diff
@@ -0,0 +1,121 @@
...
+            "    - from-dotenv\n"
+            "    image: nopush/podman-compose-test\n"
+            "\n"
+        )
+        self.assertEqual(out.decode("utf-8"), expected)
+
+    def test_compose_file_from_explicit_env_file(self) -> None:
+        base_path = compose_base_path()
+
+        out, err, returncode = self.run_subprocess(
```
p12tic: run_process_assert_returncode

tests/integration/compose_file_from_env/test_podman_compose_compose_file_from_env.py
```diff
@@ -0,0 +1,121 @@
...
+                podman_compose_path(),
+                "--env-file",
+                str(base_path / "custom.env"),
+                "config",
+            ],
+            cwd=base_path,
+        )
+        self.assertEqual(returncode, 0, f"Stderr: {err.decode()}")
+
+        expected = (
```
p12tic: inline expected

p12tic (changes requested):
Looks good, several nits related to tests.

p12tic (approved):

PR #1513, opened 2026-07-22
Reject dict format for service and build secrets

### LINKED ISSUE #1365: using invalid format for service secrets misbehaves instead of failing
**Describe the bug**
using invalid format for service secrets doesn't fail, but misbehaves instead

**Steps to reproduce the behavior**

1. create a secret `echo "aaa" | podman secret create test-secret -`
2. create `docker-compose.yml`
```
services:
  tester:
    image: debian:trixie-slim
    command: "tail -f /dev/null"
    container_name: tester
    init: true
    secrets:
      test-secret: # this is not valid according to [docker compose spec](https://docs.docker.com/reference/compose-file/services/#secrets)
        source: test-secret
        type: env
        target: MY_VAR

secrets:
  test-secret:
    external: true
```
3. run `docker compose up -d`

**Expected behavior**
the command fails because `services.tester.secrets` in the docker compose file is invalid

**Actual behavior**
command runs, and the secret `test-secret` is mounted to `/run/secrets/test-secret`, but not as env variable

**Output**

```
$ podman-compose version
podman-compose version 1.5.0
podman version 5.7.0
```
 
**Additional context**

Reason for this is likely that dicts in python are enumerable and [this line here
](https://github.com/containers/podman-compose/blob/492247172aa53fd3f1f39a22c98e77dd86ce76ed/podman_compose.py#L1207) enumerates keys of `secrets` dict, which is incorrect behaviour here.

mokibit:
Reject invalid dict format for service and build secrets instead of silently misbehaving. Previously, specifying `secrets` as a mapping (dict) caused podman-compose to iterate over keys and mount secrets incorrectly. Now it raises a clear error message.

Fixes https://github.com/containers/podman-compose/issues/1365.

### DIFF
--- a/tests/unit/test_normalize_service.py
+++ b/tests/unit/test_normalize_service.py
@@ -5,6 +5,7 @@ from typing import Union
 
 from parameterized import parameterized
 
+from podman_compose import PodmanComposeError
 from podman_compose import normalize_service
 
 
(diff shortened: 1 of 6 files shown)


tests/unit/test_normalize_service.py
```diff
@@ -117,3 +117,33 @@ def test_command_like(
...
+            {"secrets": [{"source": "my_secret", "target": "ENV_VAR"}]},
+        ),
+    ])
+    def test_secrets_normalization(
+        self, test_name: str, input_service: dict[str, Any], expected_service: dict[str, Any]
+    ) -> None:
+        self.assertEqual(normalize_service(input_service), expected_service)
+
+    def test_secrets_dict_raises(self) -> None:
+        from podman_compose import PodmanComposeError
```
p12tic: Imports should be at the top of the file.

p12tic (changes requested):

p12tic (approved):

PR #1518, opened 2026-07-24
Preserve POSIX absolute bind mount paths on Windows

### LINKED ISSUE #1352: Unable to mount paths inside WSL podman machine when running on Windows
**Describe the bug**
The `podman-compose` is unable to mount paths correctly inside the WSL podman machine when running on a Windows OS. In my scenario, I need to mount MariaDB Unix socket from a `mariadb` container on a WSL podman machine so that I can share Unix socket with the `php` container. As we know Unix sockets are not supported by Windows, so mounting them on Windows file system does not make sense.

Podman machine works in `rootless` mode.

Please make sure it's not a bug in podman (in that case report it to podman)
or your understanding of docker-compose or how rootless containers work (for example, it's normal for rootless container not to be able to listen for port less than 1024 like 80)

I am sure this is `podman-compose` bug, because `podman` mounts paths correctly. When we create the `mysqld` directory in the WSL podman machine and run:
```
podman run --name mariadb-test -v /home/user/mysqld:/var/run/mysqld --env MARIADB_USER=user --env MARIADB_PASSWORD=pass --env MARIADB_ROOT_PASSWORD=root_pass --env MARIADB_DATABASE=db mariadb:latest
```

After ssh into WSL podman machine we can see:
```
podman machine ssh
Connecting to vm podman-machine-default. To close connection, use `~.` or `exit`
Last login: Sat Dec 13 19:10:09 2025 from ::1
[user@DESKTOP-USIP4GD ~]$ ls -la mysqld/
total 12
drwxr-xr-x 2 525286 525286 4096 Dec 15 12:24 .
drwx------ 7 user   user   4096 Dec 15 12:24 ..
-rw-rw---- 1 525286 525286    2 Dec 15 12:24 mysqld.pid
srwxrwxrwx 1 525286 525286    0 Dec 15 12:24 mysqld.sock
```

**To Reproduce**
1. Lat's analyze the following `docker-compose.yml` file:
``` yml
version: '3.8'
services:
    nginx:
        image: nginx:latest
        container_name: nginx
        ports:
            - "8080:80"
        networks:
            - internal
        restart: unless-stopped
    php:
        image: php:8.5-fpm
        container_name: php
        networks:
            - internal
        restart: unless-stopped
    mariadb:
        image: mariadb:latest
        container_name: mariadb
        environment:
            MARIADB_USER: 'user'
            MARIADB_PASSWORD: 'pass'
            MARIADB_ROOT_PASSWORD: 'root_pass'
            MARIADB_DATABASE: 'db'
        ports:
            - 3306:3306
        networks:
            - internal
        volumes:
            - type: bind
              source: /home/user/mysqld
              target: /var/run/mysqld
        restart: unless-stopped

networks:
    internal:
        driver: bridge
```
2. Let's take a look at the definition of bind mount:
``` yml
    volumes:
        - type: bind
          # This path should be mounted inside the WSL podman machine, which is the Fedora distribution.
          source: /home/user/mysqld
          # This is the path to the directory where the mariadb container stores its unix socket.
          target: /var/run/mysqld
```

3. Inside directory with `docker-compose.yml` file run:
```
podman compose --verbose --dry-run up mariadb
```

4. We should see the following output:
```
>>>> Executing external compose provider "C:\\Users\\mck\\AppData\\Local\\Programs\\Python\\Python314\\Scripts\\podman-compose.exe". Please see podman-compose(1) for how to disable this message. <<<<

DEBUG:podman_compose:** excluding: {'php', 'nginx'}
INFO:podman_compose:building images: ...
INFO:podman_compose:['podman', 'ps', '--filter', 'label=io.podman.compose.project=podman-compose-test', '-a', '--format', 'json']
INFO:podman_compose:podman pod exists pod_podman-compose-test
INFO:podman_compose:podman pod create --name=pod_podman-compose-test --infra=false --share=
INFO:podman_compose:creating missing containers: ...
DEBUG:podman_compose:** skipping create: nginx
DEBUG:podman_compose:** skipping create: php
INFO:podman_compose:['podman', 'network', 'exists', 'podman-compose-test_internal']
INFO:podman_compose:podman create --name=mariadb --pod=pod_podman-compose-test --label io.podman.compose.config-hash=53998c54803774e43b5d17b3a93a3cb81575d9f206752276658e44ddb8f6c8c5 --label io.podman.compose.project=podman-compose-test --label io.podman.compose.version=1.5.0 --label PODMAN_SYSTEMD_UNIT=podman-compose@podman-compose-test.service --label com.docker.compose.project=podman-compose-test --label com.docker.compose.project.working_dir=D:\Web\podman-compose-test --label com.docker.compose.project.config_files=docker-compose.yml --label com.docker.compose.container-number=1 --label io.podman.compose.service=mariadb --label com.docker.compose.service=mariadb -e MARIADB_USER=user -e MARIADB_PASSWORD=user_pass -e MARIADB_ROOT_PASSWORD=root_pass -e MARIADB_DATABASE=db -v D:\home\user\mysqld:/var/run/mysqld --network=podman-compose-test_internal:alias=mariadb -p 3306:3306 --restart unless-stopped mariadb:latest
```

In the output above there is a snippet showing the processed mount path:
```
... -e MARIADB_DATABASE=db -v D:\home\user\mysqld:/var/run/mysqld --network=podman-compose-test_internal:alias=mariadb ...
```

As we can see the mount path is created inside the Windows filesystem in the path:
```
D:\home\user\mysqld
```

**Expected behavior**
The volume should be mounted inside the WSL podman machine in the path:
```
/home/user/mysqld
```

**Actual behavior**
The volume is mounted inside the Windows filesystem in the path:
```
D:\home\user\mysqld
```

**Output**

```
podman compose --version
>>>> Executing external compose provider "C:\\Users\\mck\\AppData\\Local\\Programs\\Python\\Python314\\Scripts\\podman-compose.exe". Please see podman-compose(1) for how to disable this message. <<<<

podman-compose version 1.5.0
podman version 5.7.1
```

**Environment:**
 - OS: Windows 11 / WSL
 - podman version: 5.7.1

**Additional context**
The `podman` was installed directly on Windows 11 using the standard installer. The `podman-compose` was installed as follows:
```
pip3 install podman-compose
```

I managed to find a workaround for this problem by defining the volume source path using the Windows npipe syntax as follows:
``` yml
    volumes:
        - type: bind
          # This Windows npipe is mounted inside the WSL podman machine in the path /mnt/wsl/pipe/mysqld (need to exist)
          source: \\.\pipe\mysqld
          target: /var/run/mysqld
```

To me this only seems like a temporary solution. This is not documented anywhere.

sgolev:
On Windows, `parse_short_mount()` and `assert_volume()` unconditionally run `os.path.abspath()` on bind mount sources. For a POSIX absolute source like `/etc/example`, `os.path.abspath()` prepends the current Windows drive, producing `C:\etc\example`. As a result, when targeting a remote Linux host, podman-compose fails with:

    statfs /mnt/c/etc/example: no such file or directory

The fix is to skip the rewrite when the source is POSIX absolute path on Windows. Relative, `~`, `.sock`, and drive paths (`C:\...`) are unaffected.

This PR fixes https://github.com/containers/podman-compose/issues/1352.

As a side note, `assert_volume()` doesn't work as intended when targeting a remote machine, because it ensures that directories exist on  the local machine. I've tried to introduce minimal changes just to fix the Windows-to-remote-Linux workflow. When a POSIX absolute path is detected on Windows, both the rewrite and the local probe are skipped.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -18,7 +18,9 @@ import hashlib
 import inspect
 import json
 import logging
+import ntpath
 import os
+import posixpath
 import random
 import re
 import shlex
@@ -44,13 +46,6 @@ from urllib.parse import quote
 import yaml
 from dotenv import dotenv_values
 
-# Python loads the appropriate path module based on the OS, but we need to be able
-# to check if a path is absolute according to BOTH major OS's rules.
-if os.name == 'posix':
-    from ntpath import isabs as secondarypathisabs
-if os.name == 'nt':
-    from posixpath import isabs as secondarypathisabs
-
 __version__ = "1.6.0"
 
 script = os.path.realpath(sys.argv[0])
@@ -64,6 +59,10 @@ def is_list(list_object: Any) -> bool:
         and not isinstance(list_object, dict)
         and hasattr(list_object, "__iter__")
     )
+
+
+def is_abs_any_os(path: str) -> bool:
+    return posixpath.isabs(path) or ntpath.isabs(path)
 
 
 def is_relative_ref(path: str) -> bool:
@@ -186,7 +185,7 @@ def parse_short_mount(mount_str: str, basedir: str) -> dict[str, Any]:
         # User-relative path
         # - ~/configs:/etc/configs/:ro
         mount_type = "bind"
-        if os.name != 'nt' or (os.name == 'nt' and ".sock" not in mount_src):
+        if not is_abs_any_os(mount_src) and (os.name != 'nt' or ".sock" not in mount_src):
             mount_src = os.path.abspath(os.path.join(basedir, os.path.expanduser(mount_src)))
     else:
         # Named volume
@@ -581,6 +580,10 @@ async def assert_volume(compose: PodmanCompose, mount_dict: dict[str, Any]) -> N
     """
     vol = mount_dict.get("_vol")
     if mount_dict["type"] == "bind":
+        # Don't rewrite POSIX absolute paths on Windows: they're host paths for
+        # a remote Linux connection.
+        if os.name == 'nt' and posixpath.isabs(mount_dict["source"]):
+            return
         basedir = os.path.realpath(compose.dirname)
         mount_src = mount_dict["source"]
         mount_src = os.path.abspath(os.path.join(basedir, os.path.expanduser(mount_src)))
@@ -3541,10 +3544,7 @@ def is_context_git_url(path: str) -> bool:
     # URL contains a ":" character, a hint of a valid URL
     # But also detects windows file paths (e.g. "C:\path\to\contextdir") as urls
     is_path_with_drive_letter = (
-        (os.path.isabs(path) or secondarypathisabs(path))
-        and len(path) > 2
-        and path[1] == ':'
-        and path[2] in ('\\', '/')
+        is_abs_any_os(path) and len(path) > 2 and path[1] == ':' and path[2] in ('\\', '/')
     )
     if r.scheme != "" and r.netloc == "" and r.path != "" and not is_path_with_drive_letter:
         return True


p12tic:
Shouldn't the proper fix be to switch to posix path module for any operations on a remote Linux machine?

sgolev:
If I understand you correctly, you're suggesting that we check whether we're running in remote mode and, in that case, switch to `posixpath`. But  it helps only with absolute paths, so the behavior is essentially the same.

BTW, I found how Docker Compose handles paths: https://github.com/compose-spec/compose-go/blob/ace34ebd2aae2913524151b2b02e67790c88983c/paths/unix.go#L38. So they check whether the path is absolute in either POSIX or Windows notation leaving it unchanged if it is.

The corresponding Python code would be:

```py
is_abs = os.path.isabs(mount_src) or secondarypathisabs(mount_src)
if not is_abs and (os.name != 'nt' or ".sock" not in mount_src):
```

p12tic:
@sgolev

> The corresponding Python code would be: <...>

Makes sense. I think we could do exactly as this, since this makes maximum compatibility with docker compose. Probably makes sense to introduce it as a function.

p12tic (approved):

PR #1520, opened 2026-07-26
Split string command and entrypoint after interpolation

### LINKED ISSUE #1514: Regression of `config` command arguments to string (should be array)
**Describe the bug**
PR #1504, which fixed variable interpolation in unquoted values (command and entrypoint), introduced a regression in how `podman-compose config `handles the type of `command` arguments.

**To Reproduce**
`docker-compose.yml`:
```
services:
  test:
    image: busybox
    command: sleep infinity
```
Run command `podman-compose config`

**Expected behavior**
```
services:
  test:
    command:
    - sleep
    - infinity
    image: busybox
```

**Actual behavior**
```
services:
  test:
    command: sleep infinity
    image: busybox
```

**Environment:**
Debian GNU/Linux 13 (trixie)
Python 3.13.5
podman-compose version 1.6.0
podman version 5.4.2

**Additional context**
`podman-compose` should keep `command` arguments as an array and behave the same as `docker-compose` does:
```
$ docker-compose config
name: test-podman-compose
services:
  test:
    command:
      - sleep
      - infinity
    image: busybox
    networks:
      default: null
networks:
  default:
    name: test-podman-compose_default
```

Sanjays2402:
Closes #1514

## Contributor Checklist:

these guidelines.

---

Bug fix, not a new feature, so no compose spec link applies.

#1504 removed the `shlex.split()` of string `command`/`entrypoint` from `normalize_service` (it ran before interpolation and broke `${VAR:?error message}`), but nothing re-applied the split afterwards, so `podman-compose config` regressed to emitting `command: sleep infinity` instead of the array docker-compose produces. `normalize_service_final` runs after `rec_subs`, so the split now happens there.

Unit tests added in `tests/unit/test_normalize_final_build.py` (fail without the fix, pass with it). I did not add a newsfragment - happy to add one if you would like it for this regression fix.

### DIFF
--- a/tests/unit/test_normalize_final_build.py
+++ b/tests/unit/test_normalize_final_build.py
@@ -104,6 +104,22 @@ class TestNormalizeFinalBuild(unittest.TestCase):
         project_dir = cwd
         self.assertEqual(normalize_service_final(input, project_dir), expected)
 
+    cases_command_normalization = [
+        ({"image": "busybox", "command": "sleep infinity"}, ["sleep", "infinity"], None),
+        ({"image": "busybox", "command": ["sleep", "infinity"]}, ["sleep", "infinity"], None),
+        ({"image": "busybox", "entrypoint": "/bin/sh -c"}, None, ["/bin/sh", "-c"]),
+    ]
+
+    @parameterized.expand(cases_command_normalization)
+    def test_normalize_service_final_splits_string_command(
+        self, input, expected_command, expected_entrypoint
+    ):
+        # Tests that string [service.command] and [service.entrypoint] are split into
+        # lists after variable interpolation
+        actual = normalize_service_final(input, cwd)
+        self.assertEqual(actual.get("command"), expected_command)
+        self.assertEqual(actual.get("entrypoint"), expected_entrypoint)
+
     @parameterized.expand(cases_simple_normalization)
     def test_normalize_returns_absolute_path_in_context(self, input, expected):
         project_dir = cwd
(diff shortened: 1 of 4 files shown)


tests/unit/test_normalize_final_build.py
```diff
@@ -104,6 +104,22 @@ def test_normalize_service_final_returns_absolute_path_in_context(self, input, e
         project_dir = cwd
         self.assertEqual(normalize_service_final(input, project_dir), expected)
 
+    cases_command_normalization = [
+        ({"image": "busybox", "command": "sleep infinity"}, ["sleep", "infinity"], None),
+        ({"image": "busybox", "command": ["sleep", "infinity"]}, ["sleep", "infinity"], None),
+        ({"image": "busybox", "entrypoint": "/bin/sh -c"}, None, ["/bin/sh", "-c"]),
+    ]
+
+    @parameterized.expand(cases_command_normalization)
```
p12tic: Just include the list inline as expand() parameter

p12tic (changes requested):
Looks great, thanks. Just one nit.

tests/unit/test_normalize_final_build.py
Sanjays2402: Done in 1ea4d1d — the list is inline in `expand()` now. 44 tests in that file pass, ruff format/check clean.

p12tic:
@Sanjays2402 All tests are failing for some reason

Sanjays2402:
The failures were four stale config-output expectations, not failures in the command splitting itself. `normalize_service_final()` now deliberately emits `command` as an argument list, so the `config` output is:

    command:
    - echo
    - from-dotenv

I updated all four `compose_file_from_env` expectations, merged current `main`, and added the required bugfix news fragment in d6141ff. The focused normalization suite is green (44 tests). I could not run the integration suite locally because Podman is not installed; CI will exercise those tests on this head.

p12tic (approved):

PR #1522, opened 2026-07-27
Show `networks` top level element in command `config`

### LINKED ISSUE #623: podman-compose config should append network name
See differnet between `podman-compose` and `docker-compose` using reproduce case in  https://github.com/containers/podman-compose/issues/620

input
```
```

docker-compose config output
```
networks:
  default:
    name: devcontainer_default
```

podman-compose config output
```
```

https://github.com/containers/podman-compose/issues/621

Implement in python.

**Additional context**

mokibit:
Previously `podman-compose config` omitted the `networks` section and did not reflect runtime network resolution such as default network injection. Now it outputs the effective network configuration, including created and external networks with correct names, matching actual compose behavior.

Fixes https://github.com/containers/podman-compose/issues/623.

### DIFF
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2426,7 +2426,6 @@ class PodmanCompose:
         self.all_services: set[Any] = set()
         self.prefer_volume_over_mount = True
         self.x_podman: dict[PodmanCompose.XPodmanSettingKey, Any] = {}
-        self.merged_yaml: Any
         self.yaml_hash = ""
         self.console_colors = [
             "\x1b[1;32m",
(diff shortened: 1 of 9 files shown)


podman_compose.py
```diff
@@ -4765,6 +4763,59 @@ async def compose_logs(compose: PodmanCompose, args: argparse.Namespace) -> None
     await asyncio.gather(*tasks)
 
 
+def build_compose_config(compose: PodmanCompose) -> dict[str, Any]:
+    config: dict[str, Any] = {
```
p12tic: This approach will lose all custom keys that it does not know about.

PR #1525, opened 2026-07-28
Add x-podman.passwd extension

### LINKED ISSUE #1448: Ability to pass --passwd flag to podman run
I recently encountered [this](https://github.com/apache/airflow/issues/62300) issue when trying to follow Apache Airflow [tutorial](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/pipeline.html) using podman and podman-compose. Someone in the comments to the issue mentioned that `--passwd=false` works as a workaround. But it's an argument to `podman run`, but I'm trying to start the compose file from the tutorial, not an individual container, so I need a way to pass this argument to the underlying `podman run` call.

The `x-podman.passwd` option in the `docker-compose.yml` would work.

An ability to pass arbitrary podman arguments would be cool.

**Additional context**
...

gerryd:
This allows controlling the --passwd flag for podman run. This flag defaults to true, so keep the default unless set explicitly to false.

Fixes #1448

### DIFF
--- a/docs/Extensions.md
+++ b/docs/Extensions.md
@@ -16,6 +16,8 @@ container is assumed to be managed externally.
 
 * `x-podman.no_hosts` - Run the container without creating /etc/hosts file
 
+* `x-podman.passwd` - Allow Podman to add entries to /etc/passwd and /etc/group when used in conjunction with the --user option. Defaults to true.
+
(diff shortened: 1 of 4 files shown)


p12tic (changes requested):
Please add release note to newsfragments/. Otherwise looks great, thanks!

gerryd:
Added. Thanks for reviewing!

PR #1526, opened 2026-07-30
Reject non-list security_opt in service definitions

### LINKED ISSUE #1444: security_opt allows value without a dash
[`security_opt`](https://github.com/compose-spec/compose-spec/blob/main/spec.md#security_opt) element should accept a list (array?):
```yml
security_opt:
  - label:user:USER
  - label:role:ROLE
```
At least for a single value, the dash is not enforced here, so following syntax is possible for podman-compose:
```yml
security_opt:
  label=disable
```
Docker-compose doesn't allow that:
```console
$ docker compose up
validating docker-compose.yml: services.<service>.security_opt must be a array
```
Following syntax works there:
```yml
security_opt:
  - label=disable
```
```console
$ podman-compose version
podman-compose version 1.5.0
podman version 5.8.2
```
**Environment:**
 - OS: Linux

Sanjays2402:
Closes #1444

The Compose Specification defines `security_opt` as a sequence, and `docker compose` rejects a bare scalar with `services.<service>.security_opt must be a array`. `podman-compose` instead coerced a plain string into a one-element list alongside `env_file` and `volumes`, so a mistyped `security_opt: label=disable` was silently accepted while the equivalent docker compose file fails. `normalize_service()` now raises `PodmanComposeError` when the value is not a list.

Compose spec: https://github.com/compose-spec/compose-spec/blob/main/spec.md#security_opt

## Contributor Checklist:

these guidelines.

- Read, and the change follows them.

- Linked above (`security_opt` in the Compose Specification).

- Not included in this PR — happy to add `newsfragments/reject_non_list_security_opt.bugfix` with the text "Reject non-list `security_opt` in service definitions instead of silently accepting a bare scalar." on request.

- Added `test_security_opt_must_be_a_list` (parametrized over a bare string and a mapping) and `test_security_opt_list_is_normalized` to `tests/unit/test_normalize_service.py`. Both new cases fail without the change and pass with it; `tests/unit` is 470 passed.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/newsfragments/reject_non_list_security_opt.bugfix
+++ b/newsfragments/reject_non_list_security_opt.bugfix
@@ -1,0 +1,2 @@
+Reject a non-list ``security_opt`` in service definitions instead of silently
+coercing a bare scalar into a one-element list.
--- a/podman_compose.py
+++ b/podman_compose.py
@@ -2099,13 +2099,15 @@ def normalize_service(service: dict[str, Any], sub_dir: str = "") -> dict[str, A
     if "build" in service and "args" in service["build"]:
         if isinstance(build["args"], dict):
             build["args"] = norm_as_list(build["args"])
-    for key in ("env_file", "security_opt", "volumes"):
+    for key in ("env_file", "volumes"):
         if key not in service:
             continue
         if isinstance(service[key], str):
             service[key] = [service[key]]
     if "security_opt" in service:
         sec_ls = service["security_opt"]
+        if not is_list(sec_ls):
+            raise PodmanComposeError("ERROR: security_opt must be a list")
         for ix, item in enumerate(sec_ls):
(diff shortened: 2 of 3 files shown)


p12tic (changes requested):
Please add release note in newsfragments/. Otherwise looks good, thanks!

Sanjays2402:
Added in 19da98b as `newsfragments/reject_non_list_security_opt.bugfix`, following the `.bugfix` towncrier type from newsfragments/README.txt:

> Reject a non-list `security_opt` in service definitions instead of silently coercing a bare scalar into a one-element list.

`tests/unit/test_normalize_service.py` is 28 passed on the new head.

Sanjays2402:
Done @p12tic — added `newsfragments/reject_non_list_security_opt.bugfix` in `newsfragments/`. Thanks!

PR #1547, opened 2026-08-29
Do not ignore explicit empty entrypoint on `run`

### LINKED ISSUE #1546: The entrypoint can't be unset via `--entrypoint=''`
The entrypoint can't be unset via `podman compose run --entrypoint=''`

**To Reproduce**
Create `compose.yaml` with the following content:
```yaml
services:
  repro:
    image: docker.io/library/busybox:1.38.0-uclibc
    entrypoint: ["/bin/echo", "original entrypoint"]
```
Run
```
podman compose -f compose.yaml run --rm --entrypoint '' repro /bin/sh -c 'echo entrypoint-cleared'
```

**Expected behavior**
The output is `entrypoint-cleared`

**Actual behavior**
The output is `original entrypoint /bin/sh -c echo entrypoint-cleared`

**Output**

```
$ podman-compose version
podman-compose version 1.6.0
podman version 6.1.0
```

**Environment:**
 - OS: Arch Linux (probably reproducible on any)
 - podman version: 
 - podman compose version: 1.6.0-1 (Arch package; it doesn't show commit hash is was built from)

kamazee:
Fixes #1546.

> If this PR adds a new feature that improves compatibility with docker-compose, please add a link

FWIW, this change is made to match what `docker compose` does; I didn't find it specified anywhere explicitly but ability to override the entrypoint makes sense in general.


kamazee:
Also, I found it somewhat not obvious that in order to run integration tests, dependencies from `test-requirements.txt` have to be installed; it doesn't seem to be mentioned in `CONTRIBUTING.md` where the rest of the setup-related points are covered.

I didn't change it here because it doesn't really fit into this PR and I wasn't sure what to do: mention `pip install -r test-dependencies.txt` or add the difference between `test-dependencies.txt` and `devel` section of `pyproject.toml`. Do they differ intentionally?

PR #1555, opened 2026-09-04
tests/integration: Remove localhost HTTP requests from integration tests

### LINKED ISSUE #998: Integration tests should not use network connectivity if not necessary
Currently integration tests do a lot of checks using network connectivity through localhost. This makes tests more brittle because it introduces a race condition: we don't wait for the webserver to fully start up before requests are issued. Most of the time this is fine so far.

Better idea would be to use `echo <something>` as the container command, possibly wait for container to exit and then check the logs.

mokibit:
This PR removes HTTP requests to localhost in integration tests to avoid startup race conditions and external network dependencies.

Fixes https://github.com/containers/podman-compose/issues/998.

### DIFF
--- a/tests/integration/nethost/test_podman_compose_nethost.py
+++ b/tests/integration/nethost/test_podman_compose_nethost.py
@@ -3,8 +3,6 @@
 import os
 import unittest
 
-import requests
-
 from tests.integration.test_utils import RunSubprocessMixin
 from tests.integration.test_utils import podman_compose_path
 from tests.integration.test_utils import test_path
@@ -15,8 +13,7 @@ def compose_yaml_path() -> str:
 
 
 class TestComposeNethost(unittest.TestCase, RunSubprocessMixin):
-    # check if container listens for http requests and sends response back
-    # as network_mode: host allows to connect to container easily
+    # check if container with network_mode: host can be started and accessed
     def test_nethost(self) -> None:
         try:
             self.run_subprocess_assert_returncode(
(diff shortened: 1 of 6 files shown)


tests/integration/nethost/test_podman_compose_nethost.py
```diff
@@ -15,8 +13,7 @@ def compose_yaml_path() -> str:
 
 
 class TestComposeNethost(unittest.TestCase, RunSubprocessMixin):
-    # check if container listens for http requests and sends response back
-    # as network_mode: host allows to connect to container easily
+    # check if container with network_mode: host can be started and accessed
     def test_nethost(self) -> None:
```
p12tic, line 17: Here we should keep network functionality, but possibly pick up a random unused port.

tests/integration/nets_test1/test_podman_compose_nets_test1.py
```diff
@@ -37,14 +35,6 @@ def test_nets_test1(self) -> None:
             self.assertIn(b"nets_test1_web1_1", output)
             self.assertIn(b"nets_test1_web2_1", output)
 
-            response = requests.get('http://localhost:8001/index.txt')
```
p12tic, line 40: We should keep this, because this is part of what test asserts.

tests/integration/nets_test2/test_podman_compose_nets_test2.py
```diff
@@ -37,14 +35,6 @@ def test_nets_test2(self) -> None:
             self.assertIn(b"nets_test2_web1_1", output)
             self.assertIn(b"nets_test2_web2_1", output)
 
-            response = requests.get('http://localhost:8001/index.txt')
```
p12tic, line 40: Keep.

PR #1558, opened 2026-09-08
Environment secrets use correct options

mnahkies:
initial support for environment secrets was added in: https://github.com/containers/podman-compose/pull/1488 which has not yet been released.

this commit aims to improve on that initial support:
- fixes target file always being prefixed by compose project name
- supports custom targets, consistent with file secrets
- sets `uid` / `gid` / `mode` / `type` when passed, consistent with external secrets

**References:**
- top level secrets: https://github.com/compose-spec/compose-spec/blob/main/schema/compose-spec.json#L1493
- service level secret grants: https://github.com/compose-spec/compose-spec/blob/main/schema/compose-spec.json#L1882
- docs: https://github.com/compose-spec/compose-spec/blob/main/09-secrets.md

**Newsfragments:**
- I've not touched, as this is improving the unreleased feature that already has a pending news fragment.


tests/integration/secrets/docker-compose.yaml
```diff
@@ -1,4 +1,3 @@
-version: "3.8"
```
mnahkies, line 1: Avoid deprecation warning in tests

tests/integration/secrets/print_secrets.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
 grep . /run/secrets/*
+echo "CUSTOM_LOCATION:"
```
mnahkies, line 4: I added this to make the `stdout` easier to interpret when the assertions are failing

I also had trouble running these tests locally to begin with, as I was on a Fedora box with SELinux enforced, which was blocking read of the test file secret content by the container.

This wasn't immediately obvious until I found the `TESTS_DEBUG` environment variable, which then logged the permission denied warnings that were going to `stderr`. Worked around by:
```
chcon -t container_file_t ~/projects/podman-compose/tests/integration/secrets/my_secret
```
but didn't feel ideal to be doing to make the test suite work. Was tempted to add `:Z` relabeling to the `docker-compose.yaml`

